PR #21, opened 2019-07-14
An input implementation using GLFW

Perksey:
# Summary of the PR
Implements Silk.NET.Input.Desktop.

https://github.com/Ultz/Silk.NET/blob/master/documentation/proposals/Proposal%20-%20Input.md

# Further Comments


src/Input/Silk.NET.Input.Desktop/Util.cs
```diff
@@ -0,0 +1,82 @@
...
+            return GlfwProvider.ThreadDispatcher.Invoke(@delegate);
+        }
+
+        public static unsafe T* Do<T>(Func<UnsafeDispatch<T>> @delegate)
+            where T:unmanaged
+        {
+            return GlfwProvider.ThreadDispatcher.Invoke(@delegate);
+        }
+
+        public static float ApplyDeadzone(Deadzone deadzone, float val)
```
devvoid: Personally, I feel this would be better suited to be put on the Deadzone struct itself, or at least in Input.Common somewhere since every backend would need it.

Perksey: Will do!

src/Input/Silk.NET.Input.Desktop/GlfwInputPlatform.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        public bool IsApplicable(IWindow window)
+        {
+            return window is GlfwWindow;
+        }
+
+        public IInputContext GetInput(IWindow window)
+        {
+            // the cast isn't needed, but is used to make sure that what we've got is actually a GlfwWindow
+            return new GlfwInputContext((GlfwWindow)window);
```
frederikja163: Maybe we should just take a GLFW window directly, that way the user decides wether or not to cast

src/Input/Silk.NET.Input.Desktop/GlfwInputContext.cs
```diff
@@ -0,0 +1,81 @@
...
+
+namespace Silk.NET.Input.Desktop
+{
+    public class GlfwInputContext : IInputContext
+    {
+        internal List<GlfwGamepad> _gamepads;
+        internal List<GlfwJoystick> _joysticks;
+        internal IKeyboard _keyboard;
+        internal IMouse _mouse;
+        internal IWindow _window;
```
frederikja163: I think we should store the keyboard, mouse and window as the glfw version, that way we are sure they are the correct type. And it makes more sense as this is a glfw context

src/Input/Silk.NET.Input.Desktop/InputHandler.cs
```diff
@@ -0,0 +1,122 @@
...
+            {
+                return;
+            }
+            
+            Contexts.Add(ctx);
+
+            // register callbacks
+            Util.Do(() => Util.Glfw.SetKeyCallback((WindowHandle*) ctx._window.Handle, KeyCallback));
+            Util.Do(() => Util.Glfw.SetMouseButtonCallback((WindowHandle*)ctx._window.Handle, MouseCallback));
+            Util.Do(() => Util.Glfw.SetScrollCallback((WindowHandle*)ctx._window.Handle, ScrollCallback));
```
frederikja163: Could avoid casting 3 times by making a local variable

PR #29, opened 2019-07-22
Overload fixes

devvoid:
# Summary of the PR
Fix some issues with the overloader and typemap to make the resulting bindings much easier to work with, mostly by reducing the number of casts needed.

- Edit typemap to map OpenGL type GLenum to CS type GLEnum instead of int
- Modify PointerParameterOverloader to also be able to use UIntPtr
- Modify ReturnTypeOverloader to work with uint parameters too, so that functions like `GenBuffer` don't include unnecessary integer parameters.

None.

# Further Comments


build/binder_typemaps/csharp.json
```diff
@@ -1,5 +1,5 @@
 {
-  "GLenum": "int",
+  "GLenum": "GLEnum",
```
Perksey: Actually, what happens if you remove this line altogether? We do extended mapping for enums (i.e. for NV enums, ARB enums, etc)

src/Core/BuildTools/Bind/Overloading/StaticCountOverloader.cs
```diff
@@ -34,33 +33,33 @@ public IEnumerable<Overload> CreateOverloads(Function function)
             for (var i = 0; i < newParameters.Count; i++)
             {
                 var param = newParameters[i];
-                if (IsApplicable(param))
+                if (!IsApplicable(param)) {
```
Perksey: Back at it again with the java braces.

src/Core/BuildTools/Convert/Construction/ProfileConstructorExtensions.cs
```diff
@@ -33,27 +32,46 @@ public static class ProfileConstructorExtensions
         /// <returns>The resultant enum.</returns>
         public static Enum ParseEnum(XElement element)
         {
+            var elementName = element.Attribute("name"); 
+            if (elementName == null) {
```
Perksey: Fix braces everywhere please.

src/Core/BuildTools/Bind/Overloading/StaticCountOverloader.cs
devvoid: Rider's doing, not mine

build/binder_typemaps/csharp.json
devvoid: Seems like parameters are fine, but return types are broken. For example, in IArbDirectStateAccess.gen.cs:

`GLenum CheckNamedFramebufferStatus([Flow(FlowDirection.In)] uint framebuffer, [Flow(FlowDirection.In)] GLEnum target);`

Perksey: Awesome, glad to know that that's been hooked up to some extent. Should be an easy fix, go into TypeMapper.cs (in the converter) and change the MapEnums function to this:
image
You'll need to do a fullrun

devvoid: Done, now it works without that line in the typemapper.

devvoid: After giving Rider some more time to process everything, a bunch of errors popped up as a result. For example, in INVPathRendering.gen.cs:

`unsafe void MatrixLoad([Flow(FlowDirection.In)] EXT mode, [Count(Count = 16), Flow(FlowDirection.In)] float* m);`

EXT doesn't exist in this project.

Perksey: Hmm will have a look at it myself.

src/Core/BuildTools/Bind/ProfileWriter.cs
```diff
@@ -202,6 +200,8 @@ public static void WriteMixedModeClasses(this Project project, Profile profile,
             // }
             if (project.IsRoot)
             {
+                var partial = File.Exists(Path.Combine(folder, profile.ClassName + ".cs")) ? "partial" : "";
```
Perksey: Remove this. All classes should be partial in case we need to add custom functionality to them. Plus in this particular branch, this will always return an empty string as the empty mixed-mode class wouldn't have been produced yet in a clean build.

PR #30, opened 2019-08-01
Fix `Get` methods

johnkellyoxford:
# Summary of the PR
Fix the get methods to a) work properly and b) work nicely

# Further Comments
These methods in themselves are abominations, and just a weird rewrite of indexing (`Get(p, 11)` is exactly equivalent to `p[11]`). Seems logical to remove them if not part of current public API

### DIFF
--- a/examples/InputTest/Program.cs
+++ b/examples/InputTest/Program.cs
@@ -15,7 +15,9 @@ namespace InputTest
     {
         private static void Main()
         {
-            var window = Window.Create(WindowOptions.Default);
+            var opts = WindowOptions.Default;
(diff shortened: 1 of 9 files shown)


johnkellyoxford:
Entirely removed the get methods given they are only used twice

Perksey:
I think we've already done this on the glfw-input branch actually on one of the more recent commits

johnkellyoxford:
I pulled the glfw-input branch and edited it on that

Perksey:
Ah ok, I thought I got them all! Will double check this and then merge it in. :)

PR #50, opened 2019-08-19
Fix the link to fred's github page

devvoid:
Link was wrong so I fixed it.


Perksey:
I assumed his last name wrong as well, turns out it's Anderson and not whatever his middle name is (starting with J)

PR #86, opened 2019-12-27
Rewrite the GLFW input backend with maintainability and performance in mind

### LINKED ISSUE #67: Missing/Wrong mapped keys on Xbox 360 Wireless
# Summary
I noticed that some keys are not mapped correctly to names.
Button number 11 (counting from zero) is responsible for DPadUp but it s not correct. When i click DPadRight then this button is true.
There is not a a DPadLeft.

# Steps to reproduce
- Platform: Desktop
1. First I did this
I created from my .Net Standard 2.1 Library
2. Then I did this
Then I called function responsible for detecting buttons

# Comments
There is list of detected buttons
First collumn is index, second current state, last is name of button
> 0 False A
1 False B
2 False X
3 False Y
4 False LeftBumper
5 False RightBumper
6 False Back
7 False Start
8 False Home
9 False LeftStick
10 False RightStick
11 False DPadUp
12 False DPadRight
13 False DPadDown

### LINKED ISSUE #80: Multiple calls to GetInput extension lead to new glfw input contexts
# Summary
Multiple calls to the GetInput extension will create a new glfw input context each time. Therefore event registering on other than the first input context will not raise the events.

# Steps to reproduce
- Platform: Any
- Framework Version: Any
1. Create one method for initialize keyboard events and one for initialize mouse events.
2. In each method use GetInput() to retrieve the input context and add event handlers for keyboard events (first method) and mouse events (second method).
3. Mouse event callbacks are not called as the used input context is not considered by Silk. Only the first one is.

The problem is the method GetInput below. It creates a new instance of the input context instead of reusing the existing one if present.

```csharp
    public class GlfwInputPlatform : IInputPlatform
    {
        public bool IsApplicable(IWindow window)
        {
            return window is GlfwWindow;
        }

        public IInputContext GetInput(IWindow window)
        {
            // the cast isn't needed, but is used to make sure that what we've got is actually a GlfwWindow
            return new GlfwInputContext((GlfwWindow)window); // <- Problem, should reuse context when called for the second time
        }
    }
```

# Possible solutions
- Reuse context on second and subsequent calls
- Documentation that each time it is called a new context is created (in this case rename to `CreateInput`)
- Make Silk register each of the created contexts and ensure that all of them are handled

I would prefer first fix.

Perksey:
# Summary of the PR
Does what it says on the tin.

Closes #80 
Fixes #67 

# Further Comments


Perksey:
Ready for review, in an astoundingly short amount of time if I do say so myself!

src/Input/Silk.NET.Input.Common/Enums/MouseButton.cs
```diff
@@ -15,7 +15,8 @@ namespace Silk.NET.Input.Common
     /// </remarks>
     public enum MouseButton
     {
-        Left,
+        Unknown = 1,
```
devvoid: Should be -1

PR #89, opened 2019-12-27
Add a WindowManager to aid management of multiple windows

### LINKED ISSUE #82: Additional windows do not properly close
# Summary
Can not close additional windows created. Messing around with a IMGUI library and was trying to implement something like Viewports from Dear IMGUI where it creates windows on the fly to support dragging around outside of the main window.

If you close the smaller window the `IsClosing` flag is set to true, but it never actually closes.

# Steps to reproduce
- Platform: Desktop

```
  static void Main(string[] args)
        {
            var windowOptions = new WindowOptions();
            windowOptions.Size = new System.Drawing.Size(800, 600);
            windowOptions.Title = "Hello World";

            var mainWindow = Window.Create(windowOptions);
            var windows = new List<IWindow>();
            windows.Add(CreateWindow("test", 200, 200));

            mainWindow.Open();
            while (!mainWindow.IsClosing)
            {
                mainWindow.DoEvents();
                mainWindow.DoUpdate();
                mainWindow.DoRender();

                for (int i = windows.Count - 1; i >= 0; i--)
                {
                    var it = windows[i];
                    if (it.IsClosing)
                    {
                        it.Close();
                        //     windows.RemoveAt(i);
                    }
                    // else
                    {

                        it.DoEvents();
                        it.DoUpdate();
                        it.DoRender();
                    }
                }

                System.Threading.Thread.Sleep(16);
            }

        }

        public static IWindow CreateWindow(string title, int width, int height)
        {
            var windowOptions = new WindowOptions();
            windowOptions.Size = new System.Drawing.Size(200, 200);
            windowOptions.Title = "Hello World";


            var window = Window.Create(windowOptions);
            window.Open();
            return window;
        }
```

Perksey:
# Summary of the PR
~~Does what it says on the tin~~ man i need to stop saying that but in all seriousness how better can I describe this?

https://discordapp.com/channels/521092042781229087/587346162802229298/650324057501335577
Closes #82 

# Further Comments


src/Windowing/Extensions/Silk.NET.Windowing.Extensions/WindowManager.cs
```diff
@@ -0,0 +1,119 @@
...
+                }
+            }
+
+            while (IsRunning)
+            {
+                foreach (var window in Windows)
+                {
+                    if (window.IsClosing)
+                    {
+                        Windows.Remove(window);
```
maxilevi: I think this will cause a "Collection was modified" type of exception since you are modifying the Windows list in the middle of the iteration

Perksey: Fixed, introduced locks on all of the modifying APIs. Please approve & merge this PR if you have no further comments.

PR #85, opened 2019-12-26
[Wip] Tutorials

frederikja163:
# Summary of the PR
Adding tutorials to chapter 1 for c#

PR #91, opened 2019-12-30
F# Tutorial 1.3 - Shaders

PhilT:
Followed 3.3 Shader Class code somewhat but although F# has support for classes to aid Interop, a more idiomatic approach is to use functions which is what I've done. It also doesn't yet have some of the other abstractions in there as they aren't used yet.

 `Shader.use` doesn't really make sense either yet as you need to pass in everything to make that call and `use` is a reserved word so be `Shader.useProgram gl programId` just to call `gl.UseProgram(programId)`. Hopefully a more obvious abstraction will present itself in later tutorials

### DIFF
--- a/examples/F#/OpenGL/Tutorial 1.3 - Shaders/shader.fs
+++ b/examples/F#/OpenGL/Tutorial 1.3 - Shaders/shader.fs
@@ -1,0 +1,48 @@
+module Shader
+
+open System
+open Silk.NET.OpenGL
+
+// Used to load and compile the vertex and fragment shaders
+let private createShader (gl: GL) (shaderType: GLEnum) path =
+    let shader = gl.CreateShader(shaderType)
+    let source = IO.File.ReadAllText(path)
+
+    // Load and compile the shader
+    gl.ShaderSource(shader, source)
+    gl.CompileShader(shader)
+
+    // Throw an exception if compiling the shader failed
+    // The info log will tell us what line the shader failed on
+    // so we can inspect the source.
+    if gl.GetShader(shader, GLEnum.CompileStatus) <> 1 then
+        let info = gl.GetShaderInfoLog(shader)
+        gl.DeleteShader(shader)
+        failwith (sprintf "Error creating shader: %A" info)
+
+    shader
+
+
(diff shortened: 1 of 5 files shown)


examples/F#/OpenGL/Tutorial 1.3 - Shaders/shader.fs
```diff
@@ -0,0 +1,48 @@
+module Shader
+
+open System
+open Silk.NET.OpenGL
+
+// Used to load and compile the vertex and fragment shaders
+let private createShader (gl: GL) (shaderType: GLEnum) path =
```
Perksey: > `Shader.use` doesn't really make sense either yet as you need to pass in everything to make that call and `use` is a reserved word so be `Shader.useProgram gl programId` just to call `gl.UseProgram(programId)`. Hopefully a more obvious abstraction will present itself in later tutorials

Note that in Fred's PR (#85), his shader abstraction caches the GL instance for later calls. I recommend trying to mimic Fred's abstraction as closely as possible. I've requested a review from him as well.

frederikja163: Yeah, caching the gl instance seems a lot better and clearner to whoever would use the api, so i think it would be a good idea to teach good habbits, however; if it is more of a f# habbit/condition to pass it in every time, im happy with having it not be cached for the f# version

PhilT: In functional programming, pure functions don't rely on any external state. Pure functions also don't effect external state which is obviously not the case here as GL is one big state machine.
 However, I still think it's pretty idiomatic unless you start on the road of developing types to handle data that you would then eventually pass to OpenGL to process but that would create a lot of boilerplate and hide what we're really trying to teach.

PhilT: If we come up with a better pattern in later tutorials I'll revisit the earlier ones as well so we don't churn the code too much and keep the flow.

PhilT:
Closing and moving to Fred's repo

PR #97, opened 2020-01-05
Implement the latest API proposals for windowing, among other fixes

### LINKED ISSUE #80: Multiple calls to GetInput extension lead to new glfw input contexts

### LINKED ISSUE #83: Monitor API

### LINKED ISSUE #84: Window & View separation
- [ ] I have definitely checked that there are no proposals for this already in the [proposals folder](https://github.com/Ultz/Silk.NET-Docs)

# Summary of feature
Currently, our windowing API has a lot of APIs that aren't really applicable on mobile platforms and after trying to write a UWP windowing backend, I figured that this needs a bit of work. I propose that we separate the core APIs that are common among desktop and mobile platforms into their own set of interfaces, prefixed by `IView`. `IWindow` will then inherit the `IView` interfaces so that the APIs aren't duplicated and to maximise ease of cross-platform support.

Also, a `GetOrCreateView` method will have to be added as mobile window usually only support one view (for UWP this is `CoreWindow`, for Android this is `ANativeWindow`, etc etc)

I'll add a proposal shortly detailing this.

# Comments
https://discordapp.com/channels/521092042781229087/587346162802229298/658816946703171585
High Priority - Required for an internal project

Perksey:
# Summary of the PR
Implements the latest API proposals, as well as fixes a bug or implements an enhancement or two.

Closes #84 
Closes #83 
Closes #80 

# Further Comments


Pyrdacor:
How is the state of this PR?

Perksey:
Currently on hold due to availability issues among the working group. We'll try and get it merged in the next couple of weeks.

Pyrdacor:
Thanks for the update. Hope you will continue soon. If you need a helping hand, just let me know. My time is also very limited but maybe I can find some time slots for the project.

Perksey:
Ah no worries I think we're all good, SuperInvoke I think is ready for Silk.NET and SuperBind is getting ready, so all the stars are starting to align for Preview 4.

src/Windowing/Silk.NET.Windowing.Desktop/GlfwWindow.cs
```diff
@@ -873,13 +874,60 @@ public unsafe IMonitor Monitor
...
                 if (monitor == null)
-                    monitor = _glfw.GetPrimaryMonitor();
+                {
+                    if (_windowState != WindowState.Fullscreen)
+                    {
+                        // Determine which monitor this window is on. [6 marks]
+                        var monitors = new GlfwMonitorEnumerable();
+                        foreach (var m in monitors)
+                        {
+                            if (m.Bounds.Contains(Position))
```
copygirl: This wouldn't return true if the window corner is offscreen. Isn't it better to use the window's center? Even better, when the position isn't on any particular monitor, you test which is closest.

Perksey: Yeah you're right, I'll change it to use the centre and try and add in an extra test.

Pyrdacor: Or the most overlap of window and monitor. Like shown here: https://stackoverflow.com/questions/21421074/how-to-create-a-full-screen-window-on-the-current-monitor-with-glfw

Perksey: I've made it use the centre and attempted to come up with a check test, but I didn't like it as it felt overengineered plus I'm not that well-versed with the maths, so I don't feel too confident writing such a solution. I've committed it but only so you could see my potential solution - I immediately reverted it.

PR #102, opened 2020-01-07
Implementing video modes

### LINKED ISSUE #98: Make switching from fullscreen to window possible
At the moment when switching to fullscreen a GLFW monitor is set. When trying to return to a different window state (Normal, Minimized, Maximized) the monitor is still set and therefore fullscreen mode will remain.

It would be nice to allow switching back from fullscreen to window. I guess it could be done by setting the monitor back to null/0 in the mentioned case.

### LINKED ISSUE #101: Make depth buffer bits configurable
For my current game I need a depth buffer of at least 24 bits and need to enforce this on window creation. At the moment the glfw depth buffer bit hint is not called at all. Please add the depth buffer bit option to `WindowOptions` and call the associated glfw hint method. The default value may be GLFW_DONT_CARE to keep compatibility.

Pyrdacor:
# Summary of the PR
Adds video mode settings (resolution, refresh rate and depth buffer bits).

[Make depth buffer bits configurable](https://github.com/Ultz/Silk.NET/issues/101)

# Further Comments
Closes #98 
Closes #101


Pyrdacor:
@Perksey Maybe you could have a look at the implementation. FYI this should also fix #98.

Some testing is needed though.

src/Windowing/Silk.NET.Windowing.Desktop/GlfwWindow.cs
```diff
@@ -470,6 +509,10 @@ public unsafe void Initialize()
                 _initialOptions.API.Profile == ContextProfile.Core ? OpenGlProfile.Core : OpenGlProfile.Compat
             );
 
+            // Set video mode
+            _glfw.WindowHint(WindowHintInt.RefreshRate, _initialOptions.VideoMode.RefreshRate);
```
Perksey: Oh shoot didn't realize you can actually change the refresh rate. **Ignore my previous comments on removing VideoMode from IViewProperties!!!!!** Instead, should we move `DepthBufferBits` to `VideoMode` or is this not applicable?

Pyrdacor: It was part of video mode in my first try but it isn't applicable imho. You can't change it nor will it have any influence after window creation. Moreover it has nothing to do with video mode. Instead it is just a hint for the window to support an appropriate context.

src/Windowing/Silk.NET.Windowing.Common/WindowOptions.cs
```diff
@@ -40,6 +42,12 @@ public WindowOptions(ViewOptions opts)
         /// <inheritdoc />
         public bool ShouldSwapAutomatically { get; set; }
 
+        /// <inheritdoc />
+        public VideoMode VideoMode { get; set; }
+
+        /// <inheritdoc />
+        public int DepthBufferBits { get; set; }
```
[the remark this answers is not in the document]
Pyrdacor: Yeah you are right. My initial intent was to pass those "dontcare" values to the underlying glfw implementation but it doesn't make sense anyway. Will adjust the code soon.

src/Windowing/Silk.NET.Windowing.Desktop/GlfwWindow.cs
Perksey: Alright :+1:

Pyrdacor: The problem I see is that a user would expect the given depth buffer bits to be active but glfw may ignore this or choose a different bit depth if the given one is not possible. The opengl context is imho the only source which could tell if the given depth buffer bits were used. Any suggestions how we could improve user experience here?

src/Windowing/Silk.NET.Windowing.Desktop/GlfwWindow.cs
```diff
@@ -152,6 +153,31 @@ public bool IsVisible
         /// </remarks>
         public bool ShouldSwapAutomatically { get; }
 
+        /// <inheritdoc />
+        public unsafe VideoMode VideoMode
+        {
+            get
+            {
+                var monitor = _glfw.GetWindowMonitor(_windowPtr);
```
Perksey: Will this throw an error if `_windowPtr` is invalid or a nullptr?

src/Windowing/Silk.NET.Windowing.Common/Interfaces/IWindowProperties.cs
```diff
@@ -1,55 +1,60 @@
...
+
+        /// <summary>
+        /// The window border.
+        /// </summary>
+        WindowBorder WindowBorder { get; set; }
+
+        /// <summary>
+        /// Preferred depth buffer bits of the window's framebuffer.
+        /// </summary>
+        int? PreferredDepthBufferBits { get; }
```
Perksey: This should be defined in IViewProperties I think as this API is available on the mobile windowing backends we plan to support.

src/Windowing/Silk.NET.Windowing.Desktop/GlfwWindow.cs
```diff
@@ -152,6 +153,31 @@ public bool IsVisible
         /// </remarks>
         public bool ShouldSwapAutomatically { get; }
 
+        /// <inheritdoc />
+        public unsafe VideoMode VideoMode
```
Perksey: Also, couldn't we just use the IMonitor property to obtain this value rather than duplicate the native transition code?

Pyrdacor: I changed it to use the Monitor property.

src/Windowing/Silk.NET.Windowing.Desktop/GlfwWindow.cs
Pyrdacor: As I changed the whole property implementation this is no longer an issue.

src/Windowing/Silk.NET.Windowing.Common/Interfaces/IWindowProperties.cs
Pyrdacor: Moved to IViewProperties

Perksey:
> Another problem I bumped into is an assertion in GlfwMonitor which will occur if the Handle is null (which is the case when fullscreen mode is not active). Maybe we should add a safety check there or throw an exception when trying to retrieve the video mode?

- Get the window's current monitor
- If it returns a nullptr, get the system's primary monitor and return that

It's the best we can do really.

Perksey:
Thanks a lot for championing this! Make any final changes and remove the WIP mark, then I can merge this bad boy in :D

Pyrdacor:
Ok I fixed the monitor null-value issue. Unfortunately the null coalescing operator isn't available for pointers yet so the null checks look a bit bad. But it works now. I tested it now on Windows (previously on Ubuntu) and even the refresh rate is working. On the current machine I have options for 50, 59 and 60 HZ. It chooses the closest one so if I set the refresh rate to 58 it will use the value 59.

I think you can merge now. Was a pleasure to help. I look forward to preview4 as I use Silk.net in my Settlers1 clone with great success. :)

PR #104, opened 2020-01-15
Fsharp tutorial 1.3

PhilT:
Pretty closely resembles the C# version except that I'm using types and modules to represent the objects which is standard practice in F#. 

It works and displays a pulsating coloured rectangle.

Shout if you have any questions about the code.


PhilT:
Argh! Wrong repo!

PR #108, opened 2020-02-01
Added cursor types and implemented whole cursor logic

Pyrdacor:
# Summary of the PR
This adds the implementation of the cursor interface. Aside custom cursors there is now also the option to use standard (built-in) cursor shapes.

### DIFF
--- a/src/Input/Silk.NET.Input.Common/Interfaces/ICursor.cs
+++ b/src/Input/Silk.NET.Input.Common/Interfaces/ICursor.cs
@@ -11,30 +11,56 @@ namespace Silk.NET.Input.Common
     public interface ICursor
     {
         /// <summary>
+        /// Type of the cursor.
+        /// </summary>
+        CursorType Type { get; set; }
+
+        /// <summary>
+        /// Standard cursor to use.
+        /// </summary>
+        /// <remarks>
+        /// Is only used if <see cref="Type"/> is <see cref="CursorType.Standard"/>.
+        /// </remarks>
+        StandardCursor StandardCursor { get; set; }
+
+        /// <summary>
         /// Hotspot on the X axis.
         /// </summary>
+        /// <remarks>
+        /// Is only used if <see cref="Type"/> is <see cref="CursorType.Custom"/>.
+        /// </remarks>
         int HotspotX { get; set; }
-        
+
         /// <summary>
         /// Hotspot on the Y axis.
         /// </summary>
+        /// <remarks>
+        /// Is only used if <see cref="Type"/> is <see cref="CursorType.Custom"/>.
+        /// </remarks>
         int HotspotY { get; set; }
-        
+
         /// <summary>
         /// Width of the cursor in pixels.
         /// </summary>
+        /// <remarks>
+        /// Is only used if <see cref="Type"/> is <see cref="CursorType.Custom"/>.
+        /// </remarks>
         int Width { get; set; }
-        
+
         /// <summary>
         /// Height of the cursor in pixels.
         /// </summary>
+        /// <remarks>
+        /// Is only used if <see cref="Type"/> is <see cref="CursorType.Custom"/>.
+        /// </remarks>
         int Height { get; set; }
-        
+
         /// <summary>
         /// Image data for the cursor.
         /// </summary>
         /// <remarks>
         /// The image MUST be in 32-bit RGBA, non-premultiplied, and in little-endian format.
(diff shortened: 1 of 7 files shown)


Pyrdacor:
I tested it and it works quiet well. The only things I am not sure about is the disposing logic (especially the ugly try-catch-blocks) and the behavior if the user provides bad values for custom cursors.

As each property must be set on its own their will always be a bad intermediate state. Like settings pixels, width and height step-by-step will always lead to inconsistent information until all the values are setup. For now the cursor is set to null (default cursor) if the custom cursor state is not valid.

src/Input/Silk.NET.Input.Common/Interfaces/ICursor.cs
```diff
@@ -44,5 +70,12 @@ public interface ICursor
         /// <param name="mode">The mode to check for.</param>
         /// <returns>Whether or not this mode is supported.</returns>
         bool IsCursorModeSupported(CursorMode mode);
```
Perksey: How about we just rename this and `IsStandardCursorSupported` to just `IsSupported` as it feels a lot cleaner.

src/Input/Silk.NET.Input.Desktop/GlfwCursor.cs
```diff
@@ -11,23 +13,147 @@ namespace Silk.NET.Input.Desktop
...
+        {
+            _handle = handle;
+        }
+
         /// <inheritdoc />
-        public int HotspotX { get; set; }
-        
+        public unsafe CursorType Type
+        {
+            get => _cursorType;
```
Perksey: It doesn't look like these fields are ever updated by GLFW itself. For example, what if the hotspot changed on GLFW's accord? Silk would never respond to that change. Please adjust to update where appropriate. This could be one update every frame or you could subscribe to an event within the GlfwMouse Subscribe method.

Pyrdacor:
Ok I will have a look at it later.

src/Input/Silk.NET.Input.Desktop/GlfwCursor.cs
Pyrdacor: You mean when a standard cursor is set? Or do you mean someone invokes glfw functions outside of our window class? Note that a cursor is always bound to a specific window. To get the windows handle and change the cursor on its own, someone must do really evil things.

I see your point that the hotspot and dimensions also should reflect the current state for standard cursors but no other case comes to mine. What did you have in mind when you said "if the hotspot changed on GLFW's accord"?

Perksey: The hotspot is likely to change. As a result, we should be keeping HotspotX and HotspotY up-to-date. If you can think of any other properties we should also keep up-to-date, we should update them too. Either subscribe to an event using GlfwEvents or hook into the update loop as seen in GlfwJoystick

Pyrdacor: The hotspot should not change. It is a constant relative offset which stays the same as long as the cursor shape is the same. Think of the hand cursor. The pointing finger is not in the upper-left corner (in contrast to the arrow cursor). So the hand cursor has a hotspot of like { x: 4, y: 0}. These values will not change unless the shape is changed. Cursor movement has no effect on the hotspot.

Perksey: Hmmm well are there any other properties you think that might be of concern?

Pyrdacor: Not really. They will only change on user action in code. The only thing is that hotspot and width will not reflect the correct values if a standard cursor is active. I will think about how to handle this nicely.

Perksey: Is that a goal for this PR or a future PR?

Pyrdacor: I guess it is not a big deal to keep it like it is. I will create a new PR later if I have a good solution. I think this PR can be merged. If you want we can rename the two mentioned methods but I don't know if I will have time for it in the next days.

Perksey: Yeah let's rename the overloads and then I'll merge this in

Pyrdacor:
@Perksey Ready for merge. :)

Perksey:
Awesome, thanks!

PR #94, opened 2020-01-02
[WIP] Fix VSync

devvoid:
# Summary of the PR
Fix the broken VSync system.

# Related issues, Discord discussions, or proposals
#93 

# Further Comments
Waiting on more info from #93 before merging.

Closes #93

PR #109, opened 2020-02-01
Replaced vsync sleep logic

### LINKED ISSUE #93: Vsync broken in GlfwWindow
# Summary
Vsync is the wrong way round.

When creating a window and setting vsync to on the framerate is unlimited and setting vsync off limits the framerate to the refresh rate. This is the exact opposite of the desired behaviour.

# Comments
Looking at [GlfwWindow.cs](https://github.com/Ultz/Silk.NET/blob/master/src/Windowing/Silk.NET.Windowing.Desktop/GlfwWindow.cs) in ''RaiseRenderFrame" on line 631 to 641 there is the below code:

```cs
// Identical to RaiseUpdateFrame.
if (FramesPerSecond > double.Epsilon
     && (VSync == VSyncMode.Off || VSync == VSyncMode.Adaptive && IsRunningSlowly))
{
    var sleepTime = _renderPeriod - _renderStopwatch.Elapsed.TotalSeconds;

    if (sleepTime > 0.0)
    {
        Thread.Sleep((int) (1000 * sleepTime));
    }
}
```

Reading the above, unless i'm mistaken if vsync is off or adaptive and running slow, we sleep the thread resulting in the framerate being limited whereas if vsync is on we don't sleep therefore the framerate is unlimited.

Surely this logic is backwards?

Pyrdacor:
# Summary of the PR
Replaced VSYNC sleep logic with an approach where the events are just not called until the time is right. This avoids many problems with frame rates and update rates.

Closes #93 
#94 

# Further Comments
Testing seems to work as expected. Needs some testing for adaptive vsync.

### DIFF
--- a/src/Windowing/Silk.NET.Windowing.Desktop/GlfwWindow.cs
+++ b/src/Windowing/Silk.NET.Windowing.Desktop/GlfwWindow.cs
@@ -625,11 +625,12 @@ namespace Silk.NET.Windowing.Desktop
 
             // Calculate delta and run frame.
             var delta = _updateStopwatch.Elapsed.TotalSeconds;
-            Update?.Invoke(delta);
 
             // Reset
             _updateStopwatch.Restart();
             _updatedWithinPeriod = false;
+
+            Update?.Invoke(delta);            
         }
 
         /// <summary>
@@ -668,13 +669,6 @@ namespace Silk.NET.Windowing.Desktop
             }
 
             var delta = _renderStopwatch.Elapsed.TotalSeconds;
-            
-            Render?.Invoke(delta);
-
-            if (ShouldSwapAutomatically)
-            {
-                SwapBuffers();
-            }
 
             //Reset
             _renderStopwatch.Restart();
@@ -684,6 +678,13 @@ namespace Silk.NET.Windowing.Desktop
             if (VSync == VSyncMode.Adaptive)
             {
                 _glfw.SwapInterval(IsRunningSlowly ? 0 : 1);
+            }
+
+            Render?.Invoke(delta);
+
+            if (ShouldSwapAutomatically)
+            {
+                SwapBuffers();
             }
         }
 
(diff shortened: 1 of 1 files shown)


Pyrdacor:
Ok I tested and fixed adaptive vsync. Should all work now.

src/Windowing/Silk.NET.Windowing.Desktop/GlfwWindow.cs
```diff
@@ -640,7 +640,7 @@ private void RaiseRenderFrame()
         {
             // Identical to RaiseUpdateFrame.
             if (FramesPerSecond >= double.Epsilon
-                && (VSync == VSyncMode.Off || VSync == VSyncMode.Adaptive && IsRunningSlowly))
+                && (VSync == VSyncMode.Off || VSync == VSyncMode.Adaptive))
```
Perksey: You are no longer using the IsRunningSlowly property, which is core to adaptive vsync.
```cs
    /// <summary>
    /// Use adaptive VSync. If the window IsRunningSlowly, VSync will be disabled. Otherwise,
    /// it will be enabled. If the framerate is uncapped, this is the same as <see cref="On"/>. 
    /// </summary>
```
Am I missing something here?

Pyrdacor: The logic that sets IsRunningSlowly is inside that if block. So if I only enter if it is set, it will never be set at all.

It is still used by SetSwapInterval.

Perksey: Ah ok. As long as you can assure me that IsRunningSlowly is used somewhere to influence the swap interval this is fine with me.

Perksey:
Finally, will this defo fix both of the issues linked above?

src/Windowing/Silk.NET.Windowing.Desktop/GlfwWindow.cs
Pyrdacor: Yeah it is. Moreover the return statement below your mentioned code only is executed in the non-adaptive case, as in adaptive mode there can never be more time than needed. If there was, running slow is obviously false and the behavior is like adaptive on. So the logic before didn't make much sense. Testing was successful for all vsync modes but it's good if someone else checks again.

Pyrdacor:
It's only one issue which should be fixed. The other one is void's PR. This PR is based on hers. So yes.

Pyrdacor:
I'm glad I could help. :) Will someone merge this into master?

PR #114, opened 2020-02-02
Monitor video modes and gamma correction

### LINKED ISSUE #110: Add a GetAllVideoModes method to IMonitor to aid video mode selection.
- [x] I have definitely checked that there are no proposals for this already in the [proposals folder](https://github.com/Ultz/Silk.NET-Docs)

# Summary of feature
We should add a GetAllVideoModes method to IMonitor so that users can access all the available video modes as described by GLFW. This will allow for better resolution changing for Silk.NET.

# Comments
Requested by Zaphyk, working group member and Project Hedra dev. High priority.

devvoid:
# Summary of the PR
Adds the ability to get all of a monitor's video mode and gamma correction.

Closes #110 

# Further Comments
GLFW primarily works with gamma ramps, not plain exponents. While it has a function to set gamma ramp via exponent, there doesn't seem to be a way to get that exponent *back* from GLFW. Because of this, we'll have to assume the default gamma is 1.0 when a monitor is first created. I don't know enough about monitors to know if that's going to be an issue or not.

Setting video modes will come in a later PR, GLFW implements this in a very odd way that makes it difficult to implement.

### DIFF
--- a/src/Windowing/Silk.NET.Windowing.Common/Interfaces/IMonitor.cs
+++ b/src/Windowing/Silk.NET.Windowing.Common/Interfaces/IMonitor.cs
@@ -3,6 +3,7 @@
 // You may modify and distribute Silk.NET under the terms
 // of the MIT license. See the LICENSE file for details.
 
+using System.Collections.Generic;
 using System.Drawing;
 
 namespace Silk.NET.Windowing.Common
@@ -41,6 +42,6 @@ namespace Silk.NET.Windowing.Common
         /// Get all video modes that this monitor supports.
         /// </summary>
         /// <returns>An array of all video modes.</returns>
-        VideoMode[] GetAllVideoModes();
+        IEnumerable<VideoMode> GetAllVideoModes();
     }
 }
--- a/src/Windowing/Silk.NET.Windowing.Common/Interfaces/IMonitor.cs
+++ b/src/Windowing/Silk.NET.Windowing.Common/Interfaces/IMonitor.cs
@@ -7,11 +7,40 @@ using System.Drawing;
 
 namespace Silk.NET.Windowing.Common
 {
+    /// <summary>
+    /// An interface representing a monitor.
+    /// </summary>
     public interface IMonitor : IWindowHost
     {
+        /// <summary>
+        /// The name of this monitor.
+        /// </summary>
         string Name { get; }
+        
+        /// <summary>
+        /// The index of this monitor.
+        /// </summary>
         int Index { get; }
+        
+        /// <summary>
+        /// The bounds of this monitor.
+        /// </summary>
         Rectangle Bounds { get; }
+        
+        /// <summary>
+        /// The current video mode of this monitor.
+        /// </summary>
         VideoMode VideoMode { get; }
+        
+        /// <summary>
+        /// This monitor's gamma correction.
+        /// </summary>
+        float Gamma { get; set; }
+
+        /// <summary>
+        /// Get all video modes that this monitor supports.
+        /// </summary>
+        /// <returns>An array of all video modes.</returns>
+        VideoMode[] GetAllVideoModes();
     }
 }
--- a/src/Windowing/Silk.NET.Windowing.Desktop/GlfwMonitor.cs
+++ b/src/Windowing/Silk.NET.Windowing.Desktop/GlfwMonitor.cs
@@ -68,18 +68,18 @@ namespace Silk.NET.Windowing.Desktop
             }
         }
 
-        public VideoMode[] GetAllVideoModes()
+        public IEnumerable<VideoMode> GetAllVideoModes()
         {
             var rawVideoModes = GlfwProvider.GLFW.Value.GetVideoModes(Handle, out var count);
 
-            List<VideoMode> videoModes = new List<VideoMode>();
+            var videoModes = new List<VideoMode>();
 
             for (var i = 0; i < count; i++)
             {
                 videoModes.Add(new VideoMode(new Size(rawVideoModes[i].Width, rawVideoModes[i].Height), rawVideoModes[i].RefreshRate));
             }
 
-            return videoModes.ToArray();
+            return videoModes;
         }
     }
 }
--- a/src/Windowing/Silk.NET.Windowing.Desktop/GlfwMonitor.cs
+++ b/src/Windowing/Silk.NET.Windowing.Desktop/GlfwMonitor.cs
@@ -3,7 +3,6 @@
 // You may modify and distribute Silk.NET under the terms
 // of the MIT license. See the LICENSE file for details.
 
-using System;
 using System.Drawing;
 using Silk.NET.GLFW;
 using Silk.NET.Windowing.Common;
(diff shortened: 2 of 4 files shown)


src/Windowing/Silk.NET.Windowing.Common/Interfaces/IMonitor.cs
```diff
@@ -1,17 +1,46 @@
...
+        /// <summary>
+        /// This monitor's gamma correction.
+        /// </summary>
+        float Gamma { get; set; }
+
+        /// <summary>
+        /// Get all video modes that this monitor supports.
+        /// </summary>
+        /// <returns>An array of all video modes.</returns>
+        VideoMode[] GetAllVideoModes();
```
Perksey: Is there a specific reason you opted for an array and not `IEnumerable<VideoMode>` here? Later on you seem to do an additional copy from list to array so I think making it IEnumerable may be more efficient.

src/Windowing/Silk.NET.Windowing.Desktop/GlfwWindow.cs
```diff
@@ -159,14 +159,12 @@ public bool IsVisible
...
-        public unsafe VideoMode VideoMode
+        public VideoMode VideoMode
         {
             get
             {
                 var monitor = Monitor;
-                return monitor != null
-                    ? monitor.VideoMode
-                    : _initialOptions.VideoMode;
+                return monitor?.VideoMode ?? _initialOptions.VideoMode;
```
Perksey: Could we get rid of the extra variable now that we no longer need it?

src/Windowing/Silk.NET.Windowing.Common/Interfaces/IMonitor.cs
Pyrdacor: Agree with @Perksey. I also would remove the mentioned `ToArray` call.

src/Windowing/Silk.NET.Windowing.Desktop/GlfwMonitor.cs
```diff
@@ -66,5 +67,19 @@ public float Gamma
                 GlfwProvider.GLFW.Value.SetGamma(Handle, value);
             }
         }
+
+        public VideoMode[] GetAllVideoModes()
+        {
+            var rawVideoModes = GlfwProvider.GLFW.Value.GetVideoModes(Handle, out var count);
+
+            List<VideoMode> videoModes = new List<VideoMode>();
```
Pyrdacor:
```suggestion
            var videoModes = new List<VideoMode>();

```

src/Windowing/Silk.NET.Windowing.Desktop/GlfwMonitor.cs
```diff
@@ -66,5 +67,19 @@ public float Gamma
                 GlfwProvider.GLFW.Value.SetGamma(Handle, value);
             }
         }
+
+        public VideoMode[] GetAllVideoModes()
```
Pyrdacor:
```suggestion
        public IEnumerable<VideoMode> GetAllVideoModes()

```

src/Windowing/Silk.NET.Windowing.Desktop/GlfwMonitor.cs
```diff
@@ -66,5 +67,19 @@ public float Gamma
...
+            var rawVideoModes = GlfwProvider.GLFW.Value.GetVideoModes(Handle, out var count);
+
+            List<VideoMode> videoModes = new List<VideoMode>();
+
+            for (var i = 0; i < count; i++)
+            {
+                videoModes.Add(new VideoMode(new Size(rawVideoModes[i].Width, rawVideoModes[i].Height), rawVideoModes[i].RefreshRate));
+            }
+
+            return videoModes.ToArray();
```
Pyrdacor:
```suggestion
            return videoModes;

```

src/Windowing/Silk.NET.Windowing.Common/Interfaces/IMonitor.cs
Pyrdacor:
```suggestion
        IEnumerable<VideoMode> GetAllVideoModes();

```

src/Windowing/Silk.NET.Windowing.Common/Interfaces/IMonitor.cs
```diff
@@ -1,17 +1,46 @@
...
-        Rectangle Bounds { get; }
-        VideoMode VideoMode { get; }
-    }
-}
+// This file is part of Silk.NET.
+// 
+// You may modify and distribute Silk.NET under the terms
+// of the MIT license. See the LICENSE file for details.
+
+using System.Drawing;
```
Pyrdacor:
```suggestion
using System.Collections.Generic;
using System.Drawing;

```

src/Windowing/Silk.NET.Windowing.Desktop/GlfwWindow.cs
Pyrdacor: What do you mean exactly?

Perksey: Currently we have `var monitor = Monitor`, which we no longer need now that we’re only accessing the monitor once

src/Windowing/Silk.NET.Windowing.Desktop/GlfwMonitor.cs
```diff
@@ -66,5 +67,19 @@ public float Gamma
                 GlfwProvider.GLFW.Value.SetGamma(Handle, value);
             }
         }
+
+        public VideoMode[] GetAllVideoModes()
+        {
+            var rawVideoModes = GlfwProvider.GLFW.Value.GetVideoModes(Handle, out var count);
+
```
Pyrdacor: I wonder if we can ignore the count parameter at all. As C# arrays have a length the count parameter makes only sense in C imho. By discarding the count parameter with `out _` we also can get rid of the for loop below and transform it to a one-liner with Linq's `Select`. So the whole method body can be transformed into a short one-liner.

src/Windowing/Silk.NET.Windowing.Desktop/GlfwWindow.cs
Pyrdacor: Ah good point.

src/Windowing/Silk.NET.Windowing.Desktop/GlfwMonitor.cs
Pyrdacor: But we need a null check here if I understand the glfw docs correctly.

devvoid: rawVideoModes isn't a C# array, but a pointer. We need Count because it lacks the length info needed for a foreach/linq expression.

Pyrdacor: Ok didn't check it as I'm at mobile phone atm. Then it's OK.

PR #121, opened 2020-02-11
Redo overloads and delete my sanity

Perksey:
# Summary of the PR
why am i doing this to myself again

Links go here.

# Further Comments


Perksey:
# Progress
## Early
- [x] String
- [x] Group

## Prompt
- [ ] Span & Ref

## Late
- [ ] IntPtr
- [ ] ArrayParameter
- [ ] ReturnType
- [ ] StringArray

## Scrapped
- Flow Pointer (replaced with ref)
- String return (handled in the string early overloader)
- Pointer parameter (replaced with span & ref)
- Pointer return value (scrapped entirely)

Perksey:
This is ready for review, but it's probably best we review the bindings rather than the binder as we've been operating under the `as long as it works idgaf what it looks like` code style for the binder.

@frederikja163 seeing as you're the main driving force behind this change, can you be responsible for testing the usefulness of this new overload system?

frederikja163:
I can take it for a test drive sure, you send me the nuget packages over and i will give it a shot on the tutorials

Perksey:
Will do, just need to fix the group overloads which seemingly still don’t work

frederikja163:
Good job!

PR #126, opened 2020-02-16
Implement cursor modes

### LINKED ISSUE #124: Cursor mode is not fully implemented :/
# Summary of feature
Cursor mode enum is there but it is not used yet nor is it changeable.

# Comments
Basically I forgot to add this functionality in my previous PR. Shame on me. :( This has still to be implemented. Shouldn't be much of work hopefully.

Pyrdacor:
# Summary of the PR
Extend ICursor interface to make cursor mode changeable. Implement cursor mode behavior for GLFWCursor.

Fixes #124

### DIFF
--- a/src/Input/Silk.NET.Input.Common/Interfaces/ICursor.cs
+++ b/src/Input/Silk.NET.Input.Common/Interfaces/ICursor.cs
@@ -24,6 +24,14 @@ namespace Silk.NET.Input.Common
         StandardCursor StandardCursor { get; set; }
 
         /// <summary>
+        /// Mode of the cursor.
+        /// </summary>
+        /// <remarks>
+        /// See <see cref="CursorMode"/> for details.
+        /// </remarks>
+        CursorMode CursorMode { get; set; }
+
+        /// <summary>
         /// Hotspot on the X axis.
         /// </summary>
         /// <remarks>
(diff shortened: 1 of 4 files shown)


Pyrdacor:
I have to add a call to `Glfw.GetInputMode` to retrieve raw mouse motion flag. The return value is bool in that case but there is already a version of `GetInputMode` with the same parameters which returns `CursorModeValue`. As C# doesn't support return type covariance how can I implement this with SuperInvoke?

My guess would be returning just `object` or use a generic version like `T GetInputMode<T>(...)` but I don't know how SuperInvoke works in that case. Would prefer the generic version if this is possible.

Perksey:
Just add a GLFW-specific struct in the Silk.NET.GLFW package and convert it to the Silk.NET.Input.Common struct after

Pyrdacor:
Hm I noticed that the return type in glfw is just int which is interpreted as bool (0 or 1) or as the enums int value in the other case. Maybe I just change it to int and convert it when I call it.

Perksey:
I’m not sure how good SuperInvokes marshalling with stuff like out bool is but if it’s just a normal bool SuperInvoke will automatically Marshall it to 0 or 1

Pyrdacor:
Should work as expected now. Can be merged I guess.

PR #136, opened 2020-03-04
Add click events

### LINKED ISSUE #78: Add built-in support for click/doubleclick
The input layer only has mouse up and down events. Double clicks are a very common use case. At the moment every Silk user is forced to implement this functionality on its own. A general built-in support is much appreciated.

Adding events for Click and DoubleClick would be nice.

Pyrdacor:
# Summary of the PR
Add events for single click and double click.

Closes #78

### DIFF
--- a/src/Input/Silk.NET.Input.Common/Interfaces/IMouse.cs
+++ b/src/Input/Silk.NET.Input.Common/Interfaces/IMouse.cs
@@ -52,6 +52,15 @@ namespace Silk.NET.Input.Common
         /// </summary>
         event Action<IMouse, MouseButton> MouseUp;
         
+        /// <summary>
+        /// Called when a single click is performed.
+        /// </summary>
+        event Action<IMouse, MouseButton> Click;
+
+        /// <summary>
+        /// Called when a double click is performed.
+        /// </summary>
+        event Action<IMouse, MouseButton> DoubleClick;
         
         /// <summary>
         /// Called when the mouse is moved.
--- a/src/Input/Silk.NET.Input.Common/Interfaces/IMouse.cs
+++ b/src/Input/Silk.NET.Input.Common/Interfaces/IMouse.cs
@@ -35,6 +35,17 @@ namespace Silk.NET.Input.Common
         /// </summary>
         ICursor Cursor { get; }
         
+        /// The maximum time in milliseconds between two
+        /// consecutive clicks to count as a double click.
+        /// </summary>
+        int DoubleClickTime { get; set; }
+
+        /// <summary>
+        /// The maximum distance in pixels between two
+        /// consecutive clicks to count as a double click.
+        /// </summary>
+        int DoubleClickRange { get; set; }
+        
         /// <summary>
         /// Checks if a specific button is currently pressed.
         /// </summary>
--- a/src/Input/Silk.NET.Input.Desktop/GlfwInputContext.cs
+++ b/src/Input/Silk.NET.Input.Desktop/GlfwInputContext.cs
@@ -53,6 +53,8 @@ namespace Silk.NET.Input.Desktop
             GlfwInputPlatform.RegisterWindow((WindowHandle*) Handle, _subscribers);
             window.Update += _update = _ =>
             {
+                _mice[0]?.Update();
+
                 foreach (var updatable in _gamepads)
                 {
                     updatable.Update();
--- a/src/Input/Silk.NET.Input.Desktop/GlfwInputContext.cs
+++ b/src/Input/Silk.NET.Input.Desktop/GlfwInputContext.cs
@@ -53,7 +53,10 @@ namespace Silk.NET.Input.Desktop
             GlfwInputPlatform.RegisterWindow((WindowHandle*) Handle, _subscribers);
             window.Update += _update = _ =>
             {
-                _mice[0]?.Update();
+                foreach (var updatable in _mice)
+                {
+                    updatable.Update();
+                }
 
                 foreach (var updatable in _gamepads)
                 {
--- a/src/Input/Silk.NET.Input.Desktop/GlfwMouse.cs
+++ b/src/Input/Silk.NET.Input.Desktop/GlfwMouse.cs
@@ -25,7 +25,7 @@ namespace Silk.NET.Input.Desktop
         private GlfwCallbacks.ScrollCallback _scroll;
         private GlfwCallbacks.CursorPosCallback _cursorPos;
         private GlfwCallbacks.MouseButtonCallback _mouseButton;
-        private bool _firstClick = false;
+        private bool _firstClick = true;
         private MouseButton? _firstClickButton = null;
         private PointF _firstClickPosition = PointF.Empty;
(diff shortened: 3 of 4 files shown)


Pyrdacor:
I didn't test anything and there is a TODO comment cause I don't know how to handle the case yet. So this is still WIP. :D

Pyrdacor:
@Perksey I can't build the current master on Ubuntu 19.04. He complains about missing references. Hundreds of lines like this:

Interfaces/IAtiVertexStreams.gen.cs(673,10): error CS0246: The type or namespace name 'NativeSymbolAttribute' could not be found (are you missing a using directive or an assembly reference?) [/home/pyrdacor/projects/Misc/Silk.NET/src/OpenGL/Extensions/Silk.NET.OpenGL.Legacy.Extensions.ATI/Silk.NET.OpenGL.Legacy.Extensions.ATI.csproj]
Interfaces/IAtiVertexStreams.gen.cs(673,10): error CS0246: The type or namespace name 'NativeSymbol' could not be found (are you missing a using directive or an assembly reference?) [/home/pyrdacor/projects/Misc/Silk.NET/src/OpenGL/Extensions/Silk.NET.OpenGL.Legacy.Extensions.ATI/Silk.NET.OpenGL.Legacy.Extensions.ATI.csproj]

I ran `generator.sh` and then `dotnet build Silk.NET.sln`. Am I missing something?

Perksey:
You'll need to delete the Interfaces, Structs, and Enums folder due to the changes made when moving away from ADL to SuperInvoke. It's probably best to commit your changes to some branch somewhere and then clone the repo anew.

Pyrdacor:
Ok thanks. The following did the trick. :D

    rm -r src
    git reset --hard

Pyrdacor:
Ok I tested it with the Triangle example and it seems to work. Maybe someone could check again. I used:
```cs
           var mouse = _window.CreateInput().Mice.First();
            mouse.Click += (mouse, button) => {
                Console.WriteLine("Single click with button: " + button.ToString());
            };
            mouse.DoubleClick += (mouse, button) => {
                Console.WriteLine("Double click with button: " + button.ToString());
            };
```

Pyrdacor:
@Perksey @devvoid I think it's ready for merge now.

Perksey:
Could you add this to the InputTest program in the lab please so we can test this again in the future? Once this is done I'll have a go at testing it on my end.

Pyrdacor:
@Perksey Done. Yeah double click logic is much harder than it looks on first glance. I read a few articles about it and there is not even a general approach for this. Some people use timers but my solution luckily doesn't even need timers or events which would possibly introduce parallel executions. It's all handled nicely by the update routine. :)

Pyrdacor:
@Perksey I re-checked my code. There might be a problem with multi-threaded update logic though as the Update event won't be tied to the same thread as the glfw event raising. I might be wrong but maybe we should consider that. It is a bit sad because it will make the logic much more complicated. :/

Pyrdacor:
I wonder if we could add an internal update event which is not affected by the "single threaded" option and which is only used for input etc. Because we don't want asynchronous input updates/events at all.

Perksey:
Should be a non-issue, glfwPollEvents is always ran on the main thread with the main game loop, which then calls the events bound within Silk.

Perksey:
I’ll have a play when I get home.

Pyrdacor:
The problem is the following code inside GlfwInputContext.cs:

```cs
window.Update += _update = _ =>
{
	foreach (var updatable in _mice
	{
		updatable.Update();
	}

	foreach (var updatable in _gamepads)
	{
		updatable.Update();
	}
                
	foreach (var updatable in _joysticks)
	{
		updatable.Update();
	}
};
```

`window.Update` is raised by a Task in multi-threaded mode. Therefore the Update methods are also called within the Task and this isn't what we want at all I guess.

So I thought about adding an `InternalUpdate` or `SingleThreadedUpdate` only to `GlfwWindow` (not to the interface) which is used in the above code instead.

GlfwWindow will then raise the internal event on the main thread in any case. As an end-user will not even see this event through the interface it is no big deal imho.

```cs
public void DoUpdate()
{
	RaiseInternalUpdate(); // this is new and specific for GlfwWindow

	if (UseSingleThreadedWindow)
	{
		RaiseUpdateFrame();
	}
	else
	{
		// Raise UpdateFrame, but don't await it yet.
		var task = Task.Run(RaiseUpdateFrame); // cast to action, ambiguous call

		// Loop while we're still updating - the Update thread might be calling the main thread
		while (!task.IsCompleted)
		{
			if (!_invokeQueue.IsEmpty && _invokeQueue.TryDequeue(out var invokeCall))
			{
				invokeCall.GetAwaiter().GetResult();
			}
		}
	}
}
```

Pyrdacor:
Another option would be to provide a MainThreadUpdate in any case for the end-user (which we could use for input then as well). So an end-user could decide to run code on the main thread or on another thread in multi-threaded mode. But I'm not sure if this is useful.

Perksey:
if (window.SingleThreaded)
    update as normal
else
    use window.Invoke

Pyrdacor:
@Perksey Hopefully I understood you right. See my last commit.

Pyrdacor:
Is the Invoke logic broken? I am not really understanding it but when I replace the `new Task<object>...` part in `GlfwWindow.Invoke` with `var task = Task.Run(() => d.DynamicInvoke(args));` it is working. Otherwise a call to `Invoke` from within the `Update` event will cause a deadlock.

Perksey:
The task can’t be started immediately, which is what Task.Run does. Instead what we do is register the delegate as a task and await it later on.

Pyrdacor:
I understand but it seems that it never is started at all. Does `invokeCall.GetAwaiter().GetResult();` even start the task?

Pyrdacor:
Wouldn't `RunSynchronously` be the right choice here?

Perksey:
GetAwaiter().GetResult() is the most common way to synchronously wait for a task to complete. I don’t think changing the invoke system is within the scope for this PR.

Pyrdacor:
I agree but my problem now is that the current invoke logic leads to a deadlock for multi-threaded windows (e.g. the input test project).

But if I understand it correctly, the multi-threaded logic will still assure that the Update event is triggered between two `DoEvents` calls. Am I right? If so we won't need the last commit of this PR at all.

Perksey:
Currently 
- the update event is called in another thread 
- on the main thread, we start waiting and executing incoming invocations until the invocation is complete

Pyrdacor:
Ok so the main loop waits till the update event is finished before it starts again. This is all that's important.

Pyrdacor:
Ok I force pushed to omit the last commit with the unneccessary invoke logic. Test project was extended as said before. So merge and test to your liking. :)

Perksey:
Tested my end, works. Approved to merge.

Pyrdacor:
Yeah we could change the default values a bit. I guess 500ms is default on windows but it seems very high.

In general it is best to set it to the OS setting but this must be handled by the API user.

PR #138, opened 2020-03-06
Window icons

Perksey:


Pyrdacor:
A side note: the cursor uses the same struct for image data as the icon. So your new struct should be used for cursors too imho. Cause otherwise we have two different representations of the same data structure.

Perksey:
Ah yeah. that's unfortunate. Perhaps we can find some way to fix that in 1.0.1 - 1.0 is releasing today!

Pyrdacor:
Yup. Just wanted to mention it so we won't forget about it.

PR #144, opened 2020-03-15
Tutorial chapter 1

frederikja163:
# Summary of the PR
Adds chapter one of the tutorials to the examples folder.

# Further Comments
Expected outputs
https://discordapp.com/channels/521092042781229087/521092043288608781/688751771551924415


frederikja163:
Could i get a review of this, mostly regarding the amount of comments, wether its too little or to much. The expected outputs are on discord, so you can see what sort of progression the tutorials have.

examples/CSharp/Tutorial 1.5 - Transformations/Shader.vert
```diff
@@ -0,0 +1,13 @@
+#version 330 core
```
Perksey: Another thing: I hate to sound like my computer science teacher yelling at me for not commenting a hello world program.....

.....but perhaps you should put some comments in the shaders?? :P

examples/CSharp/Tutorial 1.5 - Transformations/Transform.cs
```diff
@@ -0,0 +1,27 @@
+using System;
+using System.Collections.Generic;
+using System.Numerics;
+using System.Text;
+
+namespace Tutorial
+{
+    public class Transform
```
Perksey: Could this be converted to a `struct`? It feels kinda icky using a class for data such as this.

Perksey:
Superb work, nearly there right off the bat! The comments in most instances are appropriate for code-only tutorials, we can always waffle on a bit more in the web tutorials if needed.

Perksey:
Oh also, don't forget to purge the F# tutorials (and move the contents of the C# subdirectory into the root `examples` directory) as we decided against having them for now.

examples/CSharp/Tutorial 1.5 - Transformations/Transform.cs
frederikja163: Problem with this is that means we cant use field initialisers which are essential as you might not want to set a rotation for an object. This could maybe be elaboratedd on in the web version saying that people could use structs but we use classes for this reason

examples/CSharp/Tutorial 1.3 - Abstractions/Program.cs
```diff
@@ -0,0 +1,104 @@
...
+
+        private static unsafe void OnRender(double obj)
+        {
+            Gl.Clear((uint)ClearBufferMask.ColorBufferBit);
+
+            //Binding and using our VAO and shader.
+            Vao.Bind();
+            Shader.Use();
+            //Setting a uniform.
+            Shader.SetUniform("uBlue", (float)Math.Sin(DateTime.Now.Millisecond / 1000f * Math.PI));
```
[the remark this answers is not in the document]
frederikja163: Thought i already mentioned this was /1000f gives us more precision since second is just a long

examples/CSharp/Tutorial 1.4 - Textures/Texture.cs
```diff
@@ -0,0 +1,73 @@
...
+            Image<Rgba32> img = (Image<Rgba32>)Image.Load(path);
+
+            fixed (void* data = &MemoryMarshal.GetReference(img.GetPixelSpan()))
+            {
+                //Loading the actual image.
+                Load(gl, data, (uint)img.Width, (uint)img.Height);
+            }
+
+            //Deleting the img from imagesharp.
+            img.Dispose();
```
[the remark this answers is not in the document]
frederikja163: I personally dont like using as that is basically just rembering to remember, which is wrong on a lot of levels. Furthermore you might want to save the image for more further calculations and it provides us with unnecessary nesting which dosent look good in code

examples/CSharp/Tutorial 1.5 - Transformations/Shader.vert
frederikja163: im thinking about not doing this for the hello quad as that would leave us trying to explain more code than necessary for your first quad.

examples/CSharp/Tutorial 1.4 - Textures/Texture.cs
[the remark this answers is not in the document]
Perksey: Fair enough, willing to let that slide :)

examples/CSharp/Tutorial 1.5 - Transformations/Shader.vert
Perksey: Okay, but to my knowledge none of the shaders are commented and I think some explaination of what's happening, however brief, should be put in via comments. I'm not after much, just a brief sentence or two explaining what's going on)

examples/CSharp/Tutorial 1.5 - Transformations/Transform.cs
Perksey: Hmm fair enough. Reluctantly letting this slide :P

PR #149, opened 2020-04-02
[2.0] Proposal for better strings

Perksey:
# Summary of the PR
Adds the specification for a better string API.

https://discordapp.com/channels/521092042781229087/521092043288608781/694275305543696474
https://discordapp.com/channels/521092042781229087/521092043288608781/694195737197543426
https://discordapp.com/channels/521092042781229087/521092043288608781/693633976879022131

# Further Comments
**Requires a working group majority to merge, due to it being a mass proposal.**

- Assigned Void for review as she's a maintainer.
- Assigned Fred for review as he knows what's best for the beginner/average consumer.
- Assigned Max for review as he knows what's best for production.
- Request for comments @tannergooding & @sunkin351 as they've been active in discussion
- **This will be the main talking point at the meeting on Friday 10th April.**


documentation/proposals/Proposal - Better Strings.md
```diff
@@ -0,0 +1,102 @@
...
+converted to and from their existing C#/.NET types, so that the user doesn't have to deal with much
+
+# Proposed API
+Proposed Silk.NET.Core additions:
+```cs
+public struct Char8
+{
+    private byte _value; // 8-bit character
+    public Char8(byte v) => _value = v;
+    public static implicit operator Char8(char c);
```
tannergooding, line 33: This isn't necessarily a "safe" implicit conversion.

If this is part of a UTF8 multi-byte character, it would not be correctly converted to `char`.

documentation/proposals/Proposal - Better Strings.md
```diff
@@ -0,0 +1,102 @@
...
+{
+    private char _value; // 16-bit character
+    public Char16(char v) => _value = v;
+    // insert appropriate operators and methods here as seen in Char8
+}
+
+public struct Char32
+{
+    private uint _value; // 32-bit character
+    public Char32(char v0, char v1) => _value = ((uint)v0 << 16) + (uint)v1;
```
tannergooding, line 51: This isn't a correct conversion as it doesn't account for surrogates.

documentation/proposals/Proposal - Better Strings.md
```diff
@@ -0,0 +1,102 @@
...
+    // insert appropriate operators and methods here as seen in Char8
+}
+
+public ref struct String8
+{
+    private Span<Char8> _value;
+    public String8(Span<Char8> v) => _value = v;
+    public static implicit operator String8(Span<Char8> v);
+    public static implicit operator String8(string x); // not recommended for use, as we'll have to assume the encoding is UTF8.
```
tannergooding, line 60: Not sure what the comment is saying?

The only safe assumption is that `string` is a sequence of UTF-16 characters and that you would need to transcode to UTF-8

documentation/proposals/Proposal - Better Strings.md
```diff
@@ -0,0 +1,102 @@
...
+    public static implicit operator String8(Char8* v); // we don't know the length here, so the backing span will just have a length of 1
```
tannergooding, line 61: This isn't how most C APIs work. A C `char* `or `wchar*` without an associated length is null terminated, so the proper thing to do would be:
```csharp
var s = new Span<Char8>(v, int.MaxValue);
return s.Slice(0, s.IndexOf((byte)'\0')).AsString()
```

tannergooding: You could manually track the backing pointer + length so this doesn't need to be a ref struct. Then you could use `length == -1` to indicate "null-terminated, length unknown"
If you don't want to incur this as an up front cost.

documentation/proposals/Proposal - Better Strings.md
```diff
@@ -0,0 +1,102 @@
...
+    public static implicit operator String8(byte* v); // cast to Char8*, pass to the above.
+    public static implicit operator String8(char* v); // cast to byte*, pass to the above. for maintaining back-compat with Silk.NET 1.0.
```
tannergooding, line 63: This is potentially problematic, not only for the reasons above (normally this is a null terminated sequence, when used as a "string" like type) but also because it may not represent valid UTF-8 or ASCII data.

documentation/proposals/Proposal - Better Strings.md
tannergooding: Basically UTF-16 is either a standalone 16-bit character or it is two surrogates
So its either a zero extension or a different algorithm that combines the surrogate halves (with appropriate checks that v0 is a high surrogate and v1 is a low).

sunkin351: I would suggest using `char.ConvertToUtf32()` here, assuming that uses the correct algorithm.

tannergooding: Yes it does: https://source.dot.net/#System.Private.CoreLib/Char.cs,890

Would just need to cast from `int` to `uint` for the result.

Perksey:
General consensus was that this is ok-ish, needs WG approval.

documentation/proposals/Proposal - Better Strings.md
Perksey: Gotcha, will make it explicit so that overflow exceptions aren't thrown around

Perksey:
```suggestion
    public static explicit operator Char8(char c);
```

documentation/proposals/Proposal - Better Strings.md
Perksey: Will make it an explicit conversion to prevent accidental use (and maybe a warning in the form of an obsolete attribute)

Perksey:
```suggestion
    public static explicit operator String8(char* v); // cast to byte*, pass to the above. for maintaining back-compat with Silk.NET 1.0.
```

Perksey:
@tannergooding @sunkin351 changes implemented, one final lookover would be appreciated before i get wg sign-off.

Perksey:
On second thought this is really unnecessary. Just gonna use `byte*` everywhere in Silk.NET 2.0 as it makes the most sense and that's our opportunity to make breaking changes anyway.

PR #153, opened 2020-04-02
[1.X] [WIP] Platform hints proposal

Perksey:
# Summary of the PR
Currently, there are some window hints that Silk.NET.Windowing does not expose. This proposal adds a thin hint API so that these APIs are still usable somewhat eventhough Silk.NET doesn't boast explicit suport for it.

Links go here.

# Further Comments
No


Perksey:
Adapt proposal to use extension-based/optional APIs (for example you can check if a feature is available before attempting to use it), don't take any further until done so. 

Allow a user to do what they need to do in native API and hand the resultant handle back in

Perksey:
Merged in as rejected proposal, the API was found to be undesirable.

PR #154, opened 2020-04-02
Implement IDispose on IView

### LINKED ISSUE #150: IView doesn't implement IDisposable

devvoid:
# Summary of the PR
Add IDispose to IView and all deriving classes.

Closes #150 

# Further Comments

### DIFF
--- a/src/Windowing/Silk.NET.Windowing.Desktop/GlfwWindow.cs
+++ b/src/Windowing/Silk.NET.Windowing.Desktop/GlfwWindow.cs
@@ -630,13 +630,20 @@ namespace Silk.NET.Windowing.Desktop
         private void Dispose(bool disposing)
         {
             Reset();
-            _glfw.GcUtility.Unpin(_onClosing);
-            _glfw.GcUtility.Unpin(_onMaximized);
-            _glfw.GcUtility.Unpin(_onMinimized);
-            _glfw.GcUtility.Unpin(_onMove);
-            _glfw.GcUtility.Unpin(_onResize);
-            _glfw.GcUtility.Unpin(_onFileDrop);
-            _glfw.GcUtility.Unpin(_onFocusChanged);
+            
+            // All callbacks are initialized at the same time,
+            // so checking each one individually shouldn't be
+            // necessary.
+            if (_onClosing != null)
+            {
+                _glfw.GcUtility.Unpin(_onClosing);
+                _glfw.GcUtility.Unpin(_onMaximized);
+                _glfw.GcUtility.Unpin(_onMinimized);
+                _glfw.GcUtility.Unpin(_onMove);
+                _glfw.GcUtility.Unpin(_onResize);
+                _glfw.GcUtility.Unpin(_onFileDrop);
+                _glfw.GcUtility.Unpin(_onFocusChanged);
+            }
         }
 
         public void Dispose()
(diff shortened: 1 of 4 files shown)


src/Windowing/Silk.NET.Windowing.Desktop/GlfwWindow.cs
```diff
@@ -594,10 +594,53 @@ public unsafe void Reset()
...
+                _glfw.GcUtility.Unpin(_onMaximized);
+                _glfw.GcUtility.Unpin(_onMinimized);
+                _glfw.GcUtility.Unpin(_onMove);
+                _glfw.GcUtility.Unpin(_onResize);
+                _glfw.GcUtility.Unpin(_onFileDrop);
+                _glfw.GcUtility.Unpin(_onFocusChanged);
+            }
+            if (disposing)
+            {
+                _glfw?.Dispose();
```
Perksey, line 629: Isn't `_glfw` a reference to `GlfwProvider.GLFW.Value`? Wouldn't this break all other windows when disposing one?

src/Windowing/Silk.NET.Windowing.Desktop/GlfwWindow.cs
```diff
@@ -594,10 +594,53 @@ public unsafe void Reset()
...
+            {
+                _glfw.DestroyWindow(_windowPtr);
+            }
+#pragma warning disable 168
+            // If the window is already destroyed, it throws an exception,
+            // but we want the window destroyed anyways, so just ignore it
+            catch(GlfwException _e)
+#pragma warning restore 168
+            {
+
```
Perksey, line 608: Should probably move the comment into the exception block so Rider doesn't kick up a fuss.

src/Windowing/Silk.NET.Windowing.Desktop/GlfwWindow.cs
```diff
@@ -594,10 +594,53 @@ public unsafe void Reset()
...
+            
             _windowPtr = (WindowHandle*) 0;
         }
 
+        private void Dispose(bool disposing)
+        {
+            unsafe
+            {
+                Reset();
+                _glfw.GcUtility.Unpin(_onClosing);
```
Perksey, line 619: This may throw an exception when disposing a window that's never been opened, as `_onClosing` (and related callbacks) will be null in that case. Also I'm not sure when `_glfw` is assigned but we should probably allow that being null too, just to make the dispose method as unbreakable as possible.

src/Windowing/Silk.NET.Windowing.Desktop/GlfwWindow.cs
devvoid: Rider did that automatically when implementing the dispose pattern, oops.

src/Windowing/Silk.NET.Windowing.Desktop/GlfwWindow.cs
devvoid: How should I go about avoiding that?

Perksey: Probably null checks, but seems kinda messy.

src/Windowing/Silk.NET.Windowing.Desktop/GlfwWindow.cs
```diff
@@ -609,9 +609,53 @@ public unsafe void Reset()
...
+        // Disable parameter because 
+        // ReSharper disable once UnusedParameter.Local
+        private void Dispose(bool disposing)
+        {
+            Reset();
+            
+            // All callbacks are initialized at the same time,
+            // so checking each one individually shouldn't be
+            // necessary.
+            if (_onClosing != null)
```
Perksey, line 637: Hmmm in future you should probably prefer `is null` rather than `== null` as it can do weird things if the operator is improperly implemented. Luckily though, it doesn’t matter in this case it’s just something to keep in mind.

Issue #48, opened 2019-08-14
Tracking issue for Silk.NET 2.0


Perksey:
# Development
- [x] [Complete] DirectX and Windows bindings  **(!!!!!! MORE !!!!!!)** - @Perksey 
- [x] [Complete] Assimp bindings - @Perksey 
- [x] [Complete] Android support - @Perksey 
- [x] [Complete] iOS support **(!!!!! NEEDS ROBUSTNESS IMPROVEMENTS ASAP !!!!!)** - @Perksey 
- [x] [Complete] SDL support - @Perksey 
- [x] [Complete] Maths library - @HurricanKai 
- [x] [Complete] Bindings generator overhaul - @Perksey 
- [x] [Complete] OpenXR bindings - @Perksey
- [x] [Complete] Function pointer interop source generator - @HurricanKai 
- [x] [Complete] Overload overhaul (as discussed in 14/08/2020's meeting) - @Perksey

# Milestones
- [x] Preview 1 with Assimp and experimental OpenXR bindings (July 2020)
- [x] Preview 2 with SDL bindings/windowing support and new interop (~~Update Friday September~~ August 2020)
- [x] Preview 3 with mobile windowing support (~~Early October~~ Early November 2020)
- [x] Preview 4 with Maths, as well as DirectX bindings (Early December 2020)
- [x] Preview 5 release candidate (Late December 2020)
- [ ] Full release :tada: (New Year's Day 2021)

Dates are a loose estimate and are subject to change.

PR #173, opened 2020-04-24
[WIP] [2.0] A maths and hardware intrinsics library for Silk.NET and .NET 5

Perksey:
# Summary of the PR
- Adds a library for using SIMD instructions with .NET
- Adds a library containing generic matrices, vectors, and quaternions; as well as their related maths ops.

2.0 

#48 

# Further Comments
**DO NOT SQUASH AND MERGE** It must be merged in using a merge commit due to Gamma working on it too.


src/Maths/Silk.NET.Intrinsics/Avx/AvxRegister.Double.cs
```diff
@@ -0,0 +1,330 @@
+// This file is part of Silk.NET.
+// 
+// You may modify and distribute Silk.NET under the terms
+// of the MIT license. See the LICENSE file for details.
+
+namespace Silk.NET.Intrinsics.Avx
+{
+    public partial struct AvxRegister : IRegister<double>
```
tannergooding, line 8: What does "Register" mean?

src/Maths/Silk.NET.Intrinsics/Avx/AvxRegister.Float.cs
```diff
@@ -0,0 +1,330 @@
...
+        {
+            throw new System.NotImplementedException();
+        }
+
+        public WorkUnit<float> Length(WorkUnit<float> vector)
+        {
+            throw new System.NotImplementedException();
+        }
+
+        public WorkUnit<float> Normalize2(WorkUnit<float> vector)
```
tannergooding, line 20: I'm not sure `Normalize2` is a "clear" name. I understand what it means with context, but it isn't immediately obvious.

src/Maths/Silk.NET.Intrinsics/Avx/AvxRegister.Float.cs
```diff
@@ -0,0 +1,330 @@
...
+        {
+            throw new System.NotImplementedException();
+        }
+
+        public WorkUnit<float> Shuffle(WorkUnit<float> value, byte control)
+        {
+            throw new System.NotImplementedException();
+        }
+
+        public WorkUnit<float> X(WorkUnit<float> vector)
```
tannergooding, line 150: Is this `GetX`?

src/Maths/Silk.NET.Intrinsics/Avx/AvxRegister.Float.cs
```diff
@@ -0,0 +1,330 @@
...
+        {
+            throw new System.NotImplementedException();
+        }
+
+        public WorkUnit<float> MultiplyAddFast(WorkUnit<float> x, WorkUnit<float> y, WorkUnit<float> z)
+        {
+            throw new System.NotImplementedException();
+        }
+
+        public WorkUnit<float> NegateMultiplyAddFused(WorkUnit<float> x, WorkUnit<float> y, WorkUnit<float> z)
```
tannergooding, line 220: Why is an explicit `NegateMultiplyAdd` needed? Seems like an optimization the JIT should (and does) do...

src/Maths/Silk.NET.Intrinsics/Avx/AvxRegister.Float.cs
```diff
@@ -0,0 +1,330 @@
...
+        {
+            throw new System.NotImplementedException();
+        }
+
+        public unsafe WorkUnit<float> ToVector3(float* ptr)
+        {
+            throw new System.NotImplementedException();
+        }
+
+        public unsafe WorkUnit<float> ToVector4(float* ptr)
```
tannergooding, line 325: Is this supposed to be a `Load` counterpart to `Store`?

src/Maths/Silk.NET.Intrinsics/Avx/AvxRegister.Helpers.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Runtime.Intrinsics;
+using System.Runtime.Intrinsics.Arm;
+using System.Runtime.Intrinsics.X86;
+
+namespace Silk.NET.Intrinsics.Avx
+{
+    public partial struct AvxRegister
+    {
+        public static WorkUnitFlags Flags { get; } = GetFlags();
+        public static WorkUnitFlags Flags128F { get; } = Flags | WorkUnitFlags.Vector128 | WorkUnitFlags.TypeFloat;
```
tannergooding, line 15: Framework design guidelines recommends using the full non language specific name so there is no ambiguity.

That is, these should be `Flags128Single`, `Flags128UInt64`, etc.

src/Maths/Silk.NET.Intrinsics/Common/WorkUnit128.cs
```diff
@@ -0,0 +1,11 @@
+using System.Runtime.Intrinsics;
+
+namespace Silk.NET.Intrinsics
+{
+    internal struct WorkUnit128<T> where T:unmanaged
+    {   
+        public WorkUnitFlags Flags { get; set; }
+        public Vector128<T> Vector { get; set; }
+        public unsafe fixed byte Padding[16];
```
tannergooding, line 9: Why is `fixed byte Padding` and why does it need both the `Vector` and the padding?

src/Maths/Silk.NET.Intrinsics/Common/WorkUnit256.cs
```diff
@@ -0,0 +1,10 @@
+using System.Runtime.Intrinsics;
+
+namespace Silk.NET.Intrinsics
+{
+    internal struct WorkUnit256<T> where T:unmanaged
+    {
+        public WorkUnitFlags Flags { get; set; }
+        public Vector256<T> Vector { get; set; }
```
tannergooding, line 8: This `WorkUnit` is going to be 64-bytes, seems like a lot of wasted space...

src/Maths/Silk.NET.Intrinsics/Avx/AvxRegister.Double.cs
Perksey: The name is poorly chosen, but in the context of the intrinsics library a register is a class capable of performing mathematical operations for a given type using a SIMD register such as AVX or SSE.

src/Maths/Silk.NET.Intrinsics/Avx/AvxRegister.Float.cs
Perksey: XML docs will cover that when we add them post-development.

src/Maths/Silk.NET.Intrinsics/Avx/AvxRegister.Float.cs
Perksey: Yes.

src/Maths/Silk.NET.Intrinsics/Avx/AvxRegister.Float.cs
Perksey: This will use the FMA register if applicable, otherwise it will use your everyday AVX register.

src/Maths/Silk.NET.Intrinsics/Avx/AvxRegister.Float.cs
Perksey: Yeah, probably poorly named.

src/Maths/Silk.NET.Intrinsics/Avx/AvxRegister.Helpers.cs
Perksey: Gotcha, will rename though these are meant to be private properties that for some reason I left public.

src/Maths/Silk.NET.Intrinsics/Common/WorkUnit128.cs
Perksey: This is to ensure that you can convert WorkUnit128 to WorkUnit and vice versa using Unsafe.As.

src/Maths/Silk.NET.Intrinsics/Common/WorkUnit256.cs
Perksey: Possibly, however typically short-lived.

src/Maths/Silk.NET.Intrinsics/Avx/AvxRegister.Double.cs
tannergooding: Maybe `IVector` or something would be a better name?

Is `Avx` referring to `VEX` encoded (128-bit or 256-bit) vectors or strictly 256-bit vectors? If the latter, it might also benefit from a clearer name.

src/Maths/Silk.NET.Intrinsics/Avx/AvxRegister.Float.cs
tannergooding: You might also consider renaming to `NormalizeVector2` which would disambiguate and not force users to rely on docs.

src/Maths/Silk.NET.Intrinsics/Avx/AvxRegister.Double.cs
Perksey: Yeah VEX-encoded, but I don't want that leaking out to the user as I want this to be relatively easy to use and not try to introduce too many new concepts to the user - all the user needs to know is "fast maths ooh shiney"

src/Maths/Silk.NET.Intrinsics/Avx/AvxRegister.Float.cs
tannergooding: Sure, but I'm not sure that clarifies why it is needed?

I can't think of an optimization you can do knowing that it is `-(a * b) + c` vs `(a * b) + c`, in which case you should be able to just have `MultiplyFusedAdd` and let the JIT optimize `MultiplyFusedAdd(Negate(a), b, c)` when FMA is available and just do your normal math, including the negation, when it isn't

Perksey: Very good point, cc @sunkin351

src/Maths/Silk.NET.Intrinsics/Common/WorkUnit256.cs
tannergooding: What's the purpose of `WorkUnitFlags`, they look to encode the type and size of the vector, but it isn't clear how its meant to be used in this context?

Perksey: I'm considering removing it as I don't think it's worth having. Originally it was gonna hold which set of instructions to use (i.e. AVX or SSE) so that we can have a static class that redirects maths operations to the correct implementation, however I think we can do that without the flags and probably get better treatment from the JIT too.

src/Maths/Silk.NET.Intrinsics/Common/WorkUnit128.cs
tannergooding: Would it be better to have a `Vector128<T> Reserved { get; set; }` or `Vector128<T> Upper` ?

src/Maths/Silk.NET.Intrinsics/Common/WorkUnit256.cs
tannergooding: That sounds reasonable. It would also help with throughput due to reduced memory usage, etc.

src/Maths/Silk.NET.Intrinsics/Avx/AvxRegister.Float.cs
tannergooding: Is there meant to be a `SetX` or `WithX` counterpart? (depending on if mutable or immutable)

src/Maths/Silk.NET.Intrinsics/Common/WorkUnit128.cs
Perksey: I don't know, would it? As long as the space gets filled I suppose it doesn't really matter, the WorkUnitXXX types aren't really public-facing APIs.

src/Maths/Silk.NET.Intrinsics/Avx/AvxRegister.Double.cs
tannergooding: Is it being VEX encoded an important detail? The operation is ultimately the same, just with better codegen under VEX.
It normally only gets interesting when the size changes, since that changes how much you are processing, etc.

src/Maths/Silk.NET.Intrinsics/Common/WorkUnit128.cs
tannergooding: The fixed sized buffers generally generate security cookies and other bits. Also worse codegen due to being many more fields. Having it be a `Vector128<T>` would remove that and clarify how the bits are reserved and may be interpreted.
It would also allow it to be interpreted as an HVA struct for ABI purposes (assuming you do get rid of `Flags` like you mentioned might be a consideration below).

Perksey: Ah fair enough, the more you know :D 

Will implement.

Perksey:
# Generator ops
- [ ] LD1/LD2/LD3/LD4 - loads a value and stores it into a vector
- [ ] ST1/ST2/ST3/ST4 - stores the vector.
- [ ] LDC - load constant value
- [ ] LDB - loads a shuffle value
- [ ] AND - and operation
- [ ] XOR - exclusive or
- [ ] OR - or
- [ ] ADD.H - horizontal add
- [ ] ADD - add
- [ ] SUB - substract
- [ ] MUL - multiply
- [ ] DIV - divide
- [ ] REC - reciprocal
- [ ] SQRT - square root
- [ ] RECSQRT - reciprocal square root
- [ ] MAX - maximum
- [ ] MIN - minimum
- [ ] RND - round to nearest integer
- [ ] RND.Z - round to zero (truncate)
- [ ] RND.L - round down (floor)
- [ ] RND.H - round up (ceiling)
- [ ] EQ - compare equal
- [ ] NEQ - compare not equal
- [ ] GT - compare greater than
- [ ] LT - compare less than
- [ ] GTE - greater than or equal
- [ ] LTE - less than or equal
- [ ] STV - broadcasts a scalar to a vector
- [ ] PMT - permutes a vector using a control byte
- [ ] SHF - shuffles two vectors using a control byte
- [ ] SIN - computes the sine of a vector's elements
- [ ] COS - computes the cosine of a vector's elements
- [ ] TAN - computes the tangent of a vector's elements
- [ ] SIN.A - computes the sine of a vector's elements (approx)
- [ ] COS.A - computes the cosine of a vector's elements (approx)
- [ ] TAN.A - computes the tangent of a vector's elements (approx)
- [ ] ATAN - atan
- [ ] ATAN2 - atan2
- [ ] DOT2/DOT3/DOT4 - dot product
- [ ] CRO2/CRO3/CRO4 - cross product

Perksey:
Replaced with #190

PR #187, opened 2020-06-02
stabilization of update/FPS limiting

martinhorcicka:
# Summary of the PR
Stabilizes the Update/Render limiters, ie. when the FPS drops below the desired amount, then the update period gets decreased accordingly in hopes that the sudden drop gets compensated for. 

current release

### DIFF
--- a/src/Windowing/Silk.NET.Windowing.Desktop/GlfwWindow.cs
+++ b/src/Windowing/Silk.NET.Windowing.Desktop/GlfwWindow.cs
@@ -655,6 +655,9 @@ namespace Silk.NET.Windowing.Desktop
         /// <inheritdoc />
         public unsafe void Reset()
         {
+            _updateTimeDeficit = 0;
+            _renderTimeDeficit = 0;
+
             _updateStopwatch.Stop();
             _renderStopwatch.Stop();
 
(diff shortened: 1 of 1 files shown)


Perksey:
Would this work? It seems like it's just allowing one slow frame to make every single following frame faster. Continue this cycle enough times and you have the window running at an uncapped framerate against your wishes.

martinhorcicka:
When a fast frame occurs, then the difference of `delta - _renderPeriod` will be negative, which will subsequently reduce `timeDifference` accordingly. That should be the stabilization effect.

src/Windowing/Silk.NET.Windowing.Desktop/GlfwWindow.cs
```diff
@@ -830,6 +832,9 @@ private void RaiseUpdateFrame()
             // Calculate delta and run frame.
             var delta = _updateStopwatch.Elapsed.TotalSeconds;
 
+            // Calculate how much the delta differs from the desired updatePeriod.
+            _updateTimeDeficit += delta - _updatePeriod;
```
Perksey, line 836: Can you please ensure these variables are zeroed as part of the Reset call?

martinhorcicka: Done.

PR #191, opened 2020-06-15
Feature/maths vectors

HurricanKai:
This PR will add Vector2<T>, Vector3<T>, Vector4<T>
There may be changes between this and #190 as discussed.

### DIFF
--- a/src/Maths/Silk.NET.Maths/Scalar.cs
+++ b/src/Maths/Silk.NET.Maths/Scalar.cs
@@ -9,7 +9,7 @@ using System.Runtime.CompilerServices;
 namespace Silk.NET.Maths
 {
     internal static class Scalar<T>
-        where T : unmanaged
+        IFormattable where T : unmanaged, IFormattable
     {
         /* Note: The following patterns are used throughout the code here and are described here
         *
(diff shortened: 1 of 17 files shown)


src/Maths/Silk.NET.Maths/Vector2.cs
```diff
@@ -268,145 +258,340 @@ public static void BaryCentric
...
         }
 #endif
 
 #if BTEC_INTRINSICS
-        public System.Numerics.Vector<T> AsVector()
+        public unsafe Vector<T> AsVector()
         {
-            throw new NotImplementedException();
+            Vector<T> vec = new Vector<T>();
+            Unsafe.Copy(&vec, ref X);
```
Perksey, line 593: Vector\<T\> does not have a size guarantee and may not be able to fit the vector, so be sure to check on Vector\<T\>.Count before filling it.

Also, I recommend using the official constructors as I presume the various HWIntrinsics won't be used during load if filled like this (could be wrong though, better safe than sorry however)

src/Maths/Silk.NET.Maths/Vector2.cs
```diff
@@ -268,145 +258,340 @@ public static void BaryCentric
...
+            }
+
+            if (typeof(T) == typeof(long))
+            {
+                return (Vector256<T>)(object)Vector256.Create((long)(object)X, (long)(object)Y, 0, 0);
+            }
+
+            /*if (typeof(T) == typeof(Half))
+            {
+                retrn (Vector256<T>)(object)Vector256.Create((Half)(object)X, (Half)(object)Y, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0);
```
Perksey, line 571: Should probably just throw in this case with an informative exception.

src/Maths/Silk.NET.Maths/Vector2.cs
```diff
@@ -268,145 +258,340 @@ public static void BaryCentric
...
+            }
+
+            if (typeof(T) == typeof(long))
+            {
+                return (Vector128<T>)(object)Vector128.Create((long)(object)X, (long)(object)Y);
+            }
+
+            /*if (typeof(T) == typeof(Half))
+            {
+                retrn (Vector128<T>)(object)Vector128.Create((Half)(object)X, (Half)(object)Y, 0, 0, 0, 0, 0, 0);
```
Perksey, line 510: Same here.

src/Maths/Silk.NET.Maths/Vector2.cs
```diff
@@ -268,145 +258,340 @@ public static void BaryCentric
...
+            }
+
+            if (typeof(T) == typeof(int))
+            {
+                return (Vector64<T>)(object)Vector64.Create((int)(object)X, (int)(object)Y);
+            }
+
+            /*if (typeof(T) == typeof(Half))
+            {
+                return (Vector64<T>)(object)Vector64.Create((Half)(object)X, (Half)(object)Y, 0, 0);
```
Perksey, line 454: Same here.

src/Maths/Silk.NET.Maths/Vector2.cs
HurricanKai: I thought Vector<T> is guaranteed 128 bits? And as Vector<T> is 128 bits at mac itself, this seemed reasonable. Will check tomorrow

Perksey: image

HurricanKai:
Scalar<T> is mostly ready for review.
There will be some additions (like Acos right now) but overall it's consistent now.
cc @Perksey

src/Maths/Silk.NET.Maths/Scalar.cs
```diff
@@ -0,0 +1,845 @@
...
+        {
+            throw new NotSupportedException("Vector<T> too small too fit");
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static T SquareRoot(T value)
+        {
+            if (typeof(T) == typeof(byte))
+            {
+                return (T) (object) (MathF.Sqrt((byte) (object) value));
```
Perksey, line 183: MathF isn't available in .NET Standard 2.0, only 2.1 at the earliest. Is it possible to make do with the regular `Math` type?

src/Maths/Silk.NET.Maths/Scalar.cs
```diff
@@ -0,0 +1,845 @@
...
+// 
+// You may modify and distribute Silk.NET under the terms
+// of the MIT license. See the LICENSE file for details.
+
+using System;
+using System.Runtime.CompilerServices;
+
+namespace Silk.NET.Maths
+{
+    internal static class Scalar<T>
```
Perksey, line 11: Eventually I will want to look at making this a public API, but I'd need to put some time into thinking about what I want to include in it and formally spec up the API. It's good you've inadvertently made a start! So yeah, there isn't really a review comment here, other than at some point once we have an initial prototype I'd like to look at making this public. Future stuff though, all good for now.

src/Maths/Silk.NET.Maths/Scalar.cs
```diff
@@ -0,0 +1,845 @@
...
+                }
+
+                ThrowInvalidType();
+                return default; // can't be reached
+            }
+        }
+        
+        internal static void ThrowInvalidType()
+        {
+            throw new NotSupportedException(); // TODO: Document what went wrong
```
Perksey, line 143:
```suggestion
            throw new NotSupportedException("This operation isn't supported for the current type.");
```

src/Maths/Silk.NET.Maths/Scalar.cs
```diff
@@ -0,0 +1,845 @@
...
+        }
+
+        internal static void ThrowIndexOutOfRange()
+        {
+            throw new IndexOutOfRangeException();
+        }
+        
+        internal static void ThrowVectorTTooSmall()
+        {
+            throw new NotSupportedException("Vector<T> too small too fit");
```
Perksey, line 175:
```suggestion
            throw new NotSupportedException("Vector<T> too small to fit");
```

src/Maths/Silk.NET.Maths/Scalar.cs
HurricanKai: Theoretically yes, but I'd think it can affect performance.
I've added a `using MathF = System.Math` when .Net Standard 2.0 is used.

Perksey:
Awesome, Vec2 has my sign-off then :D

HurricanKai:
Happy to drop the previous commit if considered out of scope

HurricanKai:
Happy I could help :) I'm fine as is

Issue #39, opened 2019-08-10
C# 8 Nullable Support


HurricanKai:
# Summary of feature
C# Nullable to be supported / used in the libary. C# 8 is just around the corner and this would improve ease of use with C# 8. Personally I don't feel like this is slowing development of Libs at all.

# Comments

HurricanKai:
Together with this, it might make sense to implement `readonly` members. Skimming over some of the code there seem to be a lot of cases where this could be applied, and I don't see downsides to using it (besides that It'll take time to get used to it)

HurricanKai:
Also, If applicaple, `Index` and `Range` overloads could be added.

Perksey:
> 
> 
> Also, If applicaple, `Index` and `Range` overloads could be added.

To which methods?

HurricanKai:
All indexers currently using `int`

Perksey:
Ah okay. Would love some help with this! Currently, I'm not going to be able to work on Silk for a little bit as I have other Ultz projects I need to catch up on.

Perksey:
This has proven to be quite the grueling task. I've decided to move this to a future release, as the binder is currently undergoing some heavy refactorings.

Perksey:
Closed in #45

Perksey:
This was prematurely closed! In 2.0, we'd like to add nullability attributes to the following codebases:
- [ ] Maths
- [ ] Input
- [ ] Windowing
- [ ] Core

Perksey:
Done in #260

PR #196, opened 2020-06-18
Nullability

HurricanKai:
Also contains CLS compliance, but that's a tiny code change
See also #39


HurricanKai:
Please review @Perksey 😄

src/Maths/Silk.NET.Maths/ClsCompliance.cs
```diff
@@ -0,0 +1,3 @@
+﻿using System;
+
+[assembly: CLSCompliant(true)]
```
Perksey, line 3: Could this instead be defined using the csproj? It seems weird to have assembly attributes defined in a random file wheras the others are autogenerated using the csproj.

```xml
<ItemGroup>
    <AssemblyAttribute Include="System.CLSCompliant">
      <_Parameter1>true</_Parameter1>
    </AssemblyAttribute>
</ItemGroup>
```

HurricanKai: Didn't find how to do that, will apply!

HurricanKai: Well apparently
```cs
<ItemGroup>
    <AssemblyAttribute Include="System.CLSCompliant">
      <_Parameter1>true</_Parameter1>
    </AssemblyAttribute>
</ItemGroup>
```
will only function for `string`s, generating something like: `[assembly: System.CLSCompliant("true")]`
but I've renamed to AssemblyInfo.cs

Perksey: Huh that's a little annoying, all good.

PR #200, opened 2020-06-20
Scalar2

HurricanKai:
Theres a ton of changes in there, but the Code-critical ones are:
- .editorconfig has been slightly altered
- there are now two `Scalar` types, `Scalar<T>` for Properties that forward to `Scalar<T>`, and `Scalar<T>` for actual implementations
- code cleanup has been run, making some code worse, but overall consistency is better.
- `Scalar` has a new important pattern added to it. 
- `Scalar` changes should always be checked with JIT to prevent unnecessary `call`s, a comment in regards to this has been added.

Note that git is really bad at picking up on the actual changes, in many cases `Scalar<T>` has either been removed or replaced with `Scalar` but that combined with some formatting changes leads to git just marking it as one huge difference.

Also note that I've used `using static Silk.NET.Maths.Scalar;` everywhere, eliminating any possible explicit `Scalar`s (note that when there is a local method named the same as the `Scalar` method the `Scalar` method cannot be used without explicitly specifying `Scalar`, I'd overall suggest `using static Silk.NET.Maths.Scalar;` since there are nearly no uses were it isn't clear what the function means without the `Scalar` prefix)


HurricanKai:
@Perksey please review (I don't think it's necessary to review files besides Scalar / possibly ScalarTests, since the others are only find-and-replace and code cleanup)

Issue #210, opened 2020-06-28
Examples still use obsolete APIs


Perksey:

Perksey:
Fixed in #212

PR #212, opened 2020-06-30
#210 - Start of updating obsolete methods in example projects

KinsonDigital:
# Summary
Updates method calls in example projects to use non-deprecated versions of the method calls.

# Links
[#210 - Examples still use obsolete APIs](https://github.com/Ultz/Silk.NET/issues/210)

### DIFF
--- a/examples/CSharp/Tutorial 1.1 - Hello Window/Tutorial 1.1 - Hello Window.csproj
+++ b/examples/CSharp/Tutorial 1.1 - Hello Window/Tutorial 1.1 - Hello Window.csproj
@@ -10,4 +10,8 @@
       <PackageReference Include="Silk.NET" Version="1.1.1" />
     </ItemGroup>
 
+    <ItemGroup>
+      <ProjectReference Include="..\..\..\src\Core\Silk.NET\Silk.NET.csproj" />
+    </ItemGroup>
+
 </Project>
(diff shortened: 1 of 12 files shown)


examples/CSharp/Tutorial 1.5 - Transformations/Tutorial 1.5 - Transformations.csproj
```diff
@@ -8,10 +8,13 @@
     </PropertyGroup>
 
     <ItemGroup>
-      <PackageReference Include="Silk.NET" Version="1.1.1" />
       <PackageReference Include="SixLabors.ImageSharp" Version="1.0.0-beta0007" />
```
Perksey, line 11: Looks like this still uses beta 7 where the others use rc 3

examples/CSharp/Tutorial 1.1 - Hello Window/Tutorial 1.1 - Hello Window.csproj
```diff
@@ -10,4 +10,8 @@
       <PackageReference Include="Silk.NET" Version="1.1.1" />
```
Perksey, line 10: Could be wrong but GitHub mobile seems to think that you have a package reference and a project reference to silk in this project file, whereas we only need the latter.

examples/CSharp/Tutorial 1.5 - Transformations/Tutorial 1.5 - Transformations.csproj
KinsonDigital: Fixed

examples/CSharp/Tutorial 1.1 - Hello Window/Tutorial 1.1 - Hello Window.csproj
KinsonDigital: Strange.  The solution explorer showed no nuget packages but the XML was still in the csproj file.

VS bug?

Issue fixed

PR #215, opened 2020-07-01
Finalize Maths API

HurricanKai:
Currently just a draft to track progress of what's done here, see #190 for what was done in other PRs and what will be done in other PRs.

- [x] Code Cleanup (currently ~50%)
- [ ] Matrix4X4
- [ ] Matrix4X3
- [ ] Matrix4X2
- [ ] Matrix3X4
- [ ] Matrix3X3
- [ ] Matrix3X2
- [ ] Matrix2X3
- [ ] Matrix2X2
- [x] Code generator for `void Func(ref T1 a, T2 b, out TResult result`
- [x] Helper Program to convert `a * b - c / d + e` style expressions to `Scalar` based expressions (may be removed in the future)

Note that a bunch of stuff is broken, and the automatic code cleanup part even ended up removing some (all?) license notices, added regions where they are not supposed to be etc. etc. etc. so this won't reflect what the types will look like in the end.


Perksey:
Yeah I'd imagine changing up the style for the maths project has really fucked some of the actually meaningful settings XD like the license header and `#regions bad` to name a couple

HurricanKai:
yeah, I didn't actually think those settings would be affected, but I believe some of those bad changes also were dotnet-format misinterpreting settings

HurricanKai:
PublicAPI/**/PublicAPI.(Un)shipped.txt is now tracking API surface

HurricanKai:
Will be replaced

PR #229, opened 2020-07-12
Create Tutorial 3.2 for Diffuse Lighting

RhysWhy:
# Summary of the PR

The purpose of this pull request is to create a new tutorial for diffuse lighting to loosely follow the guidelines of chapter 13.2 through 13.5 of learnopengl.com using Silk.NET specific implementations.

This tutorial will be a small incremental addition built on a copy of the 3.1 tutorial, as the ambient lighting is still needed.


examples/CSharp/Tutorial 3.2 - Diffuse Lighting/MathHelper.cs
```diff
@@ -0,0 +1,12 @@
+using System;
+
+namespace Tutorial
+{
+    public static class MathHelper
+    {
+        public static float DegreesToRadians(double degrees)
+        {
+            return (float)(Math.PI / 180f * degrees);
+        }
```
HurricanKai, line 10: Weird that this returns `float` but takes `double`? Any reason for that?

examples/CSharp/Tutorial 3.2 - Diffuse Lighting/Program.cs
```diff
@@ -0,0 +1,226 @@
...
+            Camera = new Camera(Vector3.UnitZ * 6, Vector3.UnitZ * -1, Vector3.UnitY, Width / Height);
+        }
+
+        private static unsafe void OnUpdate(double deltaTime)
+        {
+            var moveSpeed = 2.5f * (float)deltaTime;
+
+            if (primaryKeyboard.IsKeyPressed(Key.W))
+            {
+                //Move forwards
```
HurricanKai, line 138: nit: imo these comments are a bit unnecessary

examples/CSharp/Tutorial 3.2 - Diffuse Lighting/Program.cs
```diff
@@ -0,0 +1,226 @@
...
+            //The Lamp cube is going to be a scaled down version of the normal cubes verticies moved to a different screen location
+            var lampMatrix = Matrix4x4.Identity;
+            lampMatrix *= Matrix4x4.CreateScale(0.2f);
+            lampMatrix *= Matrix4x4.CreateTranslation(LampPosition);
+
+            LampShader.SetUniform("uModel", lampMatrix);
+            LampShader.SetUniform("uView", Camera.GetViewMatrix());
+            LampShader.SetUniform("uProjection", Camera.GetProjectionMatrix());
+
+            Gl.DrawArrays(PrimitiveType.Triangles, 0, 36);
```
HurricanKai, line 188: I don't know GL all that well, but what's going on down here?
I'd suggest you add a comment. (or is this a duplicate?)

examples/CSharp/Tutorial 3.2 - Diffuse Lighting/Program.cs
Perksey: It's probably best to keep them in though as this is a tutorial after all idk

HurricanKai: Doesn't hurt

examples/CSharp/Tutorial 3.2 - Diffuse Lighting/MathHelper.cs
RhysWhy: My mistake, it was originally a double in and out then converted to float, refactoring left it in a weird state so I will fix that

examples/CSharp/Tutorial 3.2 - Diffuse Lighting/Program.cs
RhysWhy: There are two cubes in the scene, each drawn with a different fragment shader (as one cube is being lit while the other cube just represents the light source). I will make this clearer in the code.

PR #232, opened 2020-07-13
Add Time property to Window

### LINKED ISSUE #220: Consider exposing a window Time property
Wrapping over glfwGetTime + SDL equivalent.

Could be added for either 1.0 or 2.0, don't mind which.

RhysWhy:
# Summary of the PR
Attempt to add the Time property to Window class, wrapping around glfwGetTime.

Closes #220


Perksey:
Yeah WindowOptions having a time property is a result of it inheriting from IWindowProperties, which is designed to keep configurable properties on IWindow in-sync with WindowOptions. However, given that Time isn't configurable, it's probably best to move it into the main IWindow interface (although I think this could be promoted to IView as I don't forsee any problems implementing this on mobile)

RhysWhy:
Slightly more involved than the title would make it out to be yeah! I'll do a follow up PR to the tutorials to use this property instead of them tracking it themselves.

PR #239, opened 2020-07-15
Apply Master Changes

HurricanKai:


Perksey:
Done locally due to the windowing changes not applying properly

PR #240, opened 2020-07-15
Improve Tutorial Camera Controls

RhysWhy:
# Summary of the PR
The Camera controls in the tutorial projects are a bit clunky at the moment. They can be improved by swapping the cursor mode to raw input and tweaking the sensitivity settings to give a smoother level of control over movement through the scene.

### DIFF
--- a/examples/CSharp/Tutorial 2.1 - Co-ordinate Systems/MathHelper.cs
+++ b/examples/CSharp/Tutorial 2.1 - Co-ordinate Systems/MathHelper.cs
@@ -1,0 +1,12 @@
+using System;
+
+namespace Tutorial
+{
+    public static class MathHelper
+    {
+        public static float DegreesToRadians(double degrees)
+        {
+            return (float)(Math.PI / 180f * degrees);
+        }
+    }
+}
--- a/examples/CSharp/Tutorial 2.1 - Co-ordinate Systems/MathHelper.cs
+++ b/examples/CSharp/Tutorial 2.1 - Co-ordinate Systems/MathHelper.cs
@@ -4,9 +4,9 @@ namespace Tutorial
 {
     public static class MathHelper
     {
-        public static float DegreesToRadians(double degrees)
+        public static float DegreesToRadians(float degrees)
         {
-            return (float)(Math.PI / 180f * degrees);
+            return MathF.PI / 180f * degrees;
         }
     }
 }
--- a/examples/CSharp/Tutorial 2.2 - Camera/MathHelper.cs
+++ b/examples/CSharp/Tutorial 2.2 - Camera/MathHelper.cs
@@ -1,0 +1,12 @@
+using System;
+
+namespace Tutorial
+{
+    public static class MathHelper
+    {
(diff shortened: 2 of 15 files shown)


RhysWhy:
Note that the above commit leaves the duplication of math helpers and cameras in each project. A later PR with better documentation will aim to centralize these 'unchanging' classes to provide more focus on the specifics of each tutorial without this clutter

examples/CSharp/Tutorial 2.1 - Co-ordinate Systems/MathHelper.cs
```diff
@@ -0,0 +1,12 @@
+using System;
+
+namespace Tutorial
+{
+    public static class MathHelper
+    {
+        public static float DegreesToRadians(double degrees)
```
HurricanKai, line 7: Wasn't there a previous change that made this return double / take float?

examples/CSharp/Tutorial 2.2 - Camera/MathHelper.cs
```diff
@@ -0,0 +1,12 @@
+using System;
+
+namespace Tutorial
+{
+    public static class MathHelper
+    {
+        public static float DegreesToRadians(double degrees)
```
HurricanKai, line 7: See other comment

examples/CSharp/Tutorial 2.2 - Camera/Program.cs
```diff
@@ -28,7 +28,7 @@ class Program
         //Setup the cameras location, directions, and movement speed
         private static Vector3 CameraPosition = new Vector3(0.0f, 0.0f, 3.0f);
         private static Vector3 CameraFront = new Vector3(0.0f, 0.0f, -1.0f);
-        private static Vector3 CameraUp = new Vector3(0.0f, 1.0f, 3.0f);
+        private static Vector3 CameraUp = new Vector3(0.0f, 1.0f, 0.0f);
```
HurricanKai, line 31: Is there no UnitY constant?

examples/CSharp/Tutorial 2.2 - Camera/MathHelper.cs
RhysWhy: That's what I get for not having centralized these yet, i'll fix them all to the float float standard

examples/CSharp/Tutorial 2.2 - Camera/Program.cs
RhysWhy: There is, I hadn't used it to keep it standard with the other two, but I can change this

HurricanKai: Overall I'd say use the UnitX/Y/Z constants as much as possible, doing `-` or `3 * UnitX` > using the ctors

RhysWhy: That is the approach all of the later tutorials take, this one was done specifically to match the contents of the LearnOpenGL tutorial that first introduces cameras so that whoever was following along could see the similarities line by line

PR #243, opened 2020-07-16
Potential extension system for windowing?

Perksey:
# Summary of the PR
So I was just about to start work on SDL stuff when I thought of what could be a pretty neat extension mechanism for windowing. What do you guys think?

Links go here.

# Further Comments

### DIFF
--- a/src/Windowing/Silk.NET.Windowing.Common/Interfaces/IView.cs
+++ b/src/Windowing/Silk.NET.Windowing.Common/Interfaces/IView.cs
@@ -166,5 +166,19 @@ namespace Silk.NET.Windowing.Common
         /// To make all contexts non-current, use <see cref="IWindowPlatform.ClearCurrent"/>.
         /// </summary>
         void ClearContext();
+
+        /// <summary>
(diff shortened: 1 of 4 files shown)


HurricanKai:
I think that's just something interfaces could do, what's the advantage?

Deecellar:
Exactly in what mechanism would extension work, cost and stuff

Perksey:
Essentially the idea is you can specify extra config which the windowing system takes into account when creating a window. It's main use case is for backend-specific functionaliy that we don't expose via the windowing backend directly, such as manual specification of window hints.

In the current form of the windowing system, there's no way to know for sure what backend a window is created with (apart from in 1.0 where glfw is the only windowing backend) so you can specify backend-specific config which is applied if supported, but discarded if not.

HurricanKai:
that can easily be produced by using a combination of Interfaces and the ctor, without bloating the `IView` interface with useless, unsafe functions.

Perksey:
Closing this PR due to indecision and will hold off on introducing any kind of extension mechanism into windowing

PR #263, opened 2020-08-07
Fix Add(ICollection<string>, byte*) adding empty strings

AzyIsCool:
# Summary of the PR
Fixes Add(ICollection<string>, byte*) adding empty strings


AzyIsCool:
Done. Forgot that I had to sign it with my old account too becasue GitKraken did a oops and used my old email for commiting it

Issue #267, opened 2020-08-14
Glfw.Init / glfwInit can fail silently


Tyler-IN:
# Summary
If Glfw.Init fails, no exception is thrown.

# Steps to reproduce
- Platform: Generic Linux Ubuntu Docker Container
- Framework Version: .NET Core 3.1
- API: OpenGL
- API Version: OpenGL 3.2 Core
1. First I did this;
Built a project and set up a runner.
2. Then I did this;
Ran the project under docker.

# Comments

The return value of Init is not checked in the GlfwProvider.
https://github.com/Ultz/Silk.NET/blob/5a539465e371c9df1630c777f416afc77180ee55/src/Windowing/Silk.NET.GLFW/GlfwProvider.cs#L26

**__If you know how to fix this issue, please submit a pull request instead!__**

I'll try to make a PR in a moment and link to this issue.

PR #268, opened 2020-08-14
Throw exception when Glfw.Init fails.

Tyler-IN:
# Summary of the PR
Checks the result of Glfw.Init, retrieves any error and throws instead of continuing.

#267 : Glfw.Init / glfwInit can fail silently

# Further Comments
...


Tyler-IN:
CI failed on "Setup Feed"

Perksey:
Yeah that’s because the Setup Feed task doesn’t have the same if condition the step after it does, will fix.

Issue #265, opened 2020-08-08
API Tracking


HurricanKai:
We should enable the PublicAPI analyzer on *all* projects. This will help tracking (breaking) changes and track nullability behaviour. Would suggest we get this before the final 2.0. Not very high priority, but an easy job but, imo, a big improvement.

HurricanKai:
Assigning this to myself, but if someone else wants to pick this up, feel free to.
Adding to 2.0, but that's not in stone

Perksey:
Not sure if this should be done for the bindings as we're not in control of upstream breaking changes, but perhaps that's something we can discuss on Friday. Defo Windowing and Input at least.

PR #281, opened 2020-08-19
Create CODE_OF_CONDUCT.md

Perksey:
# Summary of the PR
Adds a code of conduct in preparation for Hacktoberfest

Links go here.

# Further Comments

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,76 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as
+contributors and maintainers pledge to making participation in our project and
+our community a harassment-free experience for everyone, regardless of age, body
+size, disability, ethnicity, sex characteristics, gender identity and expression,
+level of experience, education, socio-economic status, nationality, personal
+appearance, race, religion, or sexual identity and orientation.
+
+## Our Standards
+
+Examples of behavior that contributes to creating a positive environment
+include:
+
+* Using welcoming and inclusive language
+* Being respectful of differing viewpoints and experiences
+* Gracefully accepting constructive criticism
+* Focusing on what is best for the community
+* Showing empathy towards other community members
+
+Examples of unacceptable behavior by participants include:
+
+* The use of sexualized language or imagery and unwelcome sexual attention or
+ advances
+* Trolling, insulting/derogatory comments, and personal or political attacks
+* Public or private harassment
+* Publishing others' private information, such as a physical or electronic
+ address, without explicit permission
+* Other conduct which could reasonably be considered inappropriate in a
+ professional setting
+
+## Our Responsibilities
+
+Project maintainers are responsible for clarifying the standards of acceptable
+behavior and are expected to take appropriate and fair corrective action in
+response to any instances of unacceptable behavior.
+
+Project maintainers have the right and responsibility to remove, edit, or
+reject comments, commits, code, wiki edits, issues, and other contributions
+that are not aligned to this Code of Conduct, or to ban temporarily or
+permanently any contributor for other behaviors that they deem inappropriate,
+threatening, offensive, or harmful.
+
+## Scope
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,60 @@
...
+- **DON'T** make PRs that don't directly affect the end user, such as style changes. These are best done as part of a PR related to the area in question. Documentation is fine (and encouraged!), as this is useful to the end user.
+- **DON'T** commit code you didn't write without following its license and following our guidelines for doing so in the Contributor License Agreement. If you are unable to license some code, don't commit it.
+- **DON'T** surprise us with big pull requests or big API changes without talking to us first!
+- **DON'T** make PRs for legal or administrative documents, such as the license, file headers, or code of conduct. If you see something off, let us know and we'll change it.
+
+## Breaking changes
+Contributions must maintain API signature and behavioral compatibility. Contributions that include breaking changes will be rejected. There are exceptions to this, such as if the contribution is made against a major/breaking version branch (such as `2.0` instead of `master`), however this is subject to the team's approval. At risk of sounding like a broken record, talk to us about your idea first!
```
HurricanKai, line 19: #265 really needs to happen for this to be verifiable.
Also, we should probably add that, while binder changes should keep signatures compatible, sometimes upstream changes happen while one is working on the binder.
Imo, in that case it's best to add a comment to the PR, and make a separate PR to update bindings using the current version of the binder.
That way it will can just check that no changes were made to the API Surface, without checking for each check whether it really is an upstream change (which would be error prone)

CONTRIBUTING.md
```diff
@@ -0,0 +1,60 @@
...
+1. [Fork Silk.NET](https://github.com/Ultz/Silk.NET/fork)
+1. Create a branch on your fork.
+1. Add an empty commit to start your work off (and let you open a PR): `git commit --allow-empty -m "start of [thing you're working on]"`
+1. Open a Pull request with `[WIP]` in the title. Do this **before** you actually start working.
+1. Make your commits in small, incremental steps with clear descriptions.
+1. Tag a maintainer when you're done and ask for a review!
+
+## "Help wanted" & "Good first issue"
+If the core team are unable to champion an issue, they will mark it with a "help wanted" label. This indicates that any external contributor may pick up on this issue and implement it as part of a contribution. However, some "help wanted" issues may require intermediate knowledge of the codebase, area, and/or ecosystem; or may have uncertainty surrounding implementation details (if this is the case talk to us in the Discord or in a discussion issue). Because of this, we also mark some issues with the label "good first issue" which indicates that an issue is straightforward, and is a good place to start if you're interested in contributing but new to the codebase.
```
HurricanKai, line 33: Imo there isn't really any point in using the help wanted label. Any issue that doesn't already have a tracking issue can be worked on, right?
We could look into having a separate label that is between intermediate and good first issue

Perksey: The help wanted label is really just "the core silk.net team won't do this without the community's help", whereas most of the other issues i'm sure we will pick up on in time.

Besides, "help wanted" and "good first issue" are recognised by github and pop up in the "pro tips" section when navigating our issues.

HurricanKai: I see, welp then it's fine I guess

CONTRIBUTING.md
Perksey:
```suggestion
Contributions must maintain API signature and behavioral compatibility. Contributions that include breaking changes will be rejected. There are exceptions to this, such as if the contribution is made against a major/breaking version branch (such as `2.0` instead of `master`), however this is subject to the team's approval. At risk of sounding like a broken record, talk to us about your idea first!

### Generated bindings
Sometimes upstream specification changes happen while working on your contribution. Unfortunately, we are not in control of breaking changes created by the upstream spec sources, and if one happens while working on your contribution it is not a cause for concern.

If your contribution affects binder output, let us know by commenting on the pull request. If the change isn't as a result of your modifications to the BuildTools (if any), it is recommended that you don't commit the .gen.cs files and instead do this in a separate PR.
```

PR #293, opened 2020-08-28
New Marshaller API

### LINKED ISSUE #283: Consider using our own marshalling APIs (i.e. our own MarshalAs and UnmanagedType)
This will allow us to allow extra marshalling options (such as UTF8 strings - we're currently using ANSI where we should be using UTF8).

This would consist of the following
- Adding more SilkMarshal methods to support more string encodings (using `System.Text.Encoding`)
- Adding our own `UnmanagedType` enum
- Adding our own `UnmanagedTypeAttribute`, probably shouldn't use `MarshalAs` as that may mistake P/Invoke users who are also using Silk.NET or SilkTouch-powered bindings
- We should keep support for the regular `MarshalAsAttribute`, as this is ubiquitous and forcing our own may break other projects, such as the ClangSharp SilkTouch backend.

HurricanKai:
Closes #283

### DIFF
--- a/src/Core/Silk.NET.SilkTouch/Diagnostics.cs
+++ b/src/Core/Silk.NET.SilkTouch/Diagnostics.cs
@@ -37,5 +37,17 @@ namespace Silk.NET.SilkTouch
             helpLinkUri: IssueLink,
             customTags: WellKnownDiagnosticTags.AnalyzerException
         );
+        
+        public static DiagnosticDescriptor SilkNetCoreMissing { get; } = new DiagnosticDescriptor
+        (
+            id: "ST0003",
+            title: "Silk.NET.Core is missing",
+            messageFormat: "Silk.NET.Core is missing from references. You should use SilkTouch with Silk.NET.Core",
+            category: "SilkTouch.Internal",
+            defaultSeverity: DiagnosticSeverity.Info,
+            isEnabledByDefault: true,
+            description: null,
+            customTags: WellKnownDiagnosticTags.AnalyzerException
+        );
     }
 }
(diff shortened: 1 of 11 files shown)


HurricanKai:
@Perksey I don't think this requires changes to BuildTools? I could not find any reference in BuildTools to `MarshalAs` and only a handful of references overall, in things like OpenAL?

Perksey:
Can't lookover this as I'm on mobile, but as long as you haven't removed the built-in attributes it'll be fine. I'm like 99% sure we don't use any MarshalAs in the BuildTools, but there's some uses of it in OpenAL and GLFW (I think)

Also we should ensure SilkMarshal and UnmanagedTypeAttribute are on-par with the **.NET 5** Marshal and MarshalAsAttribute so that they have a reason for existing. I recommend using the System.Text.Encoding APIs to aid with marshalling, provided you remember to add a null terminator on the end.

HurricanKai:
Not sure what you mean. UnmangedTypeAttribute is much simpler then MarshalAs and only supports what SilkTouch supports.

Also not sure what you mean with System.Text.Encoding?
This PR doesn't add or change any functionality it just uses different attributes.

Perksey:
I don't think there's any point in introducing our own types until our own types exceed what the built-in ones do. In order to do that, we'll have to expand SilkMarshal's APIs to include multiple StringToHGlobal* (et al) methods

HurricanKai:
I agree, but I think it's best to do this in many small steps. This is just laying the foundation to expand possibilities.
Also, this makes it clearer what we actually support.

src/Core/Silk.NET.SilkTouch/Diagnostics.cs
```diff
@@ -37,5 +37,17 @@ internal static class Diagnostics
             helpLinkUri: IssueLink,
             customTags: WellKnownDiagnosticTags.AnalyzerException
         );
+        
+        public static DiagnosticDescriptor SilkNetCoreMissing { get; } = new DiagnosticDescriptor
+        (
+            id: "ST0003",
```
Perksey, line 43: iirc in the other PR you've defined a conflicting ST0003

HurricanKai: Good Catch!

Perksey (approved):

Issue #289, opened 2020-08-24
Generate VTable class


HurricanKai:
Instead of using the current dictionary method, we should just generate a proper class, and get rid of the Load on each call all together.

Perksey:
As long as they're still lazy-loaded and there's still a way to create and exchange function tables I'm fine with this.

HurricanKai:
This will probably go hand in hand with he future abstractions to the native container in regards to switching contexts.

PR #294, opened 2020-08-28
Core Abstractions

### LINKED ISSUE #272: Only allocate GCUtil slot for methods that need it
This would require a special count like `CoreGetSlotCount`, but would mean one dictionary entry less allocated per method that doesn't use GCUtil (in a typical project probably > 2000)
This will also break anyone not using SilkTouch, but I don't believe there is anyone.

HurricanKai:
- massively reduces GCUtil footprint (closes #272)
- abstracts VTable implementation (foundation for #289)

### DIFF
--- a/src/Core/Silk.NET.Core/Native/NativeApiContainer.cs
+++ b/src/Core/Silk.NET.Core/Native/NativeApiContainer.cs
@@ -31,7 +31,7 @@ namespace Silk.NET.Core.Native
                 );
             }
             if (!tableInitialized)
-                _vTable.Init(_ctx, slotCount);
+                _vTable.Initialize(_ctx, slotCount);
             GcUtility = new GcUtility(1, CoreGcSlotCount());
             // ReSharper restore VirtualMemberCallInConstructor
         }
@@ -55,7 +55,7 @@ namespace Silk.NET.Core.Native
         protected IVTable SwapVTable(IVTable newTable, bool initialized = false)
         {
             if (!initialized)
-                newTable.Init(_ctx, CoreGetSlotCount());
+                newTable.Initialize(_ctx, CoreGetSlotCount());
 
             return Interlocked.Exchange(ref _vTable, newTable);
         }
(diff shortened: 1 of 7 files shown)


src/Core/Silk.NET.SilkTouch/MarshalContext.cs
```diff
@@ -241,6 +248,11 @@ public void SetParameterToVariable(int parameter, int variable)
             ParameterVariables[parameter] = variable;
         }
 
+        public int AllocateGcSlot()
```
Perksey, line 251: Consider using expression bodied methods?

src/Core/Silk.NET.SilkTouch/MarshalContext.cs
```diff
@@ -241,6 +248,11 @@ public void SetParameterToVariable(int parameter, int variable)
...
+        {
+            return GCCount++;
```
Perksey, line 253: Are there any threading concerns here?

src/Core/Silk.NET.Core/Native/NativeApiContainer.cs
```diff
@@ -31,18 +30,35 @@ protected NativeApiContainer(INativeContext ctx)
...
+        public IVTable CurrentVTable => _vTable;
+
         public void Dispose()
         {
             _ctx.Dispose();
+            CurrentVTable.Dispose();
         }
 
         protected virtual int CoreGetSlotCount() => 0;
+        protected virtual int CoreGcSlotCount() => 0;
```
Perksey, line 53: Are these values ever different?

src/Core/Silk.NET.SilkTouch/MarshalContext.cs
HurricanKai: No, should be fine.

src/Core/Silk.NET.Core/Native/NativeApiContainer.cs
HurricanKai: The root values are always zero, otherwise they are only the same if every method slot also requires a GC Slot (Pins something via GCUtil)

Perksey: Ahhh right ok.

Perksey (approved):

Perksey:
Approved to merge once conflicts are resolved.

PR #296, opened 2020-08-29
Generate VTable class

### LINKED ISSUE #289: Generate VTable class

HurricanKai:
closes #289

### DIFF
--- a/src/Core/Silk.NET.SilkTouch/NativeApiGenerator.cs
+++ b/src/Core/Silk.NET.SilkTouch/NativeApiGenerator.cs
@@ -166,7 +166,7 @@ namespace Silk.NET.SilkTouch
                         callingConvention: NativeApiAttribute.GetCallingConvention(x.Item3, classNativeApiAttribute))
                 )
                 .ToArray();
-            List<string> slotTypes = new List<string>();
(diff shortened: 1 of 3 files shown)


src/Core/Silk.NET.SilkTouch/NativeApiGenerator.cs
```diff
@@ -346,29 +361,289 @@ void BuildLoadInvoke(ref IMarshalContext ctx, Action next)
...
+                    var name = $"_{entrypoint}";
+                    slotVars.Add(VariableDeclarator(name));
+                    loadStatements.Add
+                    (
+                        IfStatement
+                        (
+                            BinaryExpression
+                            (
+                                SyntaxKind.EqualsExpression, IdentifierName("entryPoint"),
+                                LiteralExpression(SyntaxKind.NumericLiteralExpression, Literal(entrypoint))
```
Perksey, line 439: As discussed in Discord, in the next PR this should do slot comparisons instead of entryPoint comparisons.

Perksey (approved):
Conditionally approved.

PR #302, opened 2020-08-31
WindowOptions will use Resolution of VideoMode

erri120:
`WindowOptions` constructor with existing `ViewOptions` did not use the
Resolution of the `ViewMode` from the `ViewOptions` and defaulted to
1280x720.

### DIFF
--- a/src/Windowing/Silk.NET.Windowing.Common/WindowOptions.cs
+++ b/src/Windowing/Silk.NET.Windowing.Common/WindowOptions.cs
@@ -22,7 +22,7 @@ namespace Silk.NET.Windowing.Common
             IsVisible = true;
             UseSingleThreadedWindow = opts.UseSingleThreadedWindow;
             Position = new Point(50, 50);
-            Size = new Size(1280, 720);
+            Size = opts.VideoMode.Resolution ?? new Size(1280, 720);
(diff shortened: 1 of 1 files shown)


Perksey:
I like the idea, love what you're trying to do!

Personally this isn't how I'd do it because WindowOptions too has a VideoMode property, which won't necessarily be respected.

How I'd do it is use the VideoMode resolution at the implementation level so that it applies to all windows. Also, this change should target the 2.0 branch as this is technically breaking behaviour.

https://github.com/Ultz/Silk.NET/blob/2.0/src/Windowing/Silk.NET.Windowing.Glfw/GlfwWindow.cs#L305-L310
https://github.com/Ultz/Silk.NET/blob/2.0/src/Windowing/Silk.NET.Windowing.Sdl/SdlView.cs#L131-L132

However, the VideoMode property should only be respected for Fullscreen windows. For GLFW:

```cs
(opts.WindowState == Fullscreen ? opts.VideoMode.Resolution : null) ?? opts.Size.Width
```

For SDL:

```cs
((flags & WindowFlags.WindowFullscreen) != 0 ? opts.VideoMode : null) ?? w ?? 1280
```

Perksey:
Will close due to inactivity at the end of the month.

PR #303, opened 2020-09-01
AOT Support?

HurricanKai:
This PR includes the barebones support for AOT [CoreRT](https://github.com/dotnet/corert) and modifies the sample in Labs to function.

Changes include:
- Window & InputWindowExtension changes to allow disabling of the default first party support. (This also means platforms are now lazy loaded)
- 1 new Type with one function in each of SDL/GLFW input & windowing that adds the platform. (These are the only visible types in that namespace, not sure if desirable)
- Adds .net 5 as build target to all projects, this means .net 5 will propagate through projects and dependencies, making it possible to enable features on .net 5 and higher only. This has been long overdue imo.


Perksey:
I wouldn't set projects to multi target unless it actually has .NET 5-specific enhancements. NuGet is smart enough to get a .NET 5 version of a package even if the referencing package only targets .NET Standard 2.0, but the dependency cross-targets.

src/Core/Silk.NET.Core/Loader/LibraryLoader.cs
```diff
@@ -3,7 +3,7 @@
 // You may modify and distribute Silk.NET under the terms
 // of the MIT license. See the LICENSE file for details.
 
-#if NETCOREAPP3_0
```
Perksey, line 6: Why target 5.0 when we can target 3.0 and make more users happy?

HurricanKai:
Didn't know NuGet was that smart. I'd preferably keep all the multi-targeting to allow the possibility for the future. I plan on optimizing a lot of stuff with the awesome new features.
If you prefer, I'll remove the multi-targeting here, and re-add it once needed?

Perksey:
> If you prefer, I'll remove the multi-targeting here, and re-add it once needed?

This sounds like the preferred way to go.

src/Core/Silk.NET.Core/Loader/LibraryLoader.cs
HurricanKai: won't net5.0 users be unhappy then?

Perksey: `net5.0` == `netcoreapp5.0` remember? `netcoreapp3.0`'s TFM means any version above `3.0`, including .NET 5.

HurricanKai: Ok, thanks. Rider doesn't understand this yet.

src/Windowing/Silk.NET.Windowing.Common/Window.cs
```diff
@@ -189,10 +213,8 @@ public static void PrioritizeSdl()
         /// <param name="platform">The platform to add.</param>
         public static void Add(IWindowPlatform platform)
         {
-            if (!((List<IWindowPlatform>) Platforms).Contains(platform))
-            {
```
Perksey, line 193: Code style change isn't needed, recommended revert (generally I've operated under the assumption that code style is up to the decision of the codeowner, and I personally prefer braces for verbosity and explicitness)

src/Windowing/Silk.NET.Windowing.Common/Structs/WindowOptions.cs
```diff
@@ -120,26 +121,63 @@ public WindowOptions
             VSync = isVSync;
         }
 
+        private static Lazy<WindowOptions> _default = new Lazy<WindowOptions>
```
Perksey, line 124: Why lazies? This is probably more memory noise than just the WindowOptions.

src/Windowing/Silk.NET.Windowing.Common/Structs/WindowOptions.cs
```diff
@@ -120,26 +121,63 @@ public WindowOptions
...
+        private static Lazy<WindowOptions> _vulkanDefault = new Lazy<WindowOptions>(
+            () => new WindowOptions
+            (
+                true, new Point(50, 50),
+                new Size(1280, 720), 0.0, 0.0, GraphicsAPI.DefaultVulkan,
+                Assembly.GetEntryAssembly()?.GetName().Name ?? "Silk.NET Window", WindowState.Normal,
+                WindowBorder.Resizable, false, false, VideoMode.Default
+            ));
+
+        private static Lazy<WindowOptions> _aotDefault = new Lazy<WindowOptions>
```
Perksey, line 143: Separate AOT defaults likely aren't needed. If CoreRT reflection is disabled, GetEntryAssembly will return null, thus will default to Silk.NET Window on the regular properties.

src/Windowing/Silk.NET.Windowing.Common/Internals/ViewImplementationBase.cs
```diff
@@ -210,7 +210,7 @@ public Point PointToFramebuffer(Point point)
             // TODO this monstrosity will be gone once Silk.NET.Maths is in
             if (Vector.IsHardwareAccelerated && Vector<int>.Count >= 2)
             {
-#if NETSTANDARD2_1
+#if NETSTANDARD2_1 || NET5_0
```
Perksey, line 213: Remove NET5_0

src/Lab/CoreRTTest/Program.cs
```diff
@@ -0,0 +1,275 @@
...
+        {
+            0, 1, 3,
+            1, 2, 3
+        };
+
+        private static void Main()
+        {
+            // setup window
+            Window.ShouldLoadFirstPartyPlatforms
+                (false); // Window will try to discover platforms via reflection, AOT cannot support this
```
Perksey, line 94: Is this needed? It'll fail silently if it can't load first-party platforms. If it is needed, this should be a property, but I honestly don't think it is.

src/Lab/CoreRTTest/Program.cs
```diff
@@ -0,0 +1,275 @@
...
+
+            // add first party platforms manually
+            Silk.NET.Windowing.Glfw.WindowExtensions.AddGLFW();
+            Silk.NET.Windowing.Sdl.WindowExtensions.AddSDL();
```
Perksey, line 98:
```suggestion
            // add first party platforms manually
            Silk.NET.Windowing.Glfw.GlfwWindowing.RegisterPlatform();
            Silk.NET.Windowing.Sdl.SdlWindowing.RegisterPlatform();
```

src/Lab/CoreRTTest/Program.cs
```diff
@@ -0,0 +1,275 @@
...
+            
+            // the same for input:
+            InputWindowExtensions.ShouldLoadFirstPartyPlatforms(false);
```
Perksey, line 101: Same as with windowing.

src/Lab/CoreRTTest/Program.cs
```diff
@@ -0,0 +1,275 @@
...
+            
+            Silk.NET.Input.Glfw.InputExtensions.AddGlfw();
+            Silk.NET.Input.Sdl.InputExtensions.AddSdl();
```
Perksey, line 104:
```suggestion
            Silk.NET.Input.Glfw.GlfwInput.RegisterPlatform();
            Silk.NET.Input.Sdl.SdlInput.RegisterPlatform();
```

src/Lab/CoreRTTest/Program.cs
```diff
@@ -0,0 +1,275 @@
...
+
```
Perksey, line 105: Here comes a thought: we could add a catch-all `Silk.NET.Blah.BlahPlatform.RegisterPlatform();` in the input assemblies which register both windowing and input (as the windowing internals are visible to input)

Perksey (changes requested):

src/Windowing/Silk.NET.Windowing.Common/Structs/WindowOptions.cs
HurricanKai: GetEntryAssembly & friends throw in CoreRT, so the static constructor of WindowOptions throws if this is not lazy.

src/Windowing/Silk.NET.Windowing.Common/Structs/WindowOptions.cs
HurricanKai: GetEntryAssembly & friends throw in CoreRT, so the static constructor of WindowOptions throws if this is not lazy.
I can remove the AOT defaults.

src/Lab/CoreRTTest/Program.cs
HurricanKai: It won't fail silently, It'll throw at TryAdds `Assembly.Load(assemblyName);`

src/Lab/CoreRTTest/Program.cs
HurricanKai: good idea! but I'd also like to keep the Window methods, since there's otherwise no way to use Windowing without input.

Perksey: Oh yeah of course.

src/Lab/CoreRTTest/Program.cs
Perksey: Ah. In which case we should probably add a try-catch in TryAdd

src/Windowing/Silk.NET.Windowing.Common/Structs/WindowOptions.cs
Perksey: Should probably prefer try-catching the assembly name assignment then, as this isn't the best solution.

src/Windowing/Silk.NET.Windowing.Common/Window.cs
HurricanKai: This is to make the `Add` call from within `DoLoadFirstPartyPlatformsViaReflection` not recursive.

Perksey: All you did is remove the brackets and use `_platforms` instead of casting?

HurricanKai: Yes

Perksey: Ok well can you add the brackets back.

HurricanKai: OH, sorry I didn't understand 😅 
of course.

Perksey: Thanks haha

src/Input/Silk.NET.Input.Sdl/SdlInput.cs
```diff
@@ -0,0 +1,19 @@
...
+// of the MIT license. See the LICENSE file for details.
+
+using Silk.NET.Windowing;
+using Silk.NET.Windowing.Sdl;
+
+namespace Silk.NET.Input.Sdl
+{
+    public static class SdlInput
+    {
+        public static void RegisterWindow()
```
Perksey, line 13:
```suggestion
        public static void RegisterPlatform()
```

src/Lab/CoreRTTest/Program.cs
```diff
@@ -0,0 +1,268 @@
...
+        private static readonly uint[] Indices =
+        {
+            0, 1, 3,
+            1, 2, 3
+        };
+
+        private static void Main()
+        {
+            Silk.NET.Input.Glfw.GlfwInput.RegisterPlatform();
+            Silk.NET.Input.Sdl.SdlInput.RegisterWindow();
```
Perksey, line 93:
```suggestion
            Silk.NET.Input.Sdl.SdlInput.RegisterPlatform();
```

src/Windowing/Silk.NET.Windowing.Common/Structs/WindowOptions.cs
```diff
@@ -120,26 +121,52 @@ public WindowOptions
...
+                    WindowBorder.Resizable, true, true, VideoMode.Default
+                );
+            }
+            catch
+            {
+                
+#if DEBUG
+                Console.WriteLine("Could not initialize WindowOptions.Default");
+#endif
+                Default = default;
```
Perksey, line 141: Please don't return default.

src/Windowing/Silk.NET.Windowing.Common/Structs/WindowOptions.cs
```diff
@@ -120,26 +121,52 @@ public WindowOptions
...
+                    Assembly.GetEntryAssembly()?.GetName().Name ?? "Silk.NET Window", WindowState.Normal,
+                    WindowBorder.Resizable, false, false, VideoMode.Default
+                );
+            }
+            catch
+            {
+#if DEBUG
+                Console.WriteLine("Could not initialize WindowOptions.DefaultVulkan");
+#endif
+                DefaultVulkan = default;
```
Perksey, line 158: Please don't return default.

Perksey (changes requested):
All of the unanswered comments from above, and...

src/Windowing/Silk.NET.Windowing.Common/Structs/WindowOptions.cs
HurricanKai: What else should I return? Just some other arbitrary settings?

Perksey: The default settings we'd normally return, only with a default title as getting the title is the only thing that can throw in this.

HurricanKai:
Rebased onto origin/2.0 to get the RC1 changes. Git kinda died so I had to force push 😅

src/Windowing/Silk.NET.Windowing.Common/Structs/WindowOptions.cs
```diff
@@ -120,26 +122,40 @@ public WindowOptions
             VSync = isVSync;
         }
 
+        static WindowOptions()
+        {
+            string name = "Silk.NET Window";
+            try
+            {
+                var assName = Assembly.GetEntryAssembly()?.GetName().Name;
```
Perksey, line 130: lmfao assName huehuehue he said ass

(generally i prefer `asm` instead of `ass`)

src/Windowing/Silk.NET.Windowing.Glfw/GlfwPlatform.cs
```diff
@@ -30,8 +31,11 @@ public bool IsApplicable
                 {
                     GLFW.Glfw.GetApi(); // activate the class so we can determine if we can activate the class
                 }
-                catch
+                catch (Exception ex)
                 {
+#if DEBUG
+                    Console.WriteLine($"Can't load SDL: {ex}");
```
Perksey, line 37: Should be GLFW?

src/Core/Silk.NET.Core/Loader/LibraryLoader.cs
```diff
@@ -304,7 +304,7 @@ public void FreeNativeLibrary(IntPtr handle)
         public static LibraryLoader GetPlatformDefaultLoader()
         {
 #if NETCOREAPP3_0
```
Perksey, line 306: If we're upgrading to NETCOREAPP3_0, you'll need to replace all usage of this define too.

TFMs of `netstandard2.0;netcoreapp3.1` only define:
- NETSTANDARD
- NETSTANDARD2_0
- NETCOREAPP
- NETCOREAPP3_1

Perksey (changes requested):

src/Core/Silk.NET.Core/Loader/LibraryLoader.cs
HurricanKai: Wait, now I'm realllly confused.
I thought Higher tfm means it defines everything below it. Ie net5.0 defines both NET5_0 and NETCOREAPP3_1 and everything below

HurricanKai: Just tested it, and you are right. for some reason NETCOREAPP3_1 is defined by net5.0 but NETCOREAPP3_0 isn't defined by NETCOREAPP3_1

Perksey: Yeah, it doesn't define prior versions, only the versions you define in the target framework.

build/props/common.props
```diff
@@ -24,13 +24,13 @@
         </Description>
     </PropertyGroup>
     <!-- SourceLink -->
-    <PropertyGroup>
+    <PropertyGroup Condition="'$(EnableSourceLink)' == ''">
```
Perksey, line 27: I think this may conflcit with an existing (or proposed) SDK property, so we should probably prefix it with Silk.

Perksey (approved):

PR #312, opened 2020-09-20
Update to .Net 5 RC1

HurricanKai:
This PR would require RC1+ to be installed for the project to build, because SilkTouch does.


src/Core/Silk.NET.SilkTouch/NativeApiGenerator.cs
```diff
@@ -986,9 +1002,10 @@ private static string GetCallingConvention(CallingConvention convention)
             => convention switch
             {
                 // CallingConvention.Winapi => "", netstandard2.0 doesn't allow this
```
Perksey, line 1004: Don't forget to do a manual reimplementation of this at some point in the near future!

Perksey (approved):

PR #319, opened 2020-10-02
Add BorderSize, GetFullSize(), and Centre() for IWindow

Perksey:
# Summary of the PR
Said it all in the title

https://discordapp.com/channels/521092042781229087/607634593201520651/756873490212192325

# Further Comments
Also updates release notes for 1.7.

### DIFF
--- a/src/Windowing/Silk.NET.Windowing.Common/Interfaces/IWindow.cs
+++ b/src/Windowing/Silk.NET.Windowing.Common/Interfaces/IWindow.cs
@@ -34,6 +34,17 @@ namespace Silk.NET.Windowing.Common
         new bool IsClosing { get; set; }
         
         /// <summary>
+        /// Gets the distances in screen coordinates from the edges of the content area to the corresponding edges of
+        /// the full window.
+        /// </summary>
+        /// <remarks>
+        /// Because these are distances and not coordinates, they are always zero or positive.
+        /// </remarks>
+        /// <seealso cref="WindowExtensions.GetFullSize"/>
+        // This is in IWindow because it can't be configured.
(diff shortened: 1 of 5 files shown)


src/Windowing/Silk.NET.Windowing.Common/Interfaces/IWindow.cs
```diff
@@ -33,6 +33,17 @@ public interface IWindow : IWindowProperties, IWindowHost, IView
...
+        /// <summary>
+        /// Gets the distances in screen coordinates from the edges of the content area to the corresponding edges of
+        /// the full window.
+        /// </summary>
+        /// <remarks>
+        /// Because these are distances and not coordinates, they are always zero or positive.
+        /// </remarks>
+        /// <seealso cref="WindowExtensions.GetFullSize"/>
+        // This is in IWindow because it can't be configured.
+        Rectangle BorderSize { get; }
```
HurricanKai, line 45: This is, theoretically, a breaking change. Just noting, I think it's probably fine,

HurricanKai (approved):

src/Windowing/Silk.NET.Windowing.Common/Interfaces/IWindow.cs
Perksey: What would it break?

HurricanKai: Theoretically `IWindow` is a public interface, and someone may have mocked/implemented it themselves (best case I can think of is for Unit testing) and this would break their code.

PR #321, opened 2020-10-04
Improve readability of Examples

radobot:
# Summary of the PR
A bunch of little fixes that improve the readability of Examples.

I would also change `float[] Vertices` and `uint[] Indices` (and possibly others) to twodimensional arrays to better visualise their meaning:
```C#
private static readonly float[,] Vertices =
{
	// X      Y      Z     R  G  B  A
	{  0.5f,  0.5f,  0.0f, 1, 0, 0, 1 },
	{  0.5f, -0.5f,  0.0f, 0, 0, 0, 1 },
	{ -0.5f, -0.5f,  0.0f, 0, 0, 1, 1 },
	{ -0.5f,  0.5f,  0.5f, 0, 0, 0, 1 }
};

private static readonly uint[,] Indices =
{
	{ 0, 1, 3 },
	{ 1, 2, 3 }
};

//then this still works with multidimensional arrays
fixed (float* v = Vertices)
{
	// ...
}
```
But that can be done only in Tutorial 1.2 because in others they are cast to `Span<T>` which works only with singledimensional arrays. I'm not sure if it would be a good idea to make this change in only one tutorial.


examples/CSharp/Tutorial 1.1 - Hello Window/Program.cs
```diff
@@ -10,10 +10,10 @@ class Program
     {
         private static IWindow window;
 
-        private static void Main(string[] args)
+        private static void Main()
         {
             //Create a window.
-            var options = WindowOptions.Default;
+            WindowOptions options = WindowOptions.Default;
```
Perksey, line 16: Using var is a stylistic choice throughout the codebase and it's something we'd like to keep around.

examples/CSharp/Tutorial 1.2 - Hello quad/Program.cs
```diff
@@ -136,7 +136,7 @@ private static unsafe void OnLoad()
             Gl.LinkProgram(Shader);
 
             //Checking the linking for errors.
-            Gl.GetProgram(Shader, GLEnum.LinkStatus, out var status);
+            Gl.GetProgram(Shader, GLEnum.LinkStatus, out int status);
```
Perksey, line 139: Same here, use `var` wherever possible.

examples/CSharp/Tutorial 1.3 - Abstractions/VertexArrayObject.cs
```diff
@@ -28,7 +28,7 @@ public VertexArrayObject(GL gl, BufferObject<TVertexType> vbo, BufferObject<TInd
         public unsafe void VertexAttributePointer(uint index, int count, VertexAttribPointerType type, uint vertexSize, int offSet)
         {
             //Setting up a vertex attribute pointer
-            _gl.VertexAttribPointer(index, count, type, false, vertexSize * (uint) sizeof(TVertexType), (void*) (offSet * sizeof(TVertexType)));
+            _gl.VertexAttribPointer(index, count, type, false, vertexSize * (uint) sizeof(TVertexType), (TVertexType*) (offSet * sizeof(TVertexType)));
```
Perksey, line 31: This is not a pointer to TVertexType.

examples/CSharp/Tutorial 2.1 - Co-ordinate Systems/Program.cs
```diff
@@ -128,11 +128,11 @@ private static unsafe void OnRender(double obj)
...
             //Use elapsed time to convert to radians to allow our cube to rotate over time
-            var difference = (float) (window.Time * 100);
+            float difference = (float) (window.Time * 100);
 
-            var model = Matrix4x4.CreateRotationY(MathHelper.DegreesToRadians(difference)) * Matrix4x4.CreateRotationX(MathHelper.DegreesToRadians(difference));
-            var view = Matrix4x4.CreateLookAt(CameraPosition, CameraTarget, CameraUp);
-            var projection = Matrix4x4.CreatePerspectiveFieldOfView(MathHelper.DegreesToRadians(45.0f), Width / Height, 0.1f, 100.0f);
+            Matrix4x4 model = Matrix4x4.CreateRotationY(MathHelper.DegreesToRadians(difference)) * Matrix4x4.CreateRotationX(MathHelper.DegreesToRadians(difference));
+            Matrix4x4 view = Matrix4x4.CreateLookAt(CameraPosition, CameraTarget, CameraUp);
+            Matrix4x4 projection = Matrix4x4.CreatePerspectiveFieldOfView(MathHelper.DegreesToRadians(45.0f), Width / Height, 0.1f, 100.0f);
```
Perksey, line 135: Use var.

examples/CSharp/Tutorial 3.1 - Ambient Lighting/Camera.cs
```diff
@@ -38,7 +38,7 @@ public void ModifyDirection(float xOffset, float yOffset)
             //We don't want to be able to look behind us by going over our head or under our feet so make sure it stays within these bounds
             Pitch = Math.Clamp(Pitch, -89f, 89f);
 
-            var cameraDirection = Vector3.Zero;
+            Vector3 cameraDirection = Vector3.Zero;
```
Perksey, line 41: Use var.

Perksey (changes requested):
TL;DR Use `var` and fixup some pointer types - some are intentionally `void*` as those parameters are essentially `IntPtr`, but upstream defines it as `void*`.

Perksey:
> What about fixed (float* v = Vertices)? Should that be void* too?

Nah, only the two places I outlined need to be `void*` because they're not a pointer, rather a pointer-sized integer. Everywhere else can be fine strongly-typed.

> There are places where there wasn't var before. How do I know which ones to change to var and which ones to keep explicit type? Or should I change them all to var everywhere?

Yeah, use var wherever possible. The only places where vars can't be used are for span stackallocs, weak-typed enumerators, or implicit casts.

Perksey:
Sure.

Perksey:
I’d say suggestion.

Perksey:
Is this ready for re-review?

Perksey:
Will close due to inactivity this Friday.

PR #331, opened 2020-10-16
Fixup String marshalling

HurricanKai:
Also introduces a constant for every allocated slot.
Also made me realize that I just made a quick patch to lower Slot numbers, but the slots are actually still allocated internally, just not at runtime & in memory.


HurricanKai:
of course CI uncovers more stuff...

HurricanKai:
Removed the constants stuff. Debuggability can suck it

Perksey (approved):

PR #336, opened 2020-10-23
Load override proposal

HurricanKai:
Blocking P/Invoke.
Copied proposal to below for convenience.

# Summary
Load overrides mean a way to override loading native addresses.
This is useful in the following scenarios:
- P/Invoke / statically linked scenarios, mobile.
- Advanced Performance scenarios, such as pre-loading to improve AOT / linking / R2R

While this API can be used by Hand and is available for public use it is not designed to be particularly user friendly.
It is designed with SilkTouch as the primary user. (See `PInvokeTableAttribute`)

This proposal also somewhat defines the currently loose timeline of how addresses are loaded.

# Contributors
- Kai Jellinghaus, Maintainer (at time of writing), open source community.

# Current Status
- [x] Proposed
- [ ] Discussed with API Review Board (ARB)
- [ ] Approved
- [ ] Implemented

# Design Decisions
There are two separate interfaces to allow maximum flexibility, but also to define exactly at which point slot information is lost.

# Proposed API
Note that this API may be changed in the future to use function pointers directly instead of IntPtrs.
This deliberately doesn't use nint.
```cs
/// <summary>
/// Defines a Load override, capable of loading some entrypoints by name.
/// </summary>
/// <remarks>
/// Load overrides may not cache loads.
/// Load overrides have to be thread safe.
/// </remarks>
public interface INameLoadOverride
{
    /// <summary>
    /// The <see cref="INativeContext"/> used for loading
    /// </summary>
    INativeContext Context { set; }

    /// <summary>
    /// Attempt to load an entrypoint by name.
    /// </summary>
    /// <param name="entrypoint">The entrypoint name to load</param>
    /// <param name="address">The address that was loaded</param>
    /// <returns>
    /// Whether the load was successful.
    /// </returns>
    bool TryLoad(string entrypoint, out IntPtr address);
}

/// <summary>
/// Defines a Load override, capable of loading some entrypoints by slot.
/// </summary>
/// <remarks>
/// Load overrides may not cache loads.
/// Load overrides have to be thread safe.
/// </remarks>
public interface ISlotLoadOverride
{
    /// <summary>
    /// The <see cref="INativeContext"/> used for loading
    /// </summary>
    INativeContext Context { set; }
        
    /// <summary>
    /// Attempt to load an entrypoint by name.
    /// </summary>
    /// <param name="slot">The slot to load</param>
    /// <param name="address">The address that was loaded</param>
    /// <returns>
    /// Whether the load was successful.
    /// This will flush the cache / VTable.
    /// </returns>
    bool TryLoad(int slot, out IntPtr address);
}
```

Not a critical change, but would make sense to make the core easier to understand:
```diff
- public interface INativeContext
+ public interface IAddressLoader
 : IDisposable
{
-    IntPtr GetProcAddress(string proc);
+    IntPtr GetProcAddress(string entrypoint);
}
```

```cs
public abstract class NativeApiContainer
{
    /// <summary>
    /// Registers an <see cref="INameLoadOverride"/> to participate in loading.
    /// </summary>
    /// <remarks>
    /// This method is thread safe.
    /// This will flush the cache / VTable.
    /// </remarks>
    void RegisterOverride(INameLoadOverride override);
    /// <summary>
    /// Registers an <see cref="ISlotLoadOverride"/> to participate in loading.
    /// </summary>
    /// <remarks>
    /// This method is thread safe.
    /// This will flush the cache / VTable.
    /// </remarks>
    void RegisterOverride(ISlotLoadOverride override);
    /// <summary>
    /// Registers multiple <see cref="INameLoadOverride"/>s to participate in loading.
    /// </summary>
    /// <remarks>
    /// This method is thread safe.
    /// This will flush the cache / VTable.
    /// </remarks>
    void RegisterOverrides(IEnumerable<INameLoadOverride> overrides);
    /// <summary>
    /// Registers multiple <see cref="ISlotLoadOverride"/>s to participate in loading.
    /// </summary>
    /// <remarks>
    /// This method is thread safe.
    /// This will flush the cache / VTable.
    /// </remarks>
    void RegisterOverrides(IEnumerable<ISlotLoadOverride> overrides);

    // override address loader is thread safe.
    private sealed class OverrideAddressLoader
    {
        // this is loader falls back to the normal _loader, but first calls into overrides.
        // an instance is kept around and recreated whenever a new override is registered.
        // for performance reasons it may make sense to defer the creation of this to a Lazy<OverrideAddressLoader>
    
        public IntPtr Load(int slot, string entrypoint);
    }
}
```

Load overrides MAY NOT cache addresses.

These APIs work together with the IVTable, which is the caching solution.

# Address loading timeline
- NativeApiContainer.Load(int, string)
  - VTable
    - if cached, return the cached result
    - otherwise
        - OverrideAddressLoader.Load(int, string)
            - calls slot overrides in order, returning the first result.
            - if none were able to resolve the slot, calls entrypoint overrides in order, returning the first result.
            - if none were able to resolve the entrypoint, calls down to the given fallback loader.
        - result cached.

# Open Questions
- it's unclear how this would work with VTable preloading, we don't use that in Silk.NET right now, so not a problem (for now).
- it's unclear how a user that manually swaps VTables (think Vulkan) would handle it if another user registered an override. there is no way to detect this. Likely, such a user would just not handle this.


documentation/proposals/Proposal - Load overrides.md
```diff
@@ -0,0 +1,157 @@
...
+    /// This will flush the cache / VTable.
+    /// </returns>
+    bool TryLoad(int slot, out IntPtr address);
+}
+```
+
+Not a critical change, but would make sense to make the core easier to understand:
+```diff
+- public interface INativeContext
++ public interface IAddressLoader
```
Perksey, line 84: Personally would like to vito on this name change, feel free to add docs to accomplish better understanding.

documentation/proposals/Proposal - Load overrides.md
```diff
@@ -0,0 +1,157 @@
...
+ : IDisposable
+{
+-    IntPtr GetProcAddress(string proc);
++    IntPtr GetProcAddress(string entrypoint);
```
Perksey, line 88: ...as follows:
```diff
-    IntPtr GetProcAddress(string proc);
+    // IMPLEMENTATION DETAIL: Must not throw, upon failure should just return default.
+    // NOTE: I'm not sure whether this can throw in today's codebase, but it shouldn't.
+    IntPtr GetProcAddress(string entrypoint);
+    // NOTE: Consider using a default interface implementation that returns default, as not all native contexts will support default implementations.
+    // IMPLEMENTATION DETAIL: Must not throw whatsoever, upon failure should just return default.  
+    IntPtr GetProcAddress(int wellKnownFunctionAddress);
```

Perksey (commented):
Not sure about this one. This seems to overcomplicate a lot, whereas really I would just make the P/Invoke mechansim use `INativeContext` and maybe modify that interface as follows:

documentation/proposals/Proposal - Load overrides.md
HurricanKai: What would the second overload be used for?

Perksey: "wellKnownFunctionAddress" is the slot.

HurricanKai: one thing I deliberately tried to avoid was allowing the INativeContext/IAddressLoader to know about slots. they are, imho, a purely VTable related mechanism.

documentation/proposals/Proposal - Load overrides.md
Perksey: As in, I'd rather not rename the class.

HurricanKai: I've added some more explanation, but I don't really mind this either, just figured it would fit in 🙂

documentation/proposals/Proposal - Load overrides.md
Perksey: Today they are, yes. However, P/Invoke loading is best suited as a native context as native contexts are the prime source of function pointers. If you'd like to add the ability to load function pointers using an integer, you will need to add it to the native context which I'm not against.

Again, in the future users could use well-known IDs used by their own native contexts:
```cs
[WellKnownFunction(7536)]
public partial void Abc(Xyz xyz);
```

HurricanKai: If we add slot information into INativeContext this proposal doesn't really make sense though, could just use MultiNativeContext. Which was my first thought, but I didn't really think putting slot info into the INativeContext was a good idea.

Perksey: Yeah to be honest adding the ability for "well known numbers" (such as slots) would make it a lot easier. What would probably be best is:
- Add the overload I proposed to INativeContext
- Add a new method to NativeApiContext called CreateDefaultContext which takes a string.
- CreateDefaultContext will:
    - if the string passed in matches the P/Invoke library name, a P/Invoke native context
    - otherwise, an instance of DefaultNativeContext
- Find some way to discourage and/or hide the constructor of DefaultNativeContext

Perksey:
Generally rejected proposals are kept in the repo.

Perksey (approved):
"Approved"

PR #338, opened 2020-10-24
P/Invoke capabilities in SilkTouch

HurricanKai:


src/Core/Silk.NET.Core/Native/PInvokeOverride.cs
```diff
@@ -0,0 +1,17 @@
...
+// 
+// You may modify and distribute Silk.NET under the terms
+// of the MIT license. See the LICENSE file for details.
+
+using System;
+
+namespace Silk.NET.Core.Native
+{
+    [AttributeUsage(AttributeTargets.Class)]
+    public class PInvokeOverride : Attribute
```
Perksey, line 11: You have a choice: document this now or later because between preview 5 and 2.0 we'll be having a massive documentation sprint.

Perksey (commented):
Looks good for the most part, just a couple of comments.

src/Core/Silk.NET.Core/Native/PInvokeOverride.cs
HurricanKai: I'll do it when I document SilkTouch + Core, don't really have the time to think about how to explain all this best.

Perksey (approved):

PR #340, opened 2020-10-24
[2.0] The Wednesday round-up pull request, now on a Saturday.

### LINKED ISSUE #163: Consider updating how fixed-sized buffers for non-primitive types are bound
Fixed sized buffers in C# only support a limited subset of types today: `bool`, `byte`, `char`, `short`, `int`, `long`, `sbyte`, `ushort`, `uint`, `ulong`, `float`, or `double`.

While there exists proposals for extending this to support other types as well, that support does not currently exist. This can make it difficult to work with certain structures such as [D3D12_RT_FORMAT_ARRAY](https://docs.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_rt_format_array):
```cpp
struct D3D12_RT_FORMAT_ARRAY {
  DXGI_FORMAT RTFormats[8];
  UINT        NumRenderTargets;
};
```

Today, Silk.NET is generating an inline array of fields for these (similar to and based on https://discordapp.com/channels/521092042781229087/587346162802229298/696144071164428348):
```csharp
struct D3D12_RT_FORMAT_ARRAY
{
  public DXGI_FORMAT RTFormats_0;
  public DXGI_FORMAT RTFormats_1;
  public DXGI_FORMAT RTFormats_2;
  public DXGI_FORMAT RTFormats_3;
  public DXGI_FORMAT RTFormats_4;
  public DXGI_FORMAT RTFormats_5;
  public DXGI_FORMAT RTFormats_6;
  public DXGI_FORMAT RTFormats_7;
  public uint NumRenderTargets;
}
```

Instead, it may be possible to provide a "nicer" API by abusing some of the unsafe framework APIs:
```csharp
public partial struct D3D12_RT_FORMAT_ARRAY
{
    [NativeTypeName("DXGI_FORMAT [8]")]
    public _RTFormats_e__FixedBuffer RTFormats;

    [NativeTypeName("UINT")]
    public uint NumRenderTargets;

    public partial struct _RTFormats_e__FixedBuffer
    {
        internal DXGI_FORMAT e0;
        internal DXGI_FORMAT e1;
        internal DXGI_FORMAT e2;
        internal DXGI_FORMAT e3;
        internal DXGI_FORMAT e4;
        internal DXGI_FORMAT e5;
        internal DXGI_FORMAT e6;
        internal DXGI_FORMAT e7;

        public ref DXGI_FORMAT this[int index] => ref AsSpan()[index];

        public Span<DXGI_FORMAT> AsSpan() => MemoryMarshal.CreateSpan(ref e0, 8);
    }
}
```

This is abusing the `MemoryMarshal.CreateSpan` API to create a `Span` around the entire buffer and bypasses the normal C# language safety rules which would prevent this from occuring.

Because it is normally illegal C#, it is not recommended to expose the `ref` or `Span` as part of a "safe" API surface and it would be recommended to provide an analyzer that helps remind users that storing the `ref`/`Span` will not be tracked appropriately by the language and the `ref`/`Span` could outlive the actual backing struct if not managed appropriately.

It's also worth noting that the language could get proper support for this eventually, in which case the exposed signature may need to change slightly or in an incompatible way (depending on if they require any `modreq`/`modopt`, etc).

### LINKED ISSUE #189: Support creating windows from existing handles
# Summary of feature
As far as I can tell Silk.NET insists on owning all window management itself. This makes it unsuitable for integration into other applications using toolkits such as Windows Forms, WPF or similar.

I am not requesting that you implement support for all possible toolkits out there. However, if there was a way to point Silk.NET to an existing HWND (and similar abstractions for other windowing systems) and use that for rendering it would go a long way.

### LINKED ISSUE #214: Generate proper constructors for fixed-size arrays
# Summary
as you can see here
image
the ctor for `ClearColorValue` does not work with float values. 
this stems from the fact that some of the fields occupy the the same location, but get initialized with, for example `uint_0 = default`, _AFTER_ other fields at the same location, essentially overriding them with `default`.
for me it works when i use the last fields (the uint ones) but for everything else the values get zeroed out.

# Steps to reproduce
- Platform: Desktop
i don't know if this is platform specific, i'm not that familiar with the net core spec

# _A_ solution
 would be to split up the ctor into separate ones that initialize base and then override the correct fields:
```C#
 public ClearColorValue(float float32_0 = default, float float32_1 = default, float float32_2 = default, float float32_3 = default) 
  : this() {
  Float32_0 = float32_0;
  Float32_1 = float32_1;
  Float32_2 = float32_2;
  Float32_3 = float32_3;
 }
public ClearColorValue(int int32_0 = default, int int32_1 = default, int int32_2 = default, int int32_3 = default) 
 : this() {
  Int32_0 = int32_0;
  Int32_1 = int32_1;
  Int32_2 = int32_2;
  Int32_3 = int32_3;
}
public ClearColorValue(uint uint32_0 = default, uint uint32_1 = default, uint uint32_2 = default, uint uint32_3 = default) 
 : this() {
  Uint32_0 = uint32_0;
  Uint32_1 = uint32_1;
  Uint32_2 = uint32_2;
  Uint32_3 = uint32_3;
}
```

i didn't create a pr for two main reasons:
 - this is more of a design decision, i don't want to take that away
 - i suspect this ctor is generated, if so i guess it wouldn't even as simple as just changing it in a pr. again: i don't know too much how this project works under the hood, i just know suff about c#

another solution would be to do some pointer arithmetic in the ctors so you don't need to invoke the empty ctor, but i think the first suggestion would be cleaner.

btw this took me ~5h to figure out with RenderDec ...  was the last thing i suspected.

ofcourse you can always just use it the c way by first calling the empty ctor and then setting the fields, but i prefer a propper ctor.

Perksey:
# Summary of the PR
Closes a bunch of issues up.

Closes #189 
Closes #163 
Closes #214 

# Further Comments


src/Assimp/Silk.NET.Assimp/Structs/AABB.gen.cs
```diff
@@ -21,12 +22,24 @@ public unsafe partial struct AABB
...
         (
-            System.Numerics.Vector3 mMin = default,
-            System.Numerics.Vector3 mMax = default
+            System.Numerics.Vector3? mMin = null,
+            System.Numerics.Vector3? mMax = null
         )
         {
-            MMin = mMin;
-            MMax = mMax;
+            fixed (AABB* @this = &this)
```
HurricanKai, line 29: ?? why is this better then `: this()` (I know generated code, but this just seems to be unnecessarily unsafe)

src/OpenCL/Extensions/Silk.NET.OpenCL.Extensions.ARM/ArmImportMemory.gen.cs
```diff
@@ -4,12 +4,14 @@
...
 using System;
 using System.Runtime.InteropServices;
+using System.Runtime.CompilerServices;
 using System.Text;
 using Silk.NET.OpenCL;
 using Silk.NET.Core.Native;
 using Silk.NET.Core.Attributes;
 using Silk.NET.Core.Contexts;
 using Silk.NET.Core.Loader;
+using Extension = Silk.NET.Core.Attributes.ExtensionAttribute;
```
HurricanKai, line 14: Generated file but this ease-of-use feature?

HurricanKai (approved):
Little bit difficult to review, so no guarantee I didn't miss something, but looks good overall. Added some comments I'm confused about, but nothing blocking.

src/Assimp/Silk.NET.Assimp/Structs/AABB.gen.cs
Perksey: Will that work on structs? The idea here is to force the runtime to initialise all fields to default so that we don't have to do it, which caused problems in the past.

src/OpenCL/Extensions/Silk.NET.OpenCL.Extensions.ARM/ArmImportMemory.gen.cs
Perksey: This is actually to resolve a conflict as there's a ExtensionAttribute somewhere in the BCL of .NET Standard 2.1

HurricanKai: Fair enough, `System.Runtime.CompilerServices` thought not sure why not just use `Silk.NET.Core.Attributes.ExtensionAttribute` directly

Perksey: Meh could do, just a matter of personal preference really. Can change if you require.

HurricanKai: no, ofc not. just a nit

src/Assimp/Silk.NET.Assimp/Structs/AABB.gen.cs
Perksey: Now uses : this()

PR #361, opened 2020-12-03
[WIP] added base documentation with SHFB

WholesomeIsland:
# Summary of the PR
This adds a way to generate documentation from XML Comments In the code. I used https://github.com/EWSoftware/SHFB to generate the .chm file which can be extracted to get the .html files to upload to a website.

Any thing to do with missing documentation
# Further Comments
the documentation can be found in documentation/html


Perksey:
Hi there, thanks for using Silk.NET and proposing this pull request!

I really appreciate the sentiment behind it, however I don't think this is something the team would like to bring into Silk.NET. Primarily because:
- Cross-platform is very important to us. SHFB is an older, non-SDK-style, .NET Framework-only MSBuild SDK, which means it can only be built on Windows. We only allow non-cross-platform build components if they're a) mission-critical and b) a silk.net build can function without it. This pull request only meets point B, and even then it could lead to the CHM file becoming outdated.
- https://docs.ultz.co.uk/ seems to cover us on the docs front. If this change were to occur anywhere it would be in the [docs repo](https://github.com/Ultz/docs.ultz.co.uk)
- CHM files, while incredibly useful, are more geared towards "F1 help" and given Silk.NET isn't an application, rather a library, it doesn't feel the most appropriate here; however this is a minor nit.

With all that in mind, I think I'm going to close this pull request. Again, I really really appreciate the sentiment behind it and if there's one area in which Silk.NET needs to improve it is documentation; but this workload in particular just isn't the most applicable to Silk.NET. I'm sorry for the inconvenience.

For future reference, however, we usually prefer potential pull requests to be discussed beforehand, either in our Discord server or in the GitHub Discussions tab, so that you don't end up wasting your time!

Thanks again for the pull request, there aren't many external contributors doing that for Silk.NET so it's awesome to see people like you trying to get involved! :D

PR #368, opened 2020-12-08
Proposal - Abstractions

WholesomeIsland:
# Summary of the PR
A proposal for Abstractions of OpenGL.

# Further Comments


Perksey:
- Rejected
- It's not our bindings to be high-level and abstract away
- Project Aliquip will have cross-platform C#-style abstractions
- Lack of use case per se
- Very specific abstractions
- Copied from tutorials essentially
- We already plan to make generated higher-level bindings that are less specific

PR #369, opened 2020-12-08
Bitwise scalar proposal

HurricanKai:


Perksey:
- Approved with byte -> int change

Perksey (approved):

PR #373, opened 2020-12-18
SilkTouch staging & injecting

### LINKED ISSUE #274: Allow injection of code before and after a native call
This is required for some plans for Vulkan and OpennXR that aren't mission-critical, but would be nice to have. The main use case is injecting prologue which will detect when a new `Instance` or `Device` is passed in, and notify the loader accordingly so that we don't have to mandate that the user keep the CurrentInstance and CurrentDevice properties up-to-date.
- The injected code should not affect marshalling in any way.
- This could be done by maybe adding a `PrologueAttribute` or `EpilogueAttribute` added to the class or method with a small snippet of code inside, which could be parsed by Roslyn and added to the AST that SilkTouch generates.

HurricanKai:
Resolves #274

### DIFF
--- a/src/Core/Silk.NET.Core/Attributes/InjectAttribute.cs
+++ b/src/Core/Silk.NET.Core/Attributes/InjectAttribute.cs
@@ -1,0 +1,65 @@
+﻿// This file is part of Silk.NET.
+// 
+// You may modify and distribute Silk.NET under the terms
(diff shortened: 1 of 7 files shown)


Perksey (approved):

Perksey:
Will merge after CI fails as expected

PR #376, opened 2020-12-18
2.0 Bug squashing and issue closing

### LINKED ISSUE #279: GLFW bindings on Linux need priority/location/name loader changes
# Summary
The binding on GLFW for Linux bind to libglfw.so.3 instead of libglfw.so.3.3 and then check for v3.3+.
This breaks on Ubuntu 16.04 (Xenial) / Debian 9.x (Stretch) / Mint 18 - 18.3 (Sarah, serena, Sonya, Sylvia) as they only ship GLFW v3.1.2.
It should probably try to bind to system libglfw.so.3.3 first, then fall back to system libglfw.so.3, and then check any so bins shipped with Silk.NET.

# Steps to reproduce
- Platform: Desktop Linux / Ubuntu 16.04 (Xenial) / Debian 9.x (Stretch) / Mint 18 - 18.3 (Sarah, serena, Sonya, Sylvia) 
- Framework Version: .NET Core 3.1
- API: OpenGL / OpenGL ES
- API Version: any
1. Target a Linux system with libglfw build that is prior to v3.3
2. Try to use GLFW to create a window using Silk.NET

# Comments

ULTZ Perksey;
> Hmmm I should probably rework how Linux file names work in 2.0 as that’s been a problem that’s plagued silk since its earliest > days, just hasn’t been too significant as we ship our own binaries with the names we expect
> Ah yeah, that’s another thing I need to fix: priority

### LINKED ISSUE #286: Add more bit information to VideoMode
such as RedBits, BlueBits, etc; instead of just DepthBits

### LINKED ISSUE #290: Function pointer overloads
We should provide function pointer overloads where we currently have callback delegates.
Because in net5.0s `UnmanagedCallersOnlyAttribute` marked methods can be directly called from native code, and this probably has performance implications.

### LINKED ISSUE #334: Sharing resources between different GL contexts
# Summary of feature
Implement a way to share resources between different GL contexts, which would allow for things like rendering something while loading content in the background, instead of halting the window.

### LINKED ISSUE #350: SDL events not working for multiple windows
# Summary
Currently all SDL events get blindly dispatched to the window polling the message loop. Instead the events should be dispatched to the window as provided by the event's window id or (if not set) to the window having focus.

# Steps to reproduce
- Platform: Desktop
1. Add a second window to the basic example in the 2.0 branch
2. Prioritize SDL
3. See all events getting routed to the window calling DoEvents

# Comments
Maybe something like https://github.com/stride3d/stride/blob/master/sources/engine/Stride.Graphics/SDL/Application.cs should be introduced? What would be the proper approach?

Sadly I can't really build the solution. It's complaining about files not being found under `C:\Silk.NET\...` - once I move everything there, Visual Studio goes completely nuts, eating up all of my memory and in git I see >2000 or more *.gen.cs files. I guess that's not how it's supposed to be?

Perksey:
# Summary of the PR
Closes #279 
Closes #286 
Closes #290 
Closes #334 
Closes #350

Links go here.

# Further Comments


HurricanKai:
See feature/eliminate-all-warnings
potentially merge that branch into this one, and fix up the rest of the warnings (iirc there aren't that many left)

Perksey:
If you can open a PR that does that that'd be great.

src/Core/Silk.NET.Core/Miscellaneous/BreakneckLock.cs
```diff
@@ -0,0 +1,197 @@
...
+{
+    /// <summary>
+    /// Provided a lightweight spin lock for synchronization in high performance
+    /// scenarios with a low hold time
+    /// </summary>
+    /// <remarks>
+    /// This lock is very performant, but it is very dangerous (hence breakneck).
+    /// It's recommended to use the framework-provided locks where possible.
+    /// </remarks>
+    public struct BreakneckLock
```
HurricanKai, line 47: Why is this in core?

src/Windowing/Silk.NET.Windowing.Glfw/GlfwWindow.cs
```diff
@@ -306,16 +307,29 @@ protected override void CoreInitialize(WindowOptions opts)
             // Set video mode (-1 = don't care)
             _glfw.WindowHint(WindowHintInt.RefreshRate, opts.VideoMode.RefreshRate ?? -1);
             _glfw.WindowHint(WindowHintInt.DepthBits, opts.PreferredDepthBufferBits ?? -1);
+            _glfw.WindowHint(WindowHintInt.StencilBits, opts.PreferredStencilBufferBits ?? -1);
```
HurricanKai, line 310: I don't know, but are you sure we should set the hint if the user provided no values?

src/Core/Silk.NET.Core/Miscellaneous/BreakneckLock.cs
Perksey: Where else would it be? It may be used elsewhere in the codebase in the future, and is not specific to windowing.

src/Windowing/Silk.NET.Windowing.Glfw/GlfwWindow.cs
Perksey: -1 is GLFW_DONT_CARE meaning that is has the same effect as not setting it at all

src/Core/Silk.NET.Core/Miscellaneous/BreakneckLock.cs
HurricanKai: SDL - if it is used anywhere else, I'd just copy paste it. Sure bit of extra work, but much better then having a public type `BreakneckLock`

Perksey: Personally very against this. Copying and pasting code across codebases is not the way to go when you can have a common assembly - this is literally why the core exists.

Besides, it will probably be useful for performance junkies using Silk.NET and will be documented at some point in the future.

HurricanKai: Fair point, though in that case I'd name it something different - BreakneckLock just sounds like "this should be internal but we couldn't be bothered to make it"

Perksey: Hmm yeah, I still want a word that conveys both the speed and lack of safety in this lock though and breakneck is all I could come up with.

HurricanKai: `UnsafeFastLock`?

Perksey: Ideally one word, not keen on those it doesn't really roll off the tongue well.

HurricanKai: That's the point though?

Perksey: I don't want people to not use this type, I just want to convey that it's quite easy to go wrong with it which is fine as all of Silk.NET is quite easy to go wrong with.

HurricanKai: I'd argue people that want to use this should use https://github.com/john-h-k/SpinLockSlim instead

Perksey: I mean there is zero difference between the two, and the benefit of this being in core is it's free - no need to take on a dependency. It will pretty much stay 1:1 with the original as part of our monthly updates.

HurricanKai (approved):

Perksey:
This is gonna feel good...

PR #377, opened 2020-12-18
Resolve some warnings

HurricanKai:
See https://github.com/Ultz/Silk.NET/pull/376#issuecomment-748302945


src/Core/Silk.NET.Core/Native/ConcurrentDictionaryVTable.cs
```diff
@@ -10,6 +10,10 @@
 
 namespace Silk.NET.Core.Native
 {
+    /// <summary>
+    /// A simple implementation of a <see cref="IVTable" /> using a concurrent dictionary.
+    /// See https://ultz.co.uk/blog/silktouch-slots-vtables/ for an in-depth explanation.
+    /// </summary>
     public class ConcurrentDictionaryVTable : IVTable
```
Perksey, line 17: I think providing an alternative implementation of IVTable in Core is incredibly misleading given that we discourage non-generated VTables.

HurricanKai: True - though I don't think removing it is a good idea either, it can be a useful tool. Maybe deprecating is a good idea, doesn't remove it, but strongly discourages it's use.

Perksey: Will SilkTouch even work with ConcurrentDictionaryVTables these days?

HurricanKai: actually, I don't think so. Will remove

Perksey:
Stale PR and probably outdated by now, closing.

Issue #180, opened 2020-05-20
[2.0] Vulkan function tables


Perksey:
In Vulkan, function loading is both instance and device dependent. In 2.0, we want to remove the dependency on the CurrentInstance and CurrentDevice properties. This means we have a lot of work to do as currently 1.0 can be used with multiple instances and multiple devices (albeit in a bit of a hacky way) using the `PurgeEntryPoints` method, causing Vulkan entry points to be reloaded using the new instance and device. In 2.0, we should:

- [x] Remove the extension loader's dependency on the CurrentInstance and CurrentDevice properties by making the user pass in an instance and/or device when loading extensions
- [x] Add custom function table capabilities to SilkTouch
- [x] Add a device function table
- [x] Add an instance device function table table
- [ ] Auto-detect usage of new instances and devices by using SilkTouch prologues with Vulkan, ensuring the function table currently in use is the right one

Perksey:
My initial ideas for this would severely impact performance, so what I'm going to do is revive CurrentInstance and CurrentDevice, but automatically assign them using the new code injection features. You'd only have to mess with these properties if you're using multiple instances and/or multiple devices.

PR #385, opened 2020-12-30
Bug fixes and Improvements to Vk and XR.

thargy:
# Summary of the PR
Updates `Vk` and `XR` classes to improve thread-safety, performance, and memory utilisation, whilst also removing a bug which would cause an 'infinite' loop when the number of extension properties exceeded 128.

Also adds `Vk.GetPhysicalDevices` to retrieve the cache of physical devices for an instance.

Updated `VulkanTriangles` to leverage API changes, and to review stack allocations.

The `VulkanTriangles` project was broken when accessing a device that had >128 extensions and would enter an 'infinite' loop allocating increasing amounts of memory until the application crashed.  It was also allocating large data structures (>1k) on the stack, which risksed stack overflows.

The resultant fixes re-worked the areas that caused these problems and are based on discussions on Discord.

# Further Comments
A full explanation of the code logic changes is included in `Vk.cs` in the comments.

### DIFF
--- a/src/Lab/VulkanTriangle/HelloTriangleApplication.cs
+++ b/src/Lab/VulkanTriangle/HelloTriangleApplication.cs
@@ -4,9 +4,11 @@
 // of the MIT license. See the LICENSE file for details.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using Silk.NET.Core;
 using Silk.NET.Core.Native;
@@ -175,37 +177,37 @@ namespace VulkanTriangle
         {
             for (var i = 0; i < MaxFramesInFlight; i++)
             {
-                _vk.DestroySemaphore(_device, _renderFinishedSemaphores[i], (AllocationCallbacks*) null);
-                _vk.DestroySemaphore(_device, _imageAvailableSemaphores[i], (AllocationCallbacks*) null);
-                _vk.DestroyFence(_device, _inFlightFences[i], (AllocationCallbacks*) null);
+                _vk.DestroySemaphore(_device, _renderFinishedSemaphores[i], null);
+                _vk.DestroySemaphore(_device, _imageAvailableSemaphores[i], null);
+                _vk.DestroyFence(_device, _inFlightFences[i], null);
             }
 
-            _vk.DestroyCommandPool(_device, _commandPool, (AllocationCallbacks*) null);
+            _vk.DestroyCommandPool(_device, _commandPool, null);
 
             foreach (var framebuffer in _swapchainFramebuffers)
             {
-                _vk.DestroyFramebuffer(_device, framebuffer, (AllocationCallbacks*) null);
+                _vk.DestroyFramebuffer(_device, framebuffer, null);
             }
 
-            _vk.DestroyPipeline(_device, _graphicsPipeline, (AllocationCallbacks*) null);
-            _vk.DestroyPipelineLayout(_device, _pipelineLayout, (AllocationCallbacks*) null);
-            _vk.DestroyRenderPass(_device, _renderPass, (AllocationCallbacks*) null);
+            _vk.DestroyPipeline(_device, _graphicsPipeline, null);
+            _vk.DestroyPipelineLayout(_device, _pipelineLayout, null);
+            _vk.DestroyRenderPass(_device, _renderPass, null);
 
             foreach (var imageView in _swapchainImageViews)
             {
-                _vk.DestroyImageView(_device, imageView, (AllocationCallbacks*) null);
+                _vk.DestroyImageView(_device, imageView, null);
             }
 
-            _vkSwapchain.DestroySwapchain(_device, _swapchain, (AllocationCallbacks*) null);
-            _vk.DestroyDevice(_device, (AllocationCallbacks*) null);
+            _vkSwapchain.DestroySwapchain(_device, _swapchain, null);
+            _vk.DestroyDevice(_device, null);
 
             if (EnableValidationLayers)
             {
-                _debugUtils.DestroyDebugUtilsMessenger(_instance, _debugMessenger, (AllocationCallbacks*) null);
+                _debugUtils.DestroyDebugUtilsMessenger(_instance, _debugMessenger, null);
             }
 
-            _vkSurface.DestroySurface(_instance, _surface, (AllocationCallbacks*) null);
-            _vk.DestroyInstance(_instance, (AllocationCallbacks*) null);
+            _vkSurface.DestroySurface(_instance, _surface, null);
+            _vk.DestroyInstance(_instance, null);
         }
 
         private unsafe void CreateInstance()
@@ -233,8 +235,8 @@ namespace VulkanTriangle
                 PApplicationInfo = &appInfo
             };
 
-            var extensions = (byte**) _window.VkSurface!.GetRequiredExtensions(out var extCount);
-            var newExtensions = stackalloc byte*[(int)(extCount + _instanceExtensions.Length)];
+            var extensions = _window.VkSurface!.GetRequiredExtensions(out var extCount);
+            var newExtensions = stackalloc byte*[(int) (extCount + _instanceExtensions.Length)];
             for (var i = 0; i < extCount; i++)
             {
                 newExtensions[i] = extensions[i];
@@ -269,12 +271,12 @@ namespace VulkanTriangle
             }
 
             _vk.CurrentInstance = _instance;
-            
+
             if (!_vk.TryGetInstanceExtension(_instance, out _vkSurface))
             {
                 throw new NotSupportedException("KHR_surface extension not found.");
             }
-            
+
             Marshal.FreeHGlobal((IntPtr) appInfo.PApplicationName);
             Marshal.FreeHGlobal((IntPtr) appInfo.PEngineName);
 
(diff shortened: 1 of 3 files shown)


src/Vulkan/Silk.NET.Vulkan/Vk.cs
```diff
@@ -197,18 +225,69 @@ public bool IsDeviceExtensionPresent(Instance instance, string extension)
...
+                            result = true;
+                        }
+                    }
+                }
+                finally
+                {
+                    _cachedDeviceExtensionsLock.ExitWriteLock();
+                    mem?.Dispose();
+                }
+            } // else result = false - takes 2 lookups, one to check for extension, and one to check initialisation.
```
HurricanKai, line 287: I'd expect it to be faster to use a separate bool or nullability to indicate initialization. Why the extra lookup?

src/Vulkan/Silk.NET.Vulkan/Vk.cs
```diff
@@ -221,23 +300,33 @@ public unsafe bool IsDeviceExtensionPresent(PhysicalDevice device, string extens
         /// <returns>Whether the device extension is available.</returns>
         public unsafe bool IsDeviceExtensionPresent(Instance instance, string extension, out PhysicalDevice device)
         {
-            var physicalDeviceCount = 0u;
-            EnumeratePhysicalDevices(instance, &physicalDeviceCount, null);
-            var physicalDevices = stackalloc PhysicalDevice[(int) physicalDeviceCount];
-            EnumeratePhysicalDevices(instance, &physicalDeviceCount, physicalDevices);
+            device = GetPhysicalDevices(instance).FirstOrDefault(pd => IsDeviceExtensionPresent(pd, extension));
+            return device.Handle != IntPtr.Zero;
```
HurricanKai, line 304:
```suggestion
            return GetPhysicalDevices(instance).Any(pd => IsDeviceExtensionPresent(pd, extension));
```

src/Vulkan/Silk.NET.Vulkan/Vk.cs
```diff
@@ -2,18 +2,20 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Runtime.InteropServices;
+using System.Threading;
+using Unsafe = System.Runtime.CompilerServices.Unsafe;
```
HurricanKai, line 6: please move to bottom of usings

src/Vulkan/Silk.NET.Vulkan/Vk.cs
```diff
@@ -2,18 +2,20 @@
...
 using Silk.NET.Core;
 using Silk.NET.Core.Attributes;
 using Silk.NET.Core.Contexts;
 using Silk.NET.Core.Loader;
 using Silk.NET.Core.Native;
 using LibraryLoader = Silk.NET.Core.Loader.LibraryLoader;
+using System.Collections.Concurrent;
```
HurricanKai, line 13: Please sort into rest of System usings

src/Vulkan/Silk.NET.Vulkan/Vk.cs
```diff
@@ -197,18 +225,69 @@ public bool IsDeviceExtensionPresent(Instance instance, string extension)
...
+                    mem = GlobalMemory.Allocate((int) deviceExtPropertiesCount * sizeof(ExtensionProperties));
+                    var props = (ExtensionProperties*) Unsafe.AsPointer(ref mem.GetPinnableReference());
 
-            return list.Contains(extension);
+                    // Get properties
+                    EnumerateDeviceExtensionProperties(device, (byte*) 0, &deviceExtPropertiesCount, props);
+                    for (int j = 0; j < deviceExtPropertiesCount; j++)
+                    {
+                        // Prefix the extension name
+                        var newKey = prefix_sep + Marshal.PtrToStringAnsi((IntPtr) props[j].ExtensionName);
```
HurricanKai, line 272: Preferably use SilkMarshal over Marshal to convert from ptr to string

HurricanKai (changes requested):
Amazing work, Thanks! Just a few small comments.

src/Vulkan/Silk.NET.Vulkan/Vk.cs
thargy: You have to indicate initialisation per device, which is itself unbounded, requiring a hashset.  For an optimisation we use the same hashset as the composite one.  There is an argument that you could hash the IntPtr's separately to indicate initialisation, and, in fact that might be very slightly quicker, but as we need to create the string version for the composite key anyway and the hash lookup of instance occurs only on cache miss, the additional code bloat and complexity (locking) would not make it worth it in practice.

There is always a worst case of 2 lookups - one to see if the device is initialised and one to see if the extension was found.  The naive original code would first lookup to see if initialised (is the key in the Dictionary O(1) this is the same speed as the HashSet lookup), then would initialise (an O(n) operation, scanning) if necessary and then perform a lookup in the list which is O(n).  This code does the hit test first - O(1) - checks initialisation - O(1) - and, if necessary, initialises O(n), but uses the initialization scan to also locate.  As such, this is O(1), except during initialisation.

In real world use, I'd expect the performance of this code to be significantly faster for all but the smallest collection sizes.  Certainly for collections of 100+ items the `HashSet.Contains` is significantly faster than the `List`.  Most importantly the time remains effectively constant, independent of the size of the extensions (save for the initialisation).

src/Vulkan/Silk.NET.Vulkan/Vk.cs
thargy: Your suggestion would not initialise `device` correctly, which is an `out` parameter.

src/Vulkan/Silk.NET.Vulkan/Vk.cs
thargy: This was using the existing code.  I am unsure of how `SilkMarshal` works and whether it is equivalent to `Marshal` which would therefore require more testing.

thargy: I would do this but there's also a lot of mixed-use of `Marshal` and `SilkMarshal` in `HelloTriangleApplications.cs`, and the rest of the codebase for that matter.  It doesn't strike me as a change relating to this PR to review all `Marshal` usages and update to `SilkMarshal`?  However, if this is required, there are 2 places in `VK.cs` and 1 in `XR.cs` which use `Marshal`, specifically `Marshal.PtrToStringAnsi`, and I assume can be converted to `SilkMarshal.PtrToString`.

src/Lab/VulkanTriangle/HelloTriangleApplication.cs
```diff
@@ -233,8 +235,10 @@ private unsafe void CreateInstance()
                 PApplicationInfo = &appInfo
             };
 
-            var extensions = (byte**) _window.VkSurface!.GetRequiredExtensions(out var extCount);
-            var newExtensions = stackalloc byte*[(int)(extCount + _instanceExtensions.Length)];
+            var extensions = _window.VkSurface!.GetRequiredExtensions(out var extCount);
+            // TODO Review that this count doesn't realistically exceed 1k (recommended max for stackalloc)
+            // Should probably be allocated on heap anyway as this isn't super performance critical.
```
Perksey, line 240: Yup, agree.

src/Vulkan/Silk.NET.Vulkan/Vk.cs
Perksey: Doesn't really matter in this instance, but the Mono implementation of the Marshal methods are pretty inconsistent. Doesn't have to be now, but eventually we should switchover to SilkMarshal fully.

src/Vulkan/Silk.NET.Vulkan/Vk.cs
Perksey: This shouldn't be a hot path so not too concerned.

src/Vulkan/Silk.NET.Vulkan/Vk.cs
Perksey: LGTM

src/OpenXR/Silk.NET.OpenXR/XR.cs
```diff
@@ -16,17 +17,18 @@ public partial class XR
...
-             var ret = new XR(CreateDefaultContext(new OpenXRLibraryNameContainer().GetLibraryName()));
-             return ret;
+            return new XR(CreateDefaultContext(new OpenXRLibraryNameContainer().GetLibraryName()));
         }
 
         [Obsolete("Use IsInstanceExtensionPresent instead.", true)]
         public override bool IsExtensionPresent(string extension) => IsInstanceExtensionPresent(null, extension);
-        private Dictionary<string, List<string>> _cachedInstanceExtensions = new Dictionary<string, List<string>>();
+        private readonly HashSet<string> _cachedInstanceExtensions = new HashSet<string>();
+        private readonly ReaderWriterLockSlim _cachedInstanceExtensionsLock = new ReaderWriterLockSlim();
```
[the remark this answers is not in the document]
Perksey: Oh by the way make sure this will work for multiple instances as my changes as part of #180 will add better support for multiple instances.

src/Vulkan/Silk.NET.Vulkan/Vk.cs
thargy: Further to @Perksey this is for a later review.

src/OpenXR/Silk.NET.OpenXR/XR.cs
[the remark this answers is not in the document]
thargy: `BreakneckLock` doesn't support read upgrading.  Further, the write is _very_ slow compared to the read, so a spinlock may not be the best idea (even a very fast one).  `ReaderWriterLockSlim` is effectively a fast spinlock on read, unless there is a pending (slow write) in which case it will yield.  It's designed for this asymmetric load.

src/Vulkan/Silk.NET.Vulkan/Vk.cs
thargy: Approved by @Perksey

src/OpenXR/Silk.NET.OpenXR/XR.cs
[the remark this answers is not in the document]
Perksey: Fair enough. I'm still not convinced it's entirely as slim as it could be, given it's a BCL type there's got to be some safety we can rip out, but this is good for now.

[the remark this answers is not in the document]
thargy: > Oh by the way make sure this will work for multiple instances as my changes as part of #180 will add better support for multiple instances.

Yes, even though it's a single hashset it's keyed as '<instance>|<extension>'

thargy (commented):
Finished reviews, LGTM

Perksey (approved):
LGTM - @HurricanKai any last comments?

HurricanKai (approved):
Forgot to approve after resolving conversations. LGTM

Issue #394, opened 2021-01-02
Small elapsed time on Render() when vsync is enabled


ThomasMiz:
# Summary
When VSync is on, the elapsed time passed to the Render() method is typically a very low value on some machines (in mine, 0.00001 to 0.001, which is waaay lower than the ~0.016 expected with a window configured to 60fps).

(When VSync is turned off, in my machine the elapsed time is around 0.016 as expected).

# Steps to reproduce
- Platform: Desktop
- NET Version: .NET 5
- API: OpenGL 3.3

I set up a small project to try to reproduce the problem. To run, download the 2.0 release Silk.NET.OpenGL, Windowing and Input packages and run a project with nothing but [this Program.cs file](https://cdn.discordapp.com/attachments/607634593201520651/795043831732764742/Program.cs) (don't worry, it's quite small 😉)

# Comments
Since this problem doesn't seem to be happening to anyone but me, I will investigate this issue further.

ThomasMiz:
Fixed by PR #395

thargy:
Note.  the `WindowOptions.Vsync` property sets the `SwapInterval` for OpenGL.  It therefore, doesn't have any effect automatically on Vulkan, other than ignoring the `FramesPerSecond` property, which can lead to unbounded render calls if the implementor has not enabled VSync on Vulkan. (See [this StackOverflow answer for how to implement VSync on Vulkan](https://stackoverflow.com/questions/36896021/enabling-vsync-in-vulkan))

HurricanKai:
Note on Vulkan VSync: The render calls are still unbound, and we may need to fix this. The basic approach to VSync in Vulkan is using `VK_PRESENT_MODE_MAILBOX_KHR` or `VK_PRESENT_MODE_FIFO_RELAXED_KHR` which don't do any VSync functionality. They just specify when images are presented to the screen. This can then be waited for using fences/semaphores.
not 100% sure how events in C# behave right now, but it may be possible that we call the Render functions so quickly that a queue of event handlers form that just wait for the semaphore images to become available.
For similar reasoning, I think we need to consider always enabling "VSync" (skipping the check) when using vulkan - we don't do any handling on Vulkan so I don't think it's useful to pretend we do.
[...]

PR #395, opened 2021-01-03
Fix elapsed time parameters on Window.Render and Update events

ThomasMiz:
Fixes issue #394.

In summary, the stopwatches need to be restarted _before_ the events are invoked. Or really, stopwatches need to be restarted right after the elapsed time was taken. Otherwise, the time the invocation of the method took (and therefore the whole update/render logic the library user wrote) will not be counted as elapsed time on the next invocation.

If you store a double time that starts at 0 and each update add to it the elapsed time parameter, you'd now get nearly exactly the same as if you just used a stopwatch that was started on load, even if your Update method takes a very long time. Before this change, you'd see that your time variable would stay behind (this might not have been noticeable if your Update method is very fast).

### DIFF
--- a/src/Windowing/Silk.NET.Windowing.Common/Internals/ViewImplementationBase.cs
+++ b/src/Windowing/Silk.NET.Windowing.Common/Internals/ViewImplementationBase.cs
@@ -165,13 +165,14 @@ namespace Silk.NET.Windowing.Internals
                     _swapIntervalChanged = false;
                 }
 
-                Render?.Invoke(_renderStopwatch.Elapsed.TotalSeconds);
+                delta = _renderStopwatch.Elapsed.TotalSeconds;
+                _renderStopwatch.Restart();
+                Render?.Invoke(delta);
+
                 if (ShouldSwapAutomatically)
                 {
                     GLContext?.SwapBuffers();
                 }
-
-                _renderStopwatch.Restart();
             }
         }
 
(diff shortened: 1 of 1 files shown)


thargy:
Actually, the recommended approach here is to use `Stopwatch.GetTimestamp()` which returns a `long` and store that value each time round the loop.  To calculate the `delta` you use logic like so:
```csharp
long timestamp = Stopwatch.GetTimestamp();
while (true)
{
	// Simulate loop logic
	Thread.Sleep(100);

	double delta = ((double)-Interlocked.Exchange(ref timestamp, Stopwatch.GetTimestamp()) + timestamp) / Stopwatch.Frequency;
}
```

The reason for this approach is precision.  Under the hood this is effectively how `Stopwatch` works anyway, but instead of converting everything to doubles and incrementing them, you are keeping the underlying timestamp and so you will not experience drift over time.

Perksey (approved):

thargy:
Hi @ThomasMiz,

I really don't see how you changes fix #394.  From that issue (**emphasis** added):
> **When VSync is on**, the elapsed time passed to the Render() method is typically a very low value on some machines (in mine, 0.00001 to 0.001, which is waaay lower than the ~0.016 expected with a window configured to 60fps).

However, you will note that the check on whether to call the user's Render method, occurs [here](https://github.com/ThomasMiz/Silk.NET/blob/0600b9ea7d88a4881349483dcaef6cc079991f48/src/Windowing/Silk.NET.Windowing.Common/Internals/ViewImplementationBase.cs#L155):

```csharp
if ((delta >= _renderPeriod) || VSync)
```

i.e. if `VSync` is on, the elapsed time is ignored.

In my testing, I still experience the VSync on issue described in #394.

PR #398, opened 2021-01-03
Improvements to timer logic.

thargy:
# Summary of the PR
This is a minor improvement to the timer logic on view, following a discussion of #395, and utilises the underlying system timestamps more efficiently.

This was discussed on discord.

The initial timer check is now reduced to simple `long` arithmetic for optimum speed with a single call to `Stopwatch.GetTimestamp()`, as this can be called repeatedly.  When calculating the elapsed seconds as a `double` the timestamp is retrieved once again as well as 'restarting the clock' immediately before invoking the user code, giving the most accurate (and non-drifting) time delta.

### DIFF
--- a/src/Windowing/Silk.NET.Windowing.Common/Interfaces/IViewProperties.cs
+++ b/src/Windowing/Silk.NET.Windowing.Common/Interfaces/IViewProperties.cs
@@ -39,6 +39,7 @@ namespace Silk.NET.Windowing
         /// <summary>
         /// The number of rendering operations to run every second.
         /// </summary>
+        /// <remarks>This value is ignored if <see cref="VSync">VSync</see> is set to <see langref="true"/>.</remarks>
         double FramesPerSecond { get; set; }
 
         /// <summary>
@@ -54,6 +55,8 @@ namespace Silk.NET.Windowing
         /// <summary>
         /// Whether or not VSync is enabled for this view.
         /// </summary>
+        /// <remarks>This value will set the SwapInterval for OpenGL, but on other frameworks (e.g. Vulkan) you will need to turn on
+        /// VSync manually.  Whilst <see langref="true"/>, the value of <see cref="FramesPerSecond"/> is ignored, to prevent frame skipping.</remarks>
         bool VSync { get; set; }
 
         /// <summary>
--- a/src/Windowing/Silk.NET.Windowing.Common/Internals/ViewImplementationBase.cs
+++ b/src/Windowing/Silk.NET.Windowing.Common/Internals/ViewImplementationBase.cs
@@ -165,13 +165,14 @@ namespace Silk.NET.Windowing.Internals
                     _swapIntervalChanged = false;
                 }
 
-                Render?.Invoke(_renderStopwatch.Elapsed.TotalSeconds);
+                delta = _renderStopwatch.Elapsed.TotalSeconds;
+                _renderStopwatch.Restart();
+                Render?.Invoke(delta);
+
                 if (ShouldSwapAutomatically)
                 {
                     GLContext?.SwapBuffers();
                 }
-
-                _renderStopwatch.Restart();
             }
         }
 
@@ -180,8 +181,8 @@ namespace Silk.NET.Windowing.Internals
             var delta = _updateStopwatch.Elapsed.TotalSeconds;
             if (delta >= _updatePeriod)
             {
-                Update?.Invoke(_updateStopwatch.Elapsed.TotalSeconds);
                 _updateStopwatch.Restart();
+                Update?.Invoke(delta);
             }
         }
 
--- a/src/Windowing/Silk.NET.Windowing.Common/Internals/ViewImplementationBase.cs
+++ b/src/Windowing/Silk.NET.Windowing.Common/Internals/ViewImplementationBase.cs
@@ -150,7 +150,7 @@ namespace Silk.NET.Windowing.Internals
         {
             DoInvokes();
             // Check elapsed time
-            if (((Stopwatch.GetTimestamp() - _renderTimestamp) >= _renderPeriod) || VSync)
+            if (VSync || ((Stopwatch.GetTimestamp() - _renderTimestamp) >= _renderPeriod))
             {
                 if (!(GLContext is null) && !GLContext.IsCurrent)
                 {
--- a/src/Windowing/Silk.NET.Windowing.Common/Internals/ViewImplementationBase.cs
+++ b/src/Windowing/Silk.NET.Windowing.Common/Internals/ViewImplementationBase.cs
@@ -164,9 +164,9 @@ namespace Silk.NET.Windowing.Internals
                 }
 
                 // Re-calculte the elapsed time, resetting the current timestamp
-                var newTimeStamp = Stopwatch.GetTimestamp();
-                var delta = (newTimeStamp - _renderTimestamp) / Frequency;
-                _renderTimestamp = newTimeStamp;
+                var newTimestamp = Stopwatch.GetTimestamp();
+                var delta = (newTimestamp - _renderTimestamp) / Frequency;
+                _renderTimestamp = newTimestamp;
                 Render?.Invoke(delta);
 
                 if (ShouldSwapAutomatically)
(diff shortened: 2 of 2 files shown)


thargy:
Manual testing completed.

src/Windowing/Silk.NET.Windowing.Common/Internals/ViewImplementationBase.cs
```diff
@@ -165,9 +163,8 @@ public void DoRender()
                     _swapIntervalChanged = false;
                 }
 
-                delta = _renderStopwatch.Elapsed.TotalSeconds;
-                _renderStopwatch.Restart();
-                Render?.Invoke(delta);
+                // Re-calculte the elapsed time, resetting the current timestamp
+                Render?.Invoke((-Interlocked.Exchange(ref _renderTimestamp, Stopwatch.GetTimestamp()) + _renderTimestamp) / Frequency);
```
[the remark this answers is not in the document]
Perksey: Yeah please, thanks.

src/Windowing/Silk.NET.Windowing.Common/Internals/ViewImplementationBase.cs
```diff
@@ -176,10 +179,13 @@ public void DoRender()
         public void DoUpdate()
         {
             // Check elapsed time
-            if ((Stopwatch.GetTimestamp() - _updateTimestamp) >= _updatePeriod)
+            var newTimeStamp = Stopwatch.GetTimestamp();
```
Perksey, line 182: nit: capitalisation of timestamp is inconsistent throughout this code.

src/Windowing/Silk.NET.Windowing.Common/Internals/ViewImplementationBase.cs
```diff
@@ -176,10 +179,13 @@ public void DoRender()
...
+            if ((newTimeStamp - _updateTimestamp) >= _updatePeriod)
             {
                 // Re-calculte the elapsed time, resetting the current timestamp
-                Update?.Invoke((-Interlocked.Exchange(ref _updateTimestamp, Stopwatch.GetTimestamp()) + _updateTimestamp) / Frequency);
+                var delta = (newTimeStamp - _updateTimestamp) / Frequency;
+                _updateTimestamp = newTimeStamp;
+                Update?.Invoke(delta);
```
Perksey: Update may have to be interlocked because update (once upon a time could be, we intend readd it) ran on another thread, it's only render that doesn't have to be interlocked.

thargy: `Update?.Invoke((-Interlocked.Exchange(ref _updateTimestamp, Stopwatch.GetTimestamp()) + _updateTimestamp) / Frequency);` isn't thread-safe anyway, as it accesses `_updateTimestamp` twice non-atomically.  Making this code thread-safe is actually quite complex as there is a check before change.  I'll submit as a separate commit to show a thread-safe approach.

src/Windowing/Silk.NET.Windowing.Common/Internals/ViewImplementationBase.cs
```diff
@@ -178,15 +186,38 @@ public void DoRender()
...
+             * The first check ensures that enough time has passed to perform an update, if it fails we return.
+             *
+             * If it succeeds then we try to update the timestamp, however, note that we may not 'win' if
+             * another thread updates the timestamp before us.  If we fail, we don't raise the Update event 
+             * as it will have been called by the winning thread.
+             *
+             * However, if updatePeriod <=0, there's an assumption that Update events will be called every time DoUpdate()
+             * is called.  As such we, only return if updatePeriod is positive and we failed.
+             */
+            if ((newTimestamp - updateTimestamp) < updatePeriod ||
```
thargy, line 211: Normally we'd loop on a failure to update the timestamp. However, we can leverage the nature of the Update event to avoid looping by recognizing that (when `updatePeriod` > 0) there's no guarantee that Update is to be called for every `DoUpdate()` call, indeed we're more concerned with not calling _more than_ `UpdatesPerSecond`.  This allows a thread-safe implementation without lock usage.

HurricanKai:
@Perksey can we get this into 2.1?

src/Windowing/Silk.NET.Windowing.Common/Interfaces/IViewProperties.cs
```diff
@@ -54,6 +55,8 @@ public interface IViewProperties
         /// <summary>
         /// Whether or not VSync is enabled for this view.
         /// </summary>
+        /// <remarks>This value will set the SwapInterval for OpenGL, but on other frameworks (e.g. Vulkan) you will need to turn on
+        /// VSync manually.  Whilst <see langword="true"/>, the value of <see cref="FramesPerSecond"/> is ignored, to prevent frame skipping.</remarks>
```
Perksey, line 59: 120 character rule.

Perksey, line 59:
```suggestion
        /// <remarks>This value will set the SwapInterval for OpenGL, but on other frameworks (e.g. Vulkan) you will need to turn on
        /// VSync manually using that framework. Whilst <see langword="true"/>, the value of <see cref="FramesPerSecond"/> is ignored, to prevent frame skipping.</remarks>

```

Perksey (changes requested):
It looks okay, but I'm gonna get @john-h-k to look over this as he knows more about graphics, game loops, and micro-optimisation than I do.

src/Windowing/Silk.NET.Windowing.Common/Internals/ViewImplementationBase.cs
```diff
@@ -120,14 +118,22 @@ public void Reset()
...
+            set => _renderPeriod = value <= double.Epsilon ? 0 : (long) (Frequency / value);
         }
 
         public double UpdatesPerSecond
         {
-            get => _updatePeriod <= double.Epsilon ? 0 : 1 / _updatePeriod;
-            set => _updatePeriod = value <= double.Epsilon ? 0 : 1 / value;
+            get
+            {
+                var updatePeriod = Interlocked.Read(ref _updatePeriod);
```
john-h-k, line 129: This read shouldn't be interlocked, it makes no sense if all writes to it are interlocked

src/Windowing/Silk.NET.Windowing.Common/Internals/ViewImplementationBase.cs
Perksey: Seems fine.

Perksey:
To be honest this seems like a lot of added congnitive complexity for not much gain. If our timer logic in 2.0 is broken (which I think it might be) then yeah I'm fine with merging this just fix the review comments, but I kind of want to see benchmarks to see whether this cognitive complexity is worth our while.

src/Windowing/Silk.NET.Windowing.Common/Internals/ViewImplementationBase.cs
```diff
@@ -23,16 +23,18 @@ namespace Silk.NET.Windowing.Internals
...
-        private readonly Stopwatch _renderStopwatch = new Stopwatch();
-        private readonly Stopwatch _updateStopwatch = new Stopwatch();
-        private readonly Stopwatch _lifetimeStopwatch = new Stopwatch();
-        private double _renderPeriod;
-        private double _updatePeriod;
+        private long _renderTimestamp;
+        private long _updateTimestamp;
+        private long _lifetimeTimestamp;
+        private long _renderPeriod;
+        private long _updatePeriod;
```
john-h-k, line 37: Make all these fields `volatile`

john-h-k:
> To be honest this seems like a lot of added congnitive complexity for not much gain. If our timer logic in 2.0 is broken (which I think it might be) then yeah I'm fine with merging this just fix the review comments, but I kind of want to see benchmarks to see whether this cognitive complexity is worth our while.

Agreed. I would be utterly shocked if the previous logic was even close to a bottleneck in anyway, and the new logic seems to at best remove a couple of trivial divisions and multiplications.

Perksey:
So it does look like the 2.0 timer is bugged, so we have to add some fixes and improvements anyway - seeing as we have this today might as well merge it in once the review comments are fixed.

Once again it's pretty cognitively complex (maybe add some more comments?) and I'd probably like to see some benchmarks in the future (perhaps in a Lab project), but seeing as we need to improve the timer for 2.1 this isn't urgent.

Perksey:
Closing due to inactivity.

Issue #407, opened 2021-02-12
Windowing hangs in some scenarios


Perksey:
Likely due to eager creation of platforms leading to both GLFW and SDL being initialised

Perksey:
This was fixed in 2.1.

In some cases, window creation does take around 30 seconds or so but this is a known bug upstream and not related to this original issue. Nothing we can do on our end on that front, but the users who were complaining about the window not starting at all say this is now fixed.

Issue #400, opened 2021-01-05
Missing GL_VERTEX_ARRAY from GLEnum and  the ObjectIdentifier enum


MaikelChan:
# Summary
Trying to use glObjectLabel, I was unable to use VertexArray as identifier. According to specification, these are the identifiers that can be used:

`GL_BUFFER, GL_SHADER, GL_PROGRAM, GL_VERTEX_ARRAY, GL_QUERY, GL_PROGRAM_PIPELINE, GL_TRANSFORM_FEEDBACK, GL_SAMPLER, GL_TEXTURE, GL_RENDERBUFFER, GL_FRAMEBUFFER`

Silk implements correctly those, both in GLEnum and the ObjectIdentifier enum, except for GL_VERTEX_ARRAY which is nowhere to be found.

# Steps to reproduce
- Platform: Desktop
- Framework Version: .NET Core 3
- API: OpenGL
- API Version: OpenGL 4.3 Core
1. Just try to use gl.ObjectLabel() setting VertexArray as identifier.

PR #411, opened 2021-02-13
Bindings improvements for the Februrary 2021 update

Perksey:
# Summary of the PR
- Add native handles for Windowing
- Add Veldrid support
- Add support for macro constants
- Fix some type issues
- Add support for unnamed nested anonymous struct fields
- Add better name hinting for anonymous types
- Fuse nested anonymous structs with their parent structs, allowing direct ref access on .NET Standard 2.1 platforms without going through the anonymous field
- Fix struct field size & type variation between 32-bit and 64-bit in generated C++ bytes
- Fix calling convention detection, and respect the actual calling convention we detect
- Fix being unable to specify calling convention using NativeApiAttribute
- Support OpenGL token reuse after it had been previously deprecated & removed, fixing #400 
- Stop initializing both platforms #407 

**TO BE ADDED**

# Further Comments
**PENDING SELF REVIEW DO NOT REVIEW YET**


build/csharp_typemap.json
```diff
@@ -145,9 +145,20 @@
     "_LUID": "ValueTuple<ulong, long>",
     "_LARGE_INTEGER": "long",
     "IUnknown": "Silk.NET.Core.Native.IUnknown",
-    "tagRECT": "Silk.NET.Maths.Rectangle<long>",
-    "tagPOINT": "Silk.NET.Maths.Vector2D<long>",
-    "tagSIZE": "Silk.NET.Maths.Vector2D<long>",
+    "tagRECT": "Silk.NET.Maths.Rectangle<int>",
+    "tagPOINT": "Silk.NET.Maths.Vector2D<int>",
+    "tagSIZE": "Silk.NET.Maths.Vector2D<int>",
```
HurricanKai, line 150: breaking?

src/Core/Silk.NET.BuildTools/Silk.NET.BuildTools.csproj
```diff
@@ -2,7 +2,7 @@
 
     <PropertyGroup>
         <OutputType>Exe</OutputType>
-        <TargetFramework>netcoreapp3.1</TargetFramework>
+        <TargetFramework>net5.0</TargetFramework>
```
HurricanKai, line 5: :tada:

src/Core/Silk.NET.Core/Contexts/INativeWindow.cs
```diff
@@ -0,0 +1,77 @@
...
+{
+    /// <summary>
+    /// Wraps any underlying native window handles a window (or window-like instance) may have. 
+    /// </summary>
+    public interface INativeWindow
+    {
+        /// <summary>
+        /// Gets a bitmask enum representing the underlying platforms powering this window.
+        /// </summary>
+        public NativeWindowFlags Kind { get; }
```
HurricanKai, line 16: IMO types that have a Kind property that basically just indicates what type they really are should just be multiple types.
We have great pattern matching in C# now, so just (empty) INativeWindow and then `INativeWindow[X11/Wayland/WinRT/UIKit/Win32/Vivante/Android/Glfw] : INativeWindow` would be much preferred here.

src/Core/Silk.NET.Core/Native/ComPtr.cs
```diff
@@ -43,13 +43,17 @@ public uint Release()
...
-            }
-        }
+        // 1. get a ref for this, which is a readonly ref when within a readonly function
+        // 2. interpret that ref as a reference to a T. it's actually a ref to a T* (well actually it's a ComPtr<T>),
+        //    but let's roll with it as it's as good as we're gonna get without having pointers as generic type
+        //    parameters
+        // 3. get a T* from that "reference to a T".
+        // 4. cast it to a T**, which is its true identity.
+        public readonly T** GetAddressOf()
+            => (T**) Unsafe.AsPointer(ref Unsafe.As<ComPtr<T>, T>(ref Unsafe.AsRef(in this)));
```
HurricanKai, line 53: isn't the middle Unsafe.As redundant? I believe that's just erased by the pointer cast?

src/Core/Silk.NET.Core/Native/ComPtr.cs
```diff
@@ -43,13 +43,17 @@ public uint Release()
...
+
+        public readonly ref T* GetPinnableReference() => ref *GetAddressOf();
+        public readonly ref T Get() => ref *Handle;
```
HurricanKai, line 56: Unsafe.AsRef?

src/Core/Silk.NET.Core/Native/SilkMarshal.cs
```diff
@@ -697,5 +720,52 @@ public static void ThrowHResult(int hResult)
...
+
+        [DllImport("kernel32", EntryPoint = "WaitForSingleObjectEx", ExactSpelling = true,  SetLastError = true)]
+        private static extern uint CoreWaitObjects(nint hHandle, uint dwMilliseconds, int bAlertable);
+
+        [DllImport("kernel32", EntryPoint = "WaitForMultipleObjectsEx", ExactSpelling = true, SetLastError = true)]
+        private static extern unsafe uint CoreWaitObjects
+            (uint nCount, nint* lpHandles, int bWaitAll, uint dwMilliseconds, int bAlertable);
+            
+        [DllImport("kernel32", EntryPoint = "CloseHandle", ExactSpelling = true, SetLastError = true)]
+        public static extern int CloseWindowsHandle(nint hObject);
```
HurricanKai, line 736: this doesn't mess with perf?

src/Core/Silk.NET.Core/Native/Structs/ID3D10Blob.gen.cs
```diff
@@ -22,6 +22,8 @@ namespace Silk.NET.Core.Native
     [NativeName("Name", "ID3D10Blob")]
     public unsafe partial struct ID3D10Blob
     {
+        public static readonly Guid Guid = new("8ba5fb08-5195-40e2-ac58-0d989c3a0102");
```
HurricanKai, line 25: this is defined behaviour right? like there is a spec thing or something? BuildTools isn't just making up GUIDs here?

src/Windowing/Silk.NET.GLFW/GlfwNativeWindow.cs
```diff
@@ -0,0 +1,73 @@
...
+{
+    public struct GlfwNativeWindow : INativeWindow
+    {
+        private const int GwlpHInstance = -6;
+    
+        [DllImport("user32", EntryPoint = "GetDC")]
+        private static extern nint Win32GetDC(nint hwnd);
+
+        [DllImport("user32", EntryPoint = "GetWindowLongPtrA")]
+        private static extern nint Win32GetWindowLongPtr(nint hwnd, int index);
```
HurricanKai, line 19: Could really just make user32 / kernel32 bindings at some point 🤔

src/Windowing/Silk.NET.Windowing.Common/Interfaces/IView.cs
```diff
@@ -12,7 +12,7 @@ namespace Silk.NET.Windowing
     /// <summary>
     /// Represents a window view.
     /// </summary>
-    public interface IView : IViewProperties, IGLContextSource, IVkSurfaceSource, IDisposable
+    public interface IView : IViewProperties, IGLContextSource, IVkSurfaceSource, IDisposable, INativeWindowSource
```
HurricanKai, line 15: Technically breaking, but I guess this should be fine.
Also note that if the approach I suggested above is adopted, this isn't breaking anymore.

HurricanKai (changes requested):

build/csharp_typemap.json
Perksey: Yes, but a bug fix. Our bindings literally won't work without this breaking change.

src/Core/Silk.NET.Core/Contexts/INativeWindow.cs
Perksey: This would introduce much more bloat than required: first there'd have to be lots of different interfaces in Silk.NET.Core, then there'd have to be lots of different types in both GLFW and SDL, and then the windowing backends will have to pick the correct type based on the current context. The last one isn't too much of a worry but the others you'd be exploding the codebase from 3 types to 27 types.

src/Core/Silk.NET.Core/Native/ComPtr.cs
Perksey: Good point
```suggestion
            => (T**) Unsafe.AsPointer(ref Unsafe.AsRef(in this));
```

src/Core/Silk.NET.Core/Native/ComPtr.cs
Perksey: Makes no difference and would require a needless type specification as this is not inferred from the pointer type.

src/Core/Silk.NET.Core/Native/SilkMarshal.cs
Perksey: These are fully blittable so no.

src/Core/Silk.NET.Core/Native/Structs/ID3D10Blob.gen.cs
Perksey: Each interface has an IID defined in the headers which is where we get the GUID.

src/Windowing/Silk.NET.GLFW/GlfwNativeWindow.cs
Perksey: Probably! Will probably be 3.0 though after a BuildTools rewrite - I don't have a lot of faith in it at the moment.

src/Windowing/Silk.NET.Windowing.Common/Interfaces/IView.cs
Perksey: Should be fine.

build/csharp_typemap.json
HurricanKai: Fair Enough

src/Core/Silk.NET.Core/Contexts/INativeWindow.cs
HurricanKai: 🤔 That is a fair point, but not sure if type count is such an important argument.
I'm primarily concerned about how we would annotate this in .net 6, as we can't somehow platform annotate based on a Kind, just the actual type.

Perksey: Sure, but we can annotate the properties as being specific to a certain platform. The kind enum is just another way of being able to tell which platform which properties are "available" on.

HurricanKai: That's a bit of an odd pattern though, isn't it? `SDL` and `Android` are both possibly available on Android, but also possibly not. So one would have to check whether they are on Android *and* check for both of them being non-null.
Guess this is fair enough for now though, .net 6 will be 3.0 anyways.

HurricanKai (approved):

Issue #412, opened 2021-02-14
Testing epic


Perksey:
# Development
## Tests
- [ ] [In Progress] GLFW tests - @HurricanKai 
- [ ] [In Progress] Maths tests - See maths/testing
    - [ ] Scalar.Abs<T>(T x) -> T - _unclaimed_
    - [ ] Scalar.Acos<T>(T x) -> T - _unclaimed_
    - [ ] Scalar.Acosh<T>(T x) -> T - _unclaimed_
    - [x] Scalar.Add<T>(T left, T right) -> T - @HurricanKai 
    - [ ] Scalar.As<TFrom, TTo>(TFrom val) -> TTo - _unclaimed_
    - [ ] Scalar.Asin<T>(T x) -> T - _unclaimed_
    - [ ] Scalar.Asinh<T>(T x) -> T - _unclaimed_
    - [ ] Scalar.Atan2<T>(T y, T x) -> T - _unclaimed_
    - [ ] Scalar.Atan<T>(T x) -> T - _unclaimed_
    - [ ] Scalar.Atanh<T>(T x) -> T - _unclaimed_
    - [ ] Scalar.Cbrt<T>(T x) -> T - _unclaimed_
    - [ ] Scalar.Ceiling<T>(T x) -> T - _unclaimed_
    - [ ] Scalar.Cos<T>(T x) -> T - _unclaimed_
    - [ ] Scalar.Cosh<T>(T x) -> T - _unclaimed_
    - [ ] Scalar.Divide<T>(T left, T right) -> T - _unclaimed_
    - [ ] Scalar.Equal<T>(T left, T right) -> bool - _unclaimed_
    - [ ] Scalar.Exp<T>(T x) -> T - _unclaimed_
    - [ ] Scalar.Floor<T>(T x) -> T - _unclaimed_
    - [ ] Scalar.GreaterThan<T>(T left, T right) -> bool - _unclaimed_
    - [ ] Scalar.GreaterThanOrEqual<T>(T left, T right) -> bool - _unclaimed_
    - [ ] Scalar.IEEERemainder<T>(T x, T y) -> T - _unclaimed_
    - [ ] Scalar.IsFinite<T>(T f) -> bool - _unclaimed_
    - [ ] Scalar.IsInfinity<T>(T f) -> bool - _unclaimed_
    - [ ] Scalar.IsNaN<T>(T f) -> bool - _unclaimed_
[...]

PR #413, opened 2021-02-14
Math Tests

HurricanKai:
# Summary of the PR
See #412


Perksey:
Going to close this as not much work is going into it, and will likely become out-of-date very quickly once 3.0 dev starts.

The diff will be preserved by GitHub should we need to recover this small start!

PR #433, opened 2021-03-20
Coalesce the NativePackages and SilkExtensions repositories

### LINKED ISSUE #404: Silk extensions no longer work with 2.0
I had been using the ImGuiController included in the extensions repo, this doesn't work with 2.0 as IWindow is no longer an IView? Is there any chance this can be updated?

Also would be cool if you're planning to merge these extensions in the future to make it a little more generic as the current implementation will only work with OpenGL

Perksey:
- [x] NativePackages
- [x] SilkExtensions

Fixes #404 
Supersedes and closes Ultz/SilkExtensions#3


HurricanKai (approved):
Assuming this is fine.

build/props/common.props
```diff
@@ -38,4 +38,8 @@
     <ItemGroup Condition="'$(SilkEnableSourceLink)' != '' And '$(SilkNativePackage)' == ''">
         <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All"/>
     </ItemGroup>
+    <PropertyGroup Condition="'$(SilkNativePackage)' == 'true'">
```
HurricanKai, line 41: nit: either we use `'$(A)' != ''` as "A is set" or we use `$(A)' == 'true'`

Perksey: Noted
```suggestion
    <PropertyGroup Condition="'$(SilkNativePackage)' != ''">
```

Perksey:
@HurricanKai checks are passing, are you happy with this? If so I'll merge without squash so we can preserve the history I merged in from the other repos.

HurricanKai:
Seems fine

Issue #431, opened 2021-03-19
GLFW is broken on linux


Dinolek:
# Summary
GLFW is broken on linux(any non-windows platform?)

# Steps to reproduce
- Platform: Desktop
- Framework Version: .NET Core 3
- API: Vulkan/OpenGL
- API Version: Vulkan 1.2/OpenGL 4.3 Legacy/OpenGL 4.5 Core/OpenGLES 3.2
1. Clone [main branch](https://github.com/dotnet/Silk.NET/tree/78d9c6a22f75ac362cf0fe200702a0d8e5be600e)
2. dotnet run -p src/Lab/Triangle/ --framework netcoreapp3.1

```
Unhandled exception. Silk.NET.Core.Loader.SymbolLoadingException: Native symbol not found (Symbol: glfwGetWin32Window)
   at Silk.NET.Core.Loader.LibraryLoader.ThrowSymbolLoading(String functionName) in /home/dinolek/Desktop/Silk.NET/src/Core/Silk.NET.Core/Loader/LibraryLoader.cs:line 260
   at Silk.NET.Core.Loader.LibraryLoader.LoadFunctionPointer(IntPtr handle, String functionName) in /home/dinolek/Desktop/Silk.NET/src/Core/Silk.NET.Core/Loader/LibraryLoader.cs:line 252
   at Silk.NET.Core.Loader.UnmanagedLibrary.LoadFunction(String name) in /home/dinolek/Desktop/Silk.NET/src/Core/Silk.NET.Core/Loader/UnmanagedLibrary.cs:line 119
   at Silk.NET.Core.Contexts.DefaultNativeContext.GetProcAddress(String proc, Nullable`1 slot) in /home/dinolek/Desktop/Silk.NET/src/Core/Silk.NET.Core/Contexts/DefaultNativeContext.cs:line 89
   at Silk.NET.GLFW.GlfwNativeWindow..ctor(Glfw api, WindowHandle* window) in /home/dinolek/Desktop/Silk.NET/src/Windowing/Silk.NET.GLFW/GlfwNativeWindow.cs:line 25
   at Silk.NET.Windowing.Glfw.GlfwWindow.GetNativeWindow() in /home/dinolek/Desktop/Silk.NET/src/Windowing/Silk.NET.Windowing.Glfw/GlfwWindow.cs:line 663
[...]

PR #440, opened 2021-03-26
Fix issue #431

Perksey:


src/Core/Silk.NET.Core/Contexts/LamdaNativeContext.cs
```diff
@@ -2,21 +2,50 @@
...
+                Throw(proc);
+            }
+
+            return pfn;
+            static void Throw(string proc) => throw new SymbolLoadingException(proc);
+        }
+
+        public bool TryGetProcAddress(string proc, out nint addr, int? slot = default)
+        {
+            return (addr = _getProcAddress(proc)) != 0;
```
HurricanKai, line 48: double indirection if a TryLoader was passed? Shouldn't that be the default case, and therefore should be optimized for?

src/Core/Silk.NET.Core/Contexts/MultiNativeContext.cs
```diff
@@ -42,6 +42,21 @@ public nint GetProcAddress(string proc, int? slot = default)
...
+            addr = 0;
+            foreach (var nativeContext in Contexts)
+            {
+                if (nativeContext?.TryGetProcAddress(proc, out addr, slot) ?? false)
+                {
+                    return true;
+                }
+            }
+            
+            return false;
```
HurricanKai, line 57: consider zeroing the addr at the end - just to be consistent in all and every case.

src/Core/Silk.NET.SilkTouch/NativeContextOverrideHelper.cs
```diff
@@ -0,0 +1,332 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using static Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
+
+namespace Silk.NET.SilkTouch
+{
+    public static class NativeContextOverrideHelper
+    {
+        public static MethodDeclarationSyntax GetProcAddress => HelperMethods.Value.GetProcAddress;
```
HurricanKai, line 14: this lazy seems entirely pointless. SyntaxFactory isn't very expensive and this code is impossible to read (in part) because of the lazy + tuple combination.
Additionally, I'd prefer if this was cleaned up again, roslyn quoter does a great job, but often isn't very smart about constructor use, and this massively bloats the code size in cases like this.

src/Windowing/Silk.NET.SDL/SdlContext.cs
```diff
@@ -102,7 +103,34 @@ public void Dispose()
...
         {
             AssertCreated();
-            return (nint) _sdl.GLGetProcAddress(proc);
+            _sdl.ClearError();
+            var ret = (nint) _sdl.GLGetProcAddress(proc);
+            _sdl.ThrowError();
+            if (ret == 0)
+            {
+                throw new SymbolLoadingException(proc);
+            }
```
HurricanKai, line 112: not sure, but inlining?

src/Windowing/Silk.NET.SDL/SdlContext.cs
```diff
@@ -102,7 +103,34 @@ public void Dispose()
...
+        }
+
+        public bool TryGetProcAddress(string proc, out nint addr, int? slot = default)
+        {
+            addr = 0;
+            _sdl.ClearError();
+            if (_ctx is null)
+            {
+                return false;
+            }
```
HurricanKai, line 124: why do we care for _ctx here?

src/Windowing/Silk.NET.Windowing.Glfw/GlfwWindow.cs
```diff
@@ -521,7 +522,21 @@ public override void Dispose()
...
+            {
+                Throw(proc);
+            }
+
+            return ret;
+            static void Throw(string proc) => throw new SymbolLoadingException(proc);
+        }
+
+        public bool TryGetProcAddress(string proc, out nint addr, int? slot = default)
+            => (addr = _glfw.GetProcAddress(proc)) != 0;
```
HurricanKai, line 539: Not an expert, but doesn't GLFW.ThrowExceptions(); throw a managed exception, so this could still throw an exception?

src/Core/Silk.NET.Core/Native/SilkMarshal.cs
```diff
@@ -10,13 +10,14 @@
...
 using System.Threading;
+using Silk.NET.Core.Loader;
 
 namespace Silk.NET.Core.Native
 {
     /// <summary>
     /// A helper class that marshals between native and managed types.
     /// </summary>
-    public static class SilkMarshal
+    public static partial class SilkMarshal
```
HurricanKai, line 20: Maybe I'm blind, but why the partial?

src/Windowing/Silk.NET.GLFW/GlfwContext.cs
```diff
@@ -28,7 +29,22 @@ public unsafe GlfwContext(Glfw glfw, WindowHandle* window, IGLContextSource? sou
...
+                Throw(proc);
+            }
+
+            return ret;
+            static void Throw(string proc) => throw new SymbolLoadingException(proc);
+        }
+
+        /// <inheritdoc />
+        public bool TryGetProcAddress(string proc, out nint addr, int? slot = default)
+            => (addr = _glfw.GetProcAddress(proc)) != 0;
```
HurricanKai, line 47: > Not an expert, but doesn't GLFW.ThrowExceptions(); throw a managed exception, so this could still throw an exception?

HurricanKai (changes requested):
It seems to me like code got a bit inconsistent with GetProcAddress and CoreGetProcAddress, can you clarify what the change is? (I think (?) it's just in the SilkTouch emitted code, but then it should at least also be named that in SilkTouch variables)

src/Windowing/Silk.NET.SDL/SdlContext.cs
Perksey: We can't load OpenGL functions without a valid OpenGL context.

src/Windowing/Silk.NET.Windowing.Glfw/GlfwWindow.cs
Perksey: GLFW.ThrowExceptions isn't called here, but on Windows this effectively implicitly called so the point still stands, will fix.

src/Windowing/Silk.NET.SDL/SdlContext.cs
HurricanKai: While that makes sense logically, I don't get it from a code perspective. Seems like we could call into SDL just fine without OpenGL

src/Windowing/Silk.NET.SDL/SdlContext.cs
Perksey: Doubt this would be inlined, but wouldn't hurt to fix.

src/Core/Silk.NET.Core/Contexts/LamdaNativeContext.cs
Perksey: > double indirection if a TryLoader was passed?

Likely.

> Shouldn't that be the default case, and therefore should be optimized for?

LambdaNativeContext is pretty much always used with a delegate that can return 0.

src/Core/Silk.NET.Core/Contexts/MultiNativeContext.cs
Perksey: Good shout!

src/Core/Silk.NET.Core/Native/SilkMarshal.cs
Perksey: No idea, will remove.

src/Windowing/Silk.NET.GLFW/GlfwContext.cs
Perksey: Will fix.

src/Windowing/Silk.NET.SDL/SdlContext.cs
Perksey: Possibly, but then we'll possibly do a completely unnecessary native call which isn't cheap, and for a Try method - a defensive method style - it makes sense to put in defensive logic in here.

Perksey: It's also an intentional design choice - we don't want any of the methods on the context object to be "alive" until the context that is represents is also "alive". Other methods do this using `AssertCreated` which throws if `ctx` is `null`, while this one eliminates the throw.

src/Core/Silk.NET.SilkTouch/NativeContextOverrideHelper.cs
Perksey: > this lazy seems entirely pointless. SyntaxFactory isn't very expensive and this code is impossible to read (in part) because of the lazy + tuple combination.

Perhaps not performance-wise, but it's a lot of unnecessary memory stress otherwise.

> Additionally, I'd prefer if this was cleaned up again, roslyn quoter does a great job, but often isn't very smart about constructor use, and this massively bloats the code size in cases like this.

Well I don't know what's unnecessary and what's not, this isn't really my domain.

HurricanKai: Fair enough. I'll clean it up afterwards; if you'd like to give it a shot, it's pretty much just "look what you can give the constructor and give it that" x1000 - but yeah, probably necessary to pick that up.

HurricanKai (approved):

PR #454, opened 2021-04-10
SilkTouch work for 2.4

HurricanKai:
- Removes ConcurrentDictionaryVTable
- few minor changes to core types
- VTable overhaul
- slots removed
- general cleanup
- improved cashing branching
- fixed a bug regarding device function pointer loading in Vulkan
- Output many files in Release mode (Disabled)


HurricanKai:
Changelog:
- Removes ConcurrentDictionaryVTable
- few minor changes to core types
- VTable overhaul
- slots removed
- general cleanup
(it may be useful to review this per-commit)

src/Core/Silk.NET.Core/Native/NativeApiContainer.cs
```diff
@@ -46,10 +36,9 @@ public void Dispose()
             CurrentVTable.Dispose();
         }
 
-        protected virtual int CoreGetSlotCount() => 0;
         protected virtual int CoreGcSlotCount() => 0;
-        protected virtual IVTable CreateVTable() => new ConcurrentDictionaryVTable();
-        protected IVTable SwapVTable() => Interlocked.Exchange(ref _vTable, CreateVTable());
+        protected abstract IVTable CreateVTable();
```
Perksey, line 40: Can we make this be virtual but throw instead? I don't want this to be visible to the user and this will likely result in a "abstract method not implemented" warning by most static analysis tools such as IntelliSense and ReSharper.

```suggestion
        protected virtual IVTable CreateVTable()
            => throw new NotImplementedException("Must be implemented by SilkTouch");
```

Perksey (changes requested):

Perksey (approved):

src/Core/Silk.NET.Core/Native/NativeApiContainer.cs
```diff
@@ -72,11 +63,22 @@ public void PurgeEntryPoints()
...
         {
-            return CurrentVTable.Load(slot, entryPoint);
+            return Load(entryPoint);
         }
 
+        protected nint Load(string entryPoint)
+        {
+            var result = CurrentVTable.Load(entryPoint);
+            if (result == 0)
+                ThrowSymbolLoadingEx(entryPoint);
```
Perksey, line 76:
```suggestion
            if (result == 0)
            {
                ThrowSymbolLoadingEx(entryPoint);
            }
```

Perksey (changes requested):

Perksey (approved):

HurricanKai:
Converting this back to a draft since a couple more work items have occurred, I'll just bundle them in this PR

src/Vulkan/Silk.NET.Vulkan/Vk.cs
```diff
@@ -67,67 +68,15 @@ public static Vk GetApi()
...
-            fixed (InstanceCreateInfo* infoPtr = &info)
-            {
-                fixed (Instance* instancePtr = &instance)
-                {
-                    ret.CreateInstance(infoPtr, null, instancePtr);
-                }
-            }
-
+            var ret = GetApi();
+            instance = default;
```
Perksey, line 72: Is this necessary?

Perksey (approved):

src/Vulkan/Silk.NET.Vulkan/Vk.cs
HurricanKai: no, just simplifying code?

Perksey: No as in do we have to initialize the instance variable? Surely this happens when we pass it along...

HurricanKai: I thought I had just carried that over

Perksey (changes requested):
Unapproving until we can investigate the massive regression in build time.

HurricanKai:
image
Well NormalizeWhitespace isn't optimized enough for us to do multi-file yet. Unfortunate, I'll look for better options eventually. This PR should be good as is anyways.

Perksey (approved):

PR #456, opened 2021-04-11
Add a Rider/ReSharper plugin

Perksey:
More words will be here in due course.


Perksey:
I think the next steps for this is fixing up the overloads so that they're less specific/hardcoded, and then cleaning up the code a bit (i.e. removing commented out stuff). I'm happy to ship with the functionality we have now given those two points.

Perksey:
The longest-standing PR currently open today is now ready for review.

Perksey:
In case the reader is not aware: we have decided not to include codefixes at this time due to time constraints on the maintenance team. We definitely want this however and will accept community contributions for adding them.

HurricanKai:
(Code fixes are expected for 3.0)

Perksey:
Will try and revisit this for 2.16, sorry this is taking so long.

Perksey:
I'm closing this as I won't be coming back to this, and with all maintainers being dedicated to 3.0 I'm not sure if anyone else would either. I've protected the branch, so it should stay put should anyone else want to pick this up.

PR #457, opened 2021-04-12
Hotfix/tutorial tweaks

Toby-Lawrance:
# Summary of the PR
Two changes to the tutorial example code: 
The removal of superfluous Use() calls from the tutorial Shader SetUniform definitions. These can cause confusion when debugging. (Use calls the shaders UseProgram)

Assorted small spelling and grammar tweaks, primarily commas, apostrophes and the occasional "tho" -> "though" 

Perksey was happy for me to make this pull request. 
https://discord.com/channels/521092042781229087/607634593201520651/831162386195021845

# Further Comments

### DIFF
--- a/examples/CSharp/Tutorial 1.4 - Textures/Shader.cs
+++ b/examples/CSharp/Tutorial 1.4 - Textures/Shader.cs
@@ -42,7 +42,6 @@ namespace Tutorial
             {
                 throw new Exception($"{name} uniform not found on shader.");
             }
-            Use();
             _gl.Uniform1(location, value);
         }
 
@@ -53,7 +52,6 @@ namespace Tutorial
             {
                 throw new Exception($"{name} uniform not found on shader.");
             }
-            Use();
             _gl.Uniform1(location, value);
         }
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,6 +5,7 @@ on:
     paths:
       - 'src/**/*'
       - 'build/nuke/*'
+      - 'examples/**/*'
     branches:
       - 'main'
       - 'release/*'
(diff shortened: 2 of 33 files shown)


examples/CSharp/Tutorial 1.4 - Textures/Shader.cs
```diff
@@ -53,7 +52,6 @@ public void SetUniform(string name, float value)
             {
                 throw new Exception($"{name} uniform not found on shader.");
             }
-            Use();
```
Perksey, line 56: Have you made sure we've definitely called UseProgram before we use any of the SetUniform functions? While we don't need to do it every call, we need to at least make sure it's called once prior to the uniform calls.

Perksey (commented):
Awesome, thanks so much for this contribution! See below some questions (the PR looks good as is, just want to do some sanity checks).

Finally, have you ran through these tutorials after your changes to make sure they work? A simple `dotnet run` should suffice, don't bother opening the Silk.NET.sln.

examples/CSharp/Tutorial 1.4 - Textures/Shader.cs
Toby-Lawrance: Yes, in the OnRender method for each tutorial (or RenderLitCube/RenderLamp for a couple of the later ones) we have a Shader.Use() call immediately before SetUniform

Toby-Lawrance:
> 
> 
> Awesome, thanks so much for this contribution! See below some questions (the PR looks good as is, just want to do some sanity checks).
> 
> Finally, have you ran through these tutorials after your changes to make sure they work? A simple `dotnet run` should suffice, don't bother opening the Silk.NET.sln.

I have indeed called `dotnet run` on them to ensure they all still work, no issues I can spot.

examples/CSharp/Tutorial 1.4 - Textures/Shader.cs
Perksey: Awesome, thanks.

Perksey (approved):

PR #474, opened 2021-05-02
Add OpenAL Wave Player Demo

HurricanKai:


examples/openal/WavePlayer/Program.cs
```diff
@@ -0,0 +1,174 @@
+﻿using System;
+using System.Buffers.Binary;
+using System.Collections.Generic;
+using System.IO;
+using System.Runtime.CompilerServices;
+using System.Text;
+using Silk.NET.OpenAL;
+
+namespace WavePlayer
```
Perksey, line 9: This is more lab material than an actual example. Unless you think you can make a web tutorial explaining everything in this program, it goes in the lab.

If we wanted to look at fleshing this out such that we can create an extension method in Silk.NET.OpenAL itself, then I'd reckon we'd be able to make this an example.

examples/openal/WavePlayer/WavePlayer.csproj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+    <PropertyGroup>
+        <OutputType>Exe</OutputType>
+        <TargetFramework>net5.0</TargetFramework>
+        <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <PackageReference Include="Silk.NET.OpenAL" Version="2.3.0" />
```
Perksey, line 10: ?

Perksey (changes requested):

examples/openal/WavePlayer/Program.cs
HurricanKai: Pretty sure I could explain everything here. imo it's a fine sample for playing basic audio. sure the wave loading is quite a lot, but imo this will work fine explaining the basics. 
If you want I'll add more comments explaining what each bit does?

examples/openal/WavePlayer/WavePlayer.csproj
HurricanKai: Honestly not sure what we chose for the examples, should I instead reference directly?

examples/openal/WavePlayer/Program.cs
Perksey: I just don't think it'd be a good example to make the user do this for themselves.

I personally think we should look at putting all the WAV logic into the library itself, such that you can just do something like BufferWAV or whatever.

That way, the example is a lot less significant. The examples are meant to demonstrate Silk.NET, and there's a lot of non-Silk.NET in here atm.

examples/openal/WavePlayer/WavePlayer.csproj
Perksey: Yeah we do ProjectReferences (or should prefer to)

examples/openal/WavePlayer/Program.cs
HurricanKai: We can only move the WAV logic into Silk if it works perfectly, and while this will work great, I don't think it's perfect / should be used in production.
Also, imo we should have more samples that are just neat demos of stuff you could do, not necessarily tutorials. (maybe we make a folder "tutorials" and a folder "demos"?)

Perksey: > I don't think it's perfect / should be used in production.

Ding ding ding we have a lab application! To be fair though, I think we should add a WAV loader in OpenAL in any case as it'll be useful for a lot of people and will be something a lot of production users would want to use. But to be fair I don't want to put code we think isn't perfect in the examples folder - it is expected that the examples folder demonstrates best practice.

> (maybe we make a folder "tutorials" and a folder "demos"?)

Maybe, that could get confusing though and would complicate the structure we have atm. I don't really want to change the structure we have at the moment, so it'll have to be `example/Demos` and no further changes which you can decide whether you're cool with. Having two root-level "exemplary" folders would be confusing.

HurricanKai: well I'd argue they are still best practice, just not perfect, wav is an old format and there are 1000000 extensions to it. I mean I can move it to lab, but imo mixing the reallllly lab apps and the perfectly working lab apps isn't great

Perksey: Yeah I think we should move it to the Lab until such a time comes we can make this handle all sorts of cowboy WAV files, and subsequently ship it with our library.

HurricanKai: That time will never come lol
image
these are all the possible formats, and we can only easily support PCM

Perksey: I'm sure we can do some of the Creative ones too, just probably won't be able to test them.

But if that's the case, let's make our PCM support solid enough to ship and throw if it's not PCM!

Perksey:
I think I've decided how the folder structure should be.

- examples
    - OpenGL Tutorials (renamed from OpenGL)
        - CSharp
            - Tutorial 1...... 
    - (maybe in the future) OpenGL Demos
    - (maybe in the future) OpenAL Tutorials
    - OpenAL Demos
        - CSharp
            - SimpleWavPlayer
    - etc...

HurricanKai:
I'm happy with that, though I want to suggest swapping the CSharp folder to the outside, because it makes browsing multiple projects in the same folder easier. In your example:
- examples
  - CSharp
    - OpenGL Tutorials
      - Tutorial 1...
    - (maybe in the future) OpenGL Demos
    - etc.
If you think having the language folder inside is better, I'm ok with that too.

Perksey:
> If you think having the language folder inside is better, I'm ok with that too.

I think it's best as having it on the outside adds more clicks (currently needlessly)

examples/CSharp/OpenAL/WavePlayer/Program.cs
```diff
@@ -0,0 +1,174 @@
+﻿using System;
```
Perksey, line 1: Should be in `examples/CSharp/OpenAL Demos`

Perksey (changes requested):

examples/openal/WavePlayer/Program.cs
Perksey: Yeah we'll keep this as a demo for now.

I do want to do this in the future though!

Perksey (approved):

PR #484, opened 2021-05-07
More HLU robustness improvements

Perksey:


src/Windowing/Silk.NET.Windowing.Common/Internals/ViewImplementationBase.cs
```diff
@@ -213,6 +213,52 @@ public bool VSync
...
+                        "OpenGL functions can only be used after initialization (just before the Load callback is " +
+                        "executed)"
+                    );
+
+                    Throw();
+                }
+                else if (API.API != ContextAPI.OpenGL && API.API != ContextAPI.OpenGLES)
+                {
+                    return null;
+                }
```
HurricanKai, line 233: I think throwing even if Vk is requested makes sense. It removes some checks here, and makes more sense from the callers side.

HurricanKai: Fine with me.

HurricanKai (approved):

PR #498, opened 2021-05-23
Fix loading some OpenAL extensions

Perksey:


Perksey:
@HurricanKai whoops this really should've gone in 2.5 but forgot about it. Has been ready for review.

src/OpenAL/Silk.NET.OpenAL/AL/AL.cs
```diff
@@ -358,7 +358,8 @@ public static AL GetApi(bool soft = false)
         public TExtension GetExtension<TExtension>()
             where TExtension : NativeExtension<AL>
         {
-            return IsExtensionPresent(ExtensionAttribute.GetExtensionAttribute(typeof(TExtension)).Name)
+            var extAttr = ExtensionAttribute.GetExtensionAttribute(typeof(TExtension));
+            return (extAttr is null || IsExtensionPresent(extAttr.Name))
```
HurricanKai, line 362: that check doesn't look right? Do we want to create extension types without an attribute?

HurricanKai (changes requested):

src/OpenAL/Silk.NET.OpenAL/AL/AL.cs
Perksey: There was a reason why I thought it was necessary, but I've since forgotten it. Reason #5435 for commenting. Removed!

HurricanKai (approved):

Perksey:
OML HOW DID THIS MISS ANOTHER FREAKING UPDATE

PR #505, opened 2021-06-04
This is the June 2021 Update (v2.5.0) with production-ready OpenXR

### LINKED ISSUE #500: OpenXR Bindings / API cannot be used
# Summary
Trying to get the XRApi via GetApi() causes an exception, rendering the whole binding system unusable.

System.ArgumentNullException: 'Value cannot be null. Parameter name: key'
This exception was originally thrown at this call stack:
    System.Collections.Concurrent.ConcurrentDictionary<TKey, TValue>.TryAdd(TKey, TValue)
    Silk.NET.OpenXR.XR.PostInit()
    Silk.NET.OpenXR.XR.XR(Silk.NET.Core.Contexts.INativeContext)
    Silk.NET.OpenXR.XR.GetApi()

# Steps to reproduce
Just try to call XR.GetApi()

# Comments
This is where I'm trying to get the API in my engine:
https://github.com/phr00t/FocusEngine/blob/master/sources/engine/Xenko.VirtualReality/OpenXR/OpenXRHmd.cs#L142

Discord Kai (Beast) said:
"that's a big error :eyes: @Perksey apparently we try to insert a null key into the concurrent dictionary"

Perksey added:
"I don’t think there’s anything you can do to work around this at the moment. The fix will involve making the key of the VTable dictionary something that isn’t null (could probably just make the key non-nullable and use default?)"

### LINKED ISSUE #501: Fix null layer not producing nullptr in the XR extension loading system
Using `null` for the `layer` parameter of `IsInstanceExtensionPresent` or `TryGetInstanceExtension` doesn't send `nullptr` as expected to native. Instead, it sends a pointer to an empty string.

This should be fixed, ideally with appropriate nullability annotations applied.

Perksey:
Will close #501
Will close #500


build/props/common.props
```diff
@@ -6,30 +6,29 @@
...
-            - Miscellaneous windowing &amp; input bugfixes
+            - OpenXR bindings are now production-ready &amp; no longer experimental!
+            - Add (optional) native binaries for OpenAL Soft
+            - Add support for reversed-z projection matrices in Silk.NET.Maths
+            - Add bindings for a number of missing Vulkan and OpenCL extensions
+            - Add OpenGL-OpenXR interoperability helper APIs
+            - Add more helper APIs in Silk.NET.Core and Silk.NET.Maths
+            - Update to latest OpenGL and OpenGLES specifications
+            - Update to OpenXR 1.0.16
+            - Update to Vulkan 1.2.169
```
HurricanKai, line 19: This doesn't seem right?

build/props/common.props
```diff
@@ -6,30 +6,29 @@
...
+            - Fix SDL OpenGL context issues when not using OpenGL
+            - Fix GLFW native window issues on 32-bit Windows
+            - Fix OpenXR loading issues
+            - Fix incorrect bindings in OpenXR
+            - Miscellaneous windowing &amp;amp; input bugfixes &amp; improvements
```
HurricanKai, line 24: amp;amp;?

generator.json
```diff
@@ -641,7 +643,8 @@
                     "EGLConfig": "nint",
                     "EGLContext": "nint",
                     "GLXFBConfig": "nint",
-                    "VkFormat": "uint"
+                    "VkFormat": "uint",
```
HurricanKai, line 646: Maybe missing context, but that does not seem right at all?

src/Core/Silk.NET.Core/Native/SilkMarshal.cs
```diff
@@ -112,61 +112,111 @@ public static GlobalMemory StringToMemory
...
+                    {
+                        fixed (byte* bytes = span)
+                        {
+                            convertedBytes = Encoding.UTF8.GetBytes(firstChar, input.Length, bytes, span.Length - 1);
+                        }
+                    }
 
-                fixed (char* firstChar = input)
+                    span[convertedBytes] = 0;
+                    return ++convertedBytes;
```
HurricanKai, line 196: Please forward to Utf8.FromUtf16 on >= netcore3.0
Something like
```cs
#if NET5_0
Debug.Assert(Utf8.FromUtf16(input, span, out _, out convertedBytes, true, true) == OperationStatus.Done);
#endif
```
We probably don't care for the returned status, it can only be done, or theoretically, `DestinationTooSmall` but that can only happen if GetMaxSizeOf fails.
(If you don't think this is fit for this PR, do just reference this in a new issue, and I'll get back to it asap, we absolutely need to track this though, perf of the string marshalling is miserable)

src/Core/Silk.NET.Core/Native/SilkMarshal.cs
```diff
@@ -264,6 +314,33 @@ static unsafe string BStrToString(nint ptr)
...
+            {
+                fixed (byte* ptr = input)
+                {
+                    return PtrToString((nint) ptr, e)!;
+                }
+            }
+
+            return PtrToString((nint) Unsafe.AsPointer(ref input.GetPinnableReference()), e)!;
+        }
+
```
HurricanKai, line 343: - e -> encoding 
```suggestion
        /// <summary>
        /// Reads a null-terminated string from the given span, with the given native encoding.
        /// </summary>
        /// <param name="input">A span containing a null-terminated string.</param>
        /// <param name="pin">
        /// Whether to pin the span such that pointers taken from it are not subject to GC movement.
        /// </param>
        /// <param name="e">The encoding of the string in memory.</param>
        /// <returns>The string read from memory.</returns>
        public static unsafe string SpanToString
        (
            Span<byte> input,
            bool pin = false,
            NativeStringEncoding encoding = NativeStringEncoding.Ansi
        )
        {
            if (pin) // "should" be optimized away if constant
            {
                fixed (byte* ptr = input)
                {
                    return PtrToString((nint) ptr, encoding)!;
                }
            }

            return PtrToString((nint) Unsafe.AsPointer(ref input.GetPinnableReference()), encoding)!;
        }
        
```
- I don't think the pin parameter is needed, and it's suuuper dangerous to default it to false. I'd prefer we get rid of id and always pin, it's a nop on pinned things anyways.
- We need to improve our nullability annotations cleary, nit.

src/Maths/Silk.NET.Maths/Scalar.Constants.cs
```diff
@@ -87,6 +87,22 @@ public static partial class Scalar<T> where T : unmanaged
...
+        /// </remarks>
+        public static readonly T DegreesPerRadian;
+
+        /// <summary>
+        /// Represents the number of radians in a single degree.
+        /// </summary>
+        /// <remarks>
+        /// This is equivalent to <c>MathF.Pi / 180</c>.
+        /// </remarks>
+        public static readonly T RadiansPerDegree;
```
HurricanKai, line 104: These seem to be missing from the PublicAPI.Shipped.txt

src/OpenXR/Extensions/Silk.NET.OpenXR.Extensions.FB/FBSwapchainUpdateState.gen.cs
```diff
@@ -0,0 +1,37 @@
...
+
+namespace Silk.NET.OpenXR.Extensions.FB
+{
+    [Extension("XR_FB_swapchain_update_state")]
+    public unsafe partial class FBSwapchainUpdateState : NativeExtension<XR>
+    {
+        public const string ExtensionName = "XR_FB_swapchain_update_state";
+        /// <summary>To be documented.</summary>
+        [NativeApi(EntryPoint = "xrUpdateSwapchainFB")]
+        public unsafe partial Result UpdateSwapchainFB([Count(Count = 0)] Swapchain swapchain, [Count(Count = 0), Flow(FlowDirection.In)] SwapchainStateBaseHeaderFB* state);
```
HurricanKai, line 25: Just to note this, why is Count = 0 here? Consider referencing in new issue.

src/OpenXR/Silk.NET.OpenXR/NativeWindowExtensions.cs
```diff
@@ -0,0 +1,150 @@
...
+using System.Runtime.InteropServices;
+using Silk.NET.Core;
+using Silk.NET.Core.Contexts;
+
+namespace Silk.NET.OpenXR
+{
+    public static class NativeWindowExtensions
+    {
+        public static GraphicsBinding CreateOpenGLBinding<T>(this T src)
+            where T : INativeWindowSource, IGLContextSource
```
HurricanKai, line 14: Should I know what INativeWindowSource is? lmao

src/OpenXR/Silk.NET.OpenXR/Structs/GraphicsRequirements.cs
```diff
@@ -0,0 +1,73 @@
...
+        private readonly T Cast<T>(StructureType assertType)
+        {
+            if (Type != assertType)
+            {
+                static void Throw() => throw new InvalidCastException();
+                Throw();
+            }
+
+            return Unsafe.As<GraphicsRequirements, T>(ref Unsafe.AsRef(in this));
+        }
```
HurricanKai, line 53: What is this?!

src/OpenXR/Silk.NET.OpenXR/Structs/GraphicsBinding.cs
```diff
@@ -0,0 +1,100 @@
...
+        private readonly T Cast<T>(StructureType assertType)
+        {
+            if (Type != assertType)
+            {
+                static void Throw() => throw new InvalidCastException();
+                Throw();
+            }
+
+            return Unsafe.As<GraphicsBinding, T>(ref Unsafe.AsRef(in this));
+        }
```
HurricanKai, line 68: What is this?!

src/Vulkan/Silk.NET.Vulkan/Enums/ImageLayout.gen.cs
```diff
@@ -55,7 +49,7 @@ public enum ImageLayout : int
         [NativeName("Name", "VK_IMAGE_LAYOUT_FRAGMENT_DENSITY_MAP_OPTIMAL_EXT")]
         FragmentDensityMapOptimalExt = 1000218000,
         [NativeName("Name", "VK_IMAGE_LAYOUT_FRAGMENT_SHADING_RATE_ATTACHMENT_OPTIMAL_KHR")]
-        FragmentShadingRateAttachmentOptimalKhr = 1000226003,
+        FragmentShadingRateAttachmentOptimalKhr = 1000164003,
```
HurricanKai, line 52: That's not good. Is this just khronos trying to find out how many people using fragment shaind rate or is this our mistake?

src/Windowing/Silk.NET.Windowing.Common/Structs/ViewOptions.cs
```diff
@@ -40,10 +40,10 @@ Vector2D<int> IViewProperties.Size
         public int? PreferredDepthBufferBits { get; set; }
 
         /// <inheritdoc />
-        public int? PreferredStencilBufferBits { get; }
+        public int? PreferredStencilBufferBits { get; set; }
         
         /// <inheritdoc />
-        public Vector4D<int>? PreferredBitDepth { get; }
+        public Vector4D<int>? PreferredBitDepth { get; set; }
```
HurricanKai, line 46: Why?

src/Windowing/Silk.NET.Windowing.Common/Structs/WindowOptions.cs
```diff
@@ -61,7 +61,7 @@ public WindowOptions(ViewOptions opts)
         public int? PreferredStencilBufferBits { get; set; }
 
         /// <inheritdoc />
-        public Vector4D<int>? PreferredBitDepth { get; }
+        public Vector4D<int>? PreferredBitDepth { get; set; }
```
HurricanKai, line 64: Why?

HurricanKai (changes requested):
I know were late already, but got a couple of notes that I think need fixing.

src/Core/Silk.NET.Core/Native/SilkMarshal.cs
Perksey: > I don't think the pin parameter is needed, and it's suuuper dangerous to default it to false. I'd prefer we get rid of id and always pin, it's a nop on pinned things anyways.

Not on all .NET Standard platforms! Nonetheless, consider it done.

src/Windowing/Silk.NET.Windowing.Common/Structs/WindowOptions.cs
Perksey: Why not?

src/Windowing/Silk.NET.Windowing.Common/Structs/ViewOptions.cs
Perksey: Why not?

src/Vulkan/Silk.NET.Vulkan/Enums/ImageLayout.gen.cs
Perksey: https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkImageLayout.html

src/Windowing/Silk.NET.Windowing.Common/Structs/WindowOptions.cs
HurricanKai: We don't document what these do. What do they do anyways?

Perksey: They indicate your preferred red, green, blue, and alpha bits of the window's OpenGL framebuffer.

HurricanKai: Well, I'm fine with it then. Not happy, but fine.

src/Core/Silk.NET.Core/Native/SilkMarshal.cs
Perksey: Opened an issue.

src/Maths/Silk.NET.Maths/Scalar.Constants.cs
Perksey: They are not.

src/OpenXR/Silk.NET.OpenXR/Structs/GraphicsBinding.cs
Perksey: Per the documentation, GraphicsBinding is a union to address the fact there are so many GraphicsBinding structures, and to allow for abstractions.

Per the code, Cast is a utility function used by the explicit operators where we want to cast it back to an explicit type so we can check the structure type as well to check that it's a valid cast.

src/OpenXR/Silk.NET.OpenXR/Structs/GraphicsRequirements.cs
Perksey: See other instance.

src/OpenXR/Silk.NET.OpenXR/NativeWindowExtensions.cs
Perksey: Is this a review comment?

src/OpenXR/Extensions/Silk.NET.OpenXR.Extensions.FB/FBSwapchainUpdateState.gen.cs
Perksey: Same as in all Vulkan and OpenXR bindings. `Count = 0` should be treated as there being no count at all.

src/OpenXR/Silk.NET.OpenXR/NativeWindowExtensions.cs
HurricanKai: Quick explainer what this does/where it's used? I see this type being used the first time in our codebase? seems like it's OpenXR specific?

generator.json
Perksey: It is.

src/Vulkan/Silk.NET.Vulkan/Enums/ImageLayout.gen.cs
HurricanKai: I guess we used to convert this the wrong way. Resolved.

src/OpenXR/Silk.NET.OpenXR/NativeWindowExtensions.cs
Perksey: INativeWindowSource is an existing type which is used to describe a type that can be represented using a INativeWindow, an interface for getting platform-specific window handles from an object representing a window.

build/props/common.props
Perksey: Was a typo for 179, latest commit regenerates and updates to 180.

build/props/common.props
Perksey: Fixed in latest commit.

HurricanKai (changes requested):
A project has not been included in the solution and will not be shipped! "D:\a\Silk.NET\Silk.NET\src\Vulkan\Extensions\Silk.NET.Vulkan.Extensions.FB\Silk.NET.Vulkan.Extensions.FB.csproj" if this is acceptable please add the project name (excluding the path and extension) to the AllowedExclusions array in the NUKE Build.cs file.

After that, looks good to go

Perksey:
Lmfao well glad to know that's doing its job.

Perksey:
@HurricanKai can you approve so we can release?

HurricanKai (approved):

PR #520, opened 2021-06-11
Remove the main NormalizeWhitespace call in favour of our own spacing

Perksey:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,7 +31,8 @@ jobs:
       run: nuke ValidateSolution
     - name: Pack
       # TODO build native mixins such as BuildLibSilkDroid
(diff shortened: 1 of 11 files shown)


Perksey:
This can be reviewed, but again we need to decide how worth it this is.

HurricanKai:
I did have a quick glance at it, but I'm not a fan. This puts liability on anyone editing SilkTouch code to track whitespace. One of the most important reasons for us using ExpressionSyntax and friends everywhere is that it's a neat abstraction for anything, by design it's not intended to know in what context it's going to be used, so tracking whitespace is impossible / has to be very conservative with how it works. I also think the Roslyn team may do something about this, will raise with them again.

Perksey:
OK fair enough.

PR #525, opened 2021-06-13
Add a pipeline which ignores merge conflicts in the 3.0 rewrite branch

Perksey:

PR #528, opened 2021-06-18
Follow up from #525

Perksey:


HurricanKai (approved):
I don't get why this is needed but I'll trust 😅

PR #554, opened 2021-07-28
Multi-Backend Input for Silk.NET 3.0

Perksey:
This enables input to use multiple backends (i.e. so we can have windowing-based input as well as something like OpenXR) and contains other refactorings for 3.0.

Unlike previous 3.0 proposals this proposes only a minor change of the public API.

### DIFF
--- a/documentation/proposals/Proposal - Multi-Backend Input.md
+++ b/documentation/proposals/Proposal - Multi-Backend Input.md
@@ -139,11 +139,21 @@ namespace Silk.NET.Input
 +   /// <summary>
 +   /// Represents an input backend from which input devices can be retrieved.
 +   /// </summary>
++   /// <remarks>
++   /// This interface is not intended for general consumption. Consider <see cref="InputContext" /> instead.
++   /// </remarks>
 +   public interface IInputBackend : IDisposable
 +   {
 +       /// <summary>
 +       /// Gets all devices of the given type recognised by this backend.
 +       /// </summary>
++       /// <remarks>
++       /// The <typeparamref name="T" /> type parameter must be the last interface in the inheritance heirarchy (the "device type")
++       /// i.e. it must be a device type like <see cref="IJoystick" /> and not something like <see cref="IInputDevice" />.
++       /// The behaviour of using a T that isn't a device type recognised by the backend is undefined.<br />
++       /// <br />
++       /// This method is not intended for general consumption. Consider <see cref="InputContext" /> instead.
++       /// </remarks>
 +       IReadOnlyList<IInputDevice> GetDevices<T>() where T : IInputDevice;
 +
 +       /// <summary>
--- a/documentation/proposals/Proposal - Multi-Backend Input.md
+++ b/documentation/proposals/Proposal - Multi-Backend Input.md
@@ -23,15 +23,31 @@ Proposal API for backend-agnostic, refactored Input via keyboards, mice, and con
 
 ## Reference Implementation
 
-Similar to Windowing 3.0, a reference implementation will be included in the main `Silk.NET.Input` package which uses the same API or family of APIs as Windowing.
-**TODO:** define how this is exposed.
+Similar to Windowing 3.0, a reference implementation will be included in the main `Silk.NET.Input` package which uses the same API or family of APIs as Windowing. This will be exposed via the `InputBackend` class. The `InputBackend` class contains a static `Create` method which accepts the native handle for the input context on the particular underlying API i.e. on GLFW this handle represents a `GLFWwindow*`. This may be subject to change.
 
 ## Source Generator
 
 Alongside the input package there will be a source generator (shipped in a `Silk.NET.Input.Roslyn` NuGet package, which is referenced by the main input package). This source generator will be very small, but it allows us to create a link between windowing and input at compile time without a hard reference.
 
-If the source generator detects that both `Silk.NET.Input` and `Silk.NET.Windowing` are referenced, 
-**TODO: ** carry on
+If the source generator detects that both `Silk.NET.Input` and `Silk.NET.Windowing` are referenced, code with the following API surface will be injected into the assembly:
+```cs
+namespace Silk.NET.Input
+{
+    internal static class InputSurfaceExtensions // internal to avoid conflicts with other assemblies
+    {
+         public static InputContext CreateInput(this ISurface surface);
+    }
+}
+```
+
+The `CreateInput` method will use the surface, obtain its platform data (i.e. by using something like `IGlfwPlatformData`), and then feed that into `InputBackend.Create`. The returned `InputContext` will be configured to have the returned `IInputBackend` as a backend. This method will also appropriately bind the `ISurface.Update` callback:
+- The delegate passed in the `onActivated` parameter of `InputBackend.Create` will bind the `IInputBackend.Update` method to run in the `ISurface.Update` callback.
+- The delegate passed in the `onDeactivated` parameter of `InputBackend.Create` will unbind the `ISurface.Update` callback.
+- These delegates must be called by the reference implementation in its `IInputBackend.Activate` and `IInputBackend.Deactivate` methods (respectively)
+- The `IInputBackend.Activate` and `IInputBackend.Deactivate` methods must be called by the `InputContext.Add` and `InputContext.Remove` methods (respectively)
+
+**KEY POINT FOR WORKING GROUP**: The Windowing-Input integration mandates the use of source generators. Is this ok?
+
 # Proposed API
 
 ```diff
@@ -58,30 +74,72 @@ namespace Silk.NET.Input
         IReadOnlyList<IMouse> Mice { get; }
         IReadOnlyList<IInputDevice> OtherDevices { get; }
 +       IReadOnlyList<IInputBackend> Backends { get; }
-+       void AddBackend(IInputBackend backend);
-+       void RemoveBackend(IInputBackend backend);
         event Action<IInputDevice, bool>? ConnectionChanged;
-    }
-}
-```
-
-```cs
-namespace Silk.NET.Input
-{
-    public interface IInputBackend
-    {
-        IReadOnlyList<IInputDevice> ConnectedDevices { get; }
-        event Action<IInputDevice, bool>? ConnectionChanged;
++       void Add(IInputBackend backend);
++       void Remove(IInputBackend backend);
++       void Update();
     }
 }
 ```
 
 ```diff
--namespace Silk.NET.Input
--{
--    public static class InputWindowExtensions
--    {
--        // ...
--    }
--}
+namespace Silk.NET.Input
+{
++   public interface IInputBackend
++   {
++       /// <summary>
++       /// Gets all connected devices recognised by this backend.
++       /// </summary>
++       IReadOnlyList<IInputDevice> ConnectedDevices { get; }
++
++       /// <summary>
++       /// Raised when an input device is connected or disconnected.
++       /// </summary>
++       event Action<IInputDevice, bool>? ConnectionChanged;
++
++       /// <summary>
++       /// Updates this backend's input data.
++       /// </summary>
++       /// <remarks>
++       /// Input events and changes are permitted to occur outside of this method.
++       /// </remarks>
++       void Update();
++
++       /// <summary>
++       /// Activates this input backend and runs any necessary prerequisites.
++       /// </summary>
++       /// <remarks>
++       /// Called by <see cref="InputContext.Add" />
++       /// </remarks>
++       void Activate();
++
++       /// <summary>
++       /// Deactivates this input backend and tears down any resources created.
++       /// </summary>
++       /// <remarks>
++       /// Called by <see cref="InputContext.Remove" />
++       /// </remarks>
++       void Deactivate();
++   }
+}
+```
+
+```diff
+namespace Silk.NET.Input
+{
+-   public static class InputWindowExtensions
+-   {
+-       // ...
+-   }
+}
+```
+
+```diff
+namespace Silk.NET.Input
+{
++   public static class InputBackend
++   {
++       public static IInputBackend Create(void* handle, Action? onActivate = null, Action? onDeactivate = null);
++   }
+}
 ```
--- a/documentation/proposals/Proposal - Multi-Backend Input.md
+++ b/documentation/proposals/Proposal - Multi-Backend Input.md
@@ -48,6 +48,8 @@ The `CreateInput` method will use the surface, obtain its platform data (i.e. by
 
 The API surface for this is defined later in the Proposed API section.
 
+One instance of an `IInputBackend` can only belong to one input context at a time. If multiple input contexts are in use, multiple backend instances should also be used. An input context is set as active/in use by a context using the `Activate` and `Deactivate` methods. `Activate` should throw an exception if already activated, and `Deactivate` should throw an exception if not already activated.
+
 **KEY POINT FOR WORKING GROUP**: The Windowing-Input integration mandates the use of source generators. Is this ok?
 
 # Proposed API
--- a/documentation/proposals/Proposal - Multi-Backend Input.md
+++ b/documentation/proposals/Proposal - Multi-Backend Input.md
@@ -43,12 +43,10 @@ namespace Silk.NET.Input
 The `CreateInput` method will use the surface, obtain its platform data (i.e. by using something like `IGlfwPlatformData`), and then feed that into `InputBackend.Create`. The returned `InputContext` will be configured to have the returned `IInputBackend` as a backend. This method will also appropriately bind the `ISurface.Update` callback:
 - The `IInputBackend.Update` method is configured to run in the `ISurface.Update` callback.
 - The `IInputBackend.Disposing` event will be bound such that the `IInputBackend.Update` method is unbound from `ISurface.Update` event callback.
-- Once a `IInputBackend` is `Add`ed to an `InputContext`, the `InputContext` takes full ownership of the `IInputBackend` and, as a result, will `Dispose` the `IInputBackend` when it is `Remove`d or the context itself is `Dispose`d.
-    - This essentially means that you can count on `Disposing` being fired as an indicator that the `IInputBackend` will never be relevant again. 
 
 The API surface for this is defined later in the Proposed API section.
 
-One instance of an `IInputBackend` can only belong to one input context at a time. If multiple input contexts are in use, multiple backend instances should also be used. An input context is set as active/in use by a context using the `Activate` and `Deactivate` methods. `Activate` should throw an exception if already activated, and `Deactivate` should throw an exception if not already activated. Once deactivated, an input backend can be reactivated.
+One instance of an `IInputBackend` can only belong to one `InputContext` throughout the entirety of the `IInputBackend`'s lifetime. Once a `IInputBackend` is `Add`ed to an `InputContext`, the `InputContext` takes full ownership of the `IInputBackend` and, as a result, will `Dispose` the `IInputBackend` when it is `Remove`d or the context itself is `Dispose`d. If multiple `InputContext`s are in use, multiple `IInputBackend` instances must also be used. `IInputBackend` should throw an exception if its APIs are used after the backend has been disposed.
 
 **KEY POINT FOR WORKING GROUP**: The Windowing-Input integration mandates the use of source generators. Is this ok?
 
--- a/documentation/proposals/Proposal - Multi-Backend Input.md
+++ b/documentation/proposals/Proposal - Multi-Backend Input.md
@@ -68,7 +68,7 @@ namespace Silk.NET.Input
 +   /// Encapsulates input devices sourced from input backends.
 +   /// </summary>
 -   public interface IInputContext : IDisposable
-+   public class InputContext : IDisposable
++   public sealed class InputContext : IDisposable
     {
 -       nint Handle { get; }
         IReadOnlyList<IGamepad> Gamepads { get; }
--- a/documentation/proposals/Proposal - Multi-Backend Input.md
+++ b/documentation/proposals/Proposal - Multi-Backend Input.md
@@ -48,7 +48,7 @@ The `CreateInput` method will use the surface, obtain its platform data (i.e. by
 
 The API surface for this is defined later in the Proposed API section.
 
-One instance of an `IInputBackend` can only belong to one input context at a time. If multiple input contexts are in use, multiple backend instances should also be used. An input context is set as active/in use by a context using the `Activate` and `Deactivate` methods. `Activate` should throw an exception if already activated, and `Deactivate` should throw an exception if not already activated.
+One instance of an `IInputBackend` can only belong to one input context at a time. If multiple input contexts are in use, multiple backend instances should also be used. An input context is set as active/in use by a context using the `Activate` and `Deactivate` methods. `Activate` should throw an exception if already activated, and `Deactivate` should throw an exception if not already activated. Once deactivated, an input backend can be reactivated.
 
 **KEY POINT FOR WORKING GROUP**: The Windowing-Input integration mandates the use of source generators. Is this ok?
 
@@ -132,7 +132,7 @@ namespace Silk.NET.Input
 +       void Update();
 +
 +       /// <summary>
-+       /// Activates this input backend and runs any necessary prerequisites.
++       /// Activates this input backend and runs any necessary prerequisites to collecting input data.
 +       /// </summary>
 +       /// <remarks>
 +       /// Called by <see cref="InputContext.Add" />
@@ -140,7 +140,7 @@ namespace Silk.NET.Input
 +       void Activate();
 +
 +       /// <summary>
-+       /// Deactivates this input backend and tears down any resources created.
++       /// Deactivates this input backend and destroys/disables any resources responsible for collecting input data.
 +       /// </summary>
 +       /// <remarks>
 +       /// Called by <see cref="InputContext.Remove" />
--- a/documentation/proposals/Proposal - Multi-Backend Input.md
+++ b/documentation/proposals/Proposal - Multi-Backend Input.md
@@ -46,7 +46,7 @@ The `CreateInput` method will use the surface, obtain its platform data (i.e. by
 
 The API surface for this is defined later in the Proposed API section.
 
-One instance of an `IInputBackend` can only belong to one `InputContext` throughout the entirety of the `IInputBackend`'s lifetime. Once a `IInputBackend` is `Add`ed to an `InputContext`, the `InputContext` takes full ownership of the `IInputBackend` and, as a result, will `Dispose` the `IInputBackend` when it is `Remove`d or the context itself is `Dispose`d. If multiple `InputContext`s are in use, multiple `IInputBackend` instances must also be used. `IInputBackend` should throw an exception if its APIs are used after the backend has been disposed.
+One instance of an `IInputBackend` can only belong to one `InputContext` throughout the entirety of the `IInputBackend`'s lifetime. Once a `IInputBackend` is `Add`ed to an `InputContext`, the `InputContext` takes full ownership of the `IInputBackend` and, as a result, will `Dispose` the `IInputBackend` when it is `Remove`d or the context itself is `Dispose`d. If multiple `InputContext`s are in use, multiple `IInputBackend` instances must also be used. `IInputBackend` will throw an exception if its APIs are used after the backend has been disposed.
 
 **KEY POINT FOR WORKING GROUP**: The Windowing-Input integration mandates the use of source generators. Is this ok?
 
--- a/documentation/proposals/Proposal - Multi-Backend Input.md
+++ b/documentation/proposals/Proposal - Multi-Backend Input.md
@@ -124,6 +124,16 @@ namespace Silk.NET.Input
 +       event Action<IInputDevice, bool>? ConnectionChanged;
 +
 +       /// <summary>
++       /// Occurs when the backend is <see cref="Activate" />d.
++       /// </summary>
++       event Action Activated;
++
++       /// <summary>
++       /// Occurs when the backend is <see cref="Deactivate" />d.
++       /// </summary>
++       event Action Deactivated;
++
++       /// <summary>
 +       /// Updates this backend's input data.
 +       /// </summary>
 +       /// <remarks>
--- a/documentation/proposals/Proposal - Multi-Backend Input.md
+++ b/documentation/proposals/Proposal - Multi-Backend Input.md
@@ -41,10 +41,10 @@ namespace Silk.NET.Input
 ```
 
 The `CreateInput` method will use the surface, obtain its platform data (i.e. by using something like `IGlfwPlatformData`), and then feed that into `InputBackend.Create`. The returned `InputContext` will be configured to have the returned `IInputBackend` as a backend. This method will also appropriately bind the `ISurface.Update` callback:
-- The `IInputBackend.Activated` event will be bound such that the `IInputBackend.Update` method is configured to run in the `ISurface.Update` callback.
-- The `IInputBackend.Deactivated` event will be bound such that the `IInputBackend.Update` is unbound from `ISurface.Update`.
-- These delegates must be called by the reference implementation in its `IInputBackend.Activate` and `IInputBackend.Deactivate` methods (respectively)
-- The `IInputBackend.Activate` and `IInputBackend.Deactivate` methods must be called by the `InputContext.Add` and `InputContext.Remove` methods (respectively)
+- The `IInputBackend.Update` method is configured to run in the `ISurface.Update` callback.
+- The `IInputBackend.Disposing` event will be bound such that the `IInputBackend.Update` method is unbound from `ISurface.Update` event callback.
+- Once a `IInputBackend` is `Add`ed to an `InputContext`, the `InputContext` takes full ownership of the `IInputBackend` and, as a result, will `Dispose` the `IInputBackend` when it is `Remove`d or the context itself is `Dispose`d.
+    - This essentially means that you can count on `Disposing` being fired as an indicator that the `IInputBackend` will never be relevant again. 
 
 The API surface for this is defined later in the Proposed API section.
 
@@ -90,11 +90,20 @@ namespace Silk.NET.Input
 +       /// <summary>
 +       /// Adds the given backend to the list of <see cref="Backends" />.
 +       /// </summary>
++       /// <remarks>
++       /// Note that because input contexts take ownership of input backends once added, this will dispose the input backend when
++       /// <see cref="Remove" /> is called or this input context is <see cref="Dispose" />d.
++       /// As such, the input backend will not be usable in subsequent contexts - you will need to instantiate a new one instead.
++       /// </remarks>
 +       void Add(IInputBackend backend);
 +
 +       /// <summary>
 +       /// Removes the given backend from the list of <see cref="Backends" />.
 +       /// </summary>
++       /// <remarks>
++       /// Note that because input contexts take ownership of input backends once added, this will dispose the input backend.
++       /// As such, the input backend will not be usable in subsequent contexts - you will need to instantiate a new one instead.
++       /// </remarks>
 +       void Remove(IInputBackend backend);
 +
 +       /// <summary>
@@ -111,7 +120,7 @@ namespace Silk.NET.Input
 +   /// <summary>
 +   /// Represents an input backend from which input devices can be retrieved.
 +   /// </summary>
-+   public interface IInputBackend
++   public interface IInputBackend : IDisposable
 +   {
 +       /// <summary>
 +       /// Gets all connected devices recognised by this backend.
@@ -124,14 +133,9 @@ namespace Silk.NET.Input
 +       event Action<IInputDevice, bool>? ConnectionChanged;
 +
 +       /// <summary>
-+       /// Occurs when the backend is <see cref="Activate" />d.
++       /// Occurs when the backend is disposing. This can be used as a good indicator that this backend is being removed from an input context.
 +       /// </summary>
-+       event Action Activated;
-+
-+       /// <summary>
-+       /// Occurs when the backend is <see cref="Deactivate" />d.
-+       /// </summary>
-+       event Action Deactivated;
++       event Action Disposing;
 +
 +       /// <summary>
 +       /// Updates this backend's input data.
@@ -140,24 +144,6 @@ namespace Silk.NET.Input
 +       /// Input events and changes are permitted to occur outside of this method.
 +       /// </remarks>
 +       void Update();
-+
-+       /// <summary>
-+       /// Activates this input backend and runs any necessary prerequisites to collecting input data.
-+       /// Once activated, <see cref="Deactivate" /> must be used before the backend can be activated again.
-+       /// </summary>
-+       /// <remarks>
-+       /// Called by <see cref="InputContext.Add" />
-+       /// </remarks>
-+       void Activate();
-+
-+       /// <summary>
-+       /// Deactivates this input backend and destroys/disables any resources responsible for collecting input data.
-+       /// Once deactivated, <see cref="Activate" /> can be used to activate the backend again.
-+       /// </summary>
-+       /// <remarks>
-+       /// Called by <see cref="InputContext.Remove" />.
-+       /// </remarks>
-+       void Deactivate();
 +   }
 }
 ```
--- a/documentation/proposals/Proposal - Multi-Backend Input.md
+++ b/documentation/proposals/Proposal - Multi-Backend Input.md
@@ -46,6 +46,8 @@ The `CreateInput` method will use the surface, obtain its platform data (i.e. by
 - These delegates must be called by the reference implementation in its `IInputBackend.Activate` and `IInputBackend.Deactivate` methods (respectively)
 - The `IInputBackend.Activate` and `IInputBackend.Deactivate` methods must be called by the `InputContext.Add` and `InputContext.Remove` methods (respectively)
 
+The API surface for this is defined later in the Proposed API section.
+
 **KEY POINT FOR WORKING GROUP**: The Windowing-Input integration mandates the use of source generators. Is this ok?
 
 # Proposed API
@@ -176,7 +178,7 @@ namespace Silk.NET.Input
 +       /// <param name="onDeactivate">A delegate to be called when the context is deactivated. May be null.</param>
 +       /// <remarks>
 +       /// This method is implicitly called by the Windowing-Input integration. <br />
-+       /// On desktop, this uses GLFW.
++       /// On desktop, this uses GLFW. The handle refers to a <c>GLFWwindow*</c> if GLFW is in use.
 +       /// </remarks>
 +       public static unsafe IInputBackend Create(void* handle, Action? onActivate = null, Action? onDeactivate = null);
 +   }
--- a/documentation/proposals/Proposal - Multi-Backend Input.md
+++ b/documentation/proposals/Proposal - Multi-Backend Input.md
@@ -133,6 +133,7 @@ namespace Silk.NET.Input
 +
 +       /// <summary>
 +       /// Activates this input backend and runs any necessary prerequisites to collecting input data.
++       /// Once activated, <see cref="Deactivate" /> must be used before the backend can be activated again.
 +       /// </summary>
 +       /// <remarks>
 +       /// Called by <see cref="InputContext.Add" />
@@ -141,9 +142,10 @@ namespace Silk.NET.Input
 +
 +       /// <summary>
 +       /// Deactivates this input backend and destroys/disables any resources responsible for collecting input data.
++       /// Once deactivated, <see cref="Activate" /> can be used to activate the backend again.
 +       /// </summary>
 +       /// <remarks>
-+       /// Called by <see cref="InputContext.Remove" />
++       /// Called by <see cref="InputContext.Remove" />.
 +       /// </remarks>
 +       void Deactivate();
 +   }
--- a/documentation/proposals/Proposal - Multi-Backend Input.md
+++ b/documentation/proposals/Proposal - Multi-Backend Input.md
@@ -64,6 +64,9 @@ namespace Silk.NET.Input
 ```diff
 namespace Silk.NET.Input
 {
++   /// <summary>
++   /// Encapsulates input devices sourced from input backends.
++   /// </summary>
 -   public interface IInputContext : IDisposable
 +   public class InputContext : IDisposable
     {
@@ -73,10 +76,26 @@ namespace Silk.NET.Input
         IReadOnlyList<IKeyboard> Keyboards { get; }
         IReadOnlyList<IMouse> Mice { get; }
         IReadOnlyList<IInputDevice> OtherDevices { get; }
++
++       /// <summary>
++       /// Gets a list of backends from which all input devices on this input context are sourced.
++       /// </summary>
 +       IReadOnlyList<IInputBackend> Backends { get; }
         event Action<IInputDevice, bool>? ConnectionChanged;
++
++       /// <summary>
++       /// Adds the given backend to the list of <see cref="Backends" />.
++       /// </summary>
 +       void Add(IInputBackend backend);
++
++       /// <summary>
++       /// Removes the given backend from the list of <see cref="Backends" />.
++       /// </summary>
 +       void Remove(IInputBackend backend);
++
++       /// <summary>
++       /// Updates all input data on all backends.
++       /// </summary>
 +       void Update();
     }
 }
@@ -85,6 +104,9 @@ namespace Silk.NET.Input
 ```diff
 namespace Silk.NET.Input
 {
++   /// <summary>
++   /// Represents an input backend from which input devices can be retrieved.
++   /// </summary>
 +   public interface IInputBackend
 +   {
 +       /// <summary>
@@ -137,9 +159,26 @@ namespace Silk.NET.Input
 ```diff
 namespace Silk.NET.Input
 {
++   /// <summary>
++   /// Represents the out-of-the-box input backend, which uses a native API to retrieve input backends using a native handle.
++   /// </summary>
++   /// <remarks>
++   /// On desktop, this uses GLFW.
++   /// </remarks>
 +   public static class InputBackend
 +   {
-+       public static IInputBackend Create(void* handle, Action? onActivate = null, Action? onDeactivate = null);
++       /// <summary>
++       /// Creates an instance of the out-of-the-box input backend, which uses a native API to retrieve input backends using
++       /// the given native handle, optionally calling the given delegates when the input backend is activated and deactivated.
++       /// </summary>
++       /// <param name="handle">The native handle created/sourced from the underlying native API used by this input backend.</param>
++       /// <param name="onActivate">A delegate to be called when the context is activated. May be null.</param>
++       /// <param name="onDeactivate">A delegate to be called when the context is deactivated. May be null.</param>
++       /// <remarks>
++       /// This method is implicitly called by the Windowing-Input integration. <br />
++       /// On desktop, this uses GLFW.
++       /// </remarks>
++       public static unsafe IInputBackend Create(void* handle, Action? onActivate = null, Action? onDeactivate = null);
 +   }
 }
 ```
--- a/documentation/proposals/Proposal - Multi-Backend Input.md
+++ b/documentation/proposals/Proposal - Multi-Backend Input.md
@@ -1,0 +1,77 @@
+# Summary
+Proposal API for backend-agnostic, refactored Input via keyboards, mice, and controllers.
+
+# Contributors
+- Dylan Perks (@Perksey)
+
+# Current Status
+- [x] Proposed
+- [ ] Discussed with Working Group
+- [ ] Approved
+- [ ] Implemented
+
+# Design Decisions
+- This proposal refactors the internals of the Input API to be less tied to a single backend, and instead allow multiple backends to be added and removed from an input context at will.
+- This proposal aims to keep the look and feel of the Input APIs intended for general consumption similar to that of 2.0.
+- This proposal assumes knowledge of the following proposals:
+    - Windowing 3.0
+    - Input
+    - Enhanced Input Events
+- This proposal also assumes knowledge of 2.0's Input APIs as it is only a refactor and not a complete redesign of the API as in other proposals. The most noticable differences between the design of Input 2.0 and Input 3.0 are:
+    - Input no longer has a hard bond to Windowing. The integration will remain the same to the end user but will use Source Generators, more on that later. 
+    - Input contexts and the devices therein are no longer interfaces. They are instead structures containing a handle and a backend object. See the proposed API for details.
+    - They may now only be one mouse and one keyboard. This is because no supported operating system actually allows multiple mice or keyboards to be reported individually.
+    - There is now another layer to input: the LLAPI.
+
+## Input Low-Level API (LLAPI)
+
+In order to faciliate the goal of allowing multiple backends, we needed to remove the close tie between the platform and the high-level input API. To do this, this proposal adds a low-level API to be stored in the `Silk.NET.Input.Internals` namespace.
+
+**OPEN QUESTION:** what should we call this namespace? `Silk.NET.Input.Internals` is used as a placeholder.
+
+### Device Handles
+
+Each kind of device which allows multiple instances (i.e. everything but keyboard and mouse as discussed earlier) will have a handle specific to that backend. For example, there will be:
+- `JoystickHandle`
+- `GamepadHandle`
+
+Both of which containing a `ulong` handle allocated by the backend.
+
+
+# Proposed API
+- Here you do some code blocks, this is the heart and soul of the proposal. DON'T DO ANY IMPLEMENTATIONS! Just declarations.
+
+```cs
+namespace Silk.NET.Inout.Internals
+{
+    public interface IInputBackend
+    {
+        void QueryDevices<TDeviceHandle>(out int numDevices) where TDeviceHandle : unmanaged;
+        void QueryDevices<TDeviceHandle>(Span<TDeviceHandle> deviceHandles) where TDeviceHandle : unmanaged;
+        void QueryDevice<TDeviceHandle, TDataType>(TDeviceHandle device, DeviceDataKind key, out TDataType value) where TDeviceHandle : unmanaged where TDataType : unmanaged;
+        void QueryGlobal<TDataType>(GlobalDataKind key, out TDataType value) where TDataType : unmanaged;
+        void Update();
+    }
+}
+```
+
+```cs
+/// <summary>
+/// Represents a grenade.
+/// </summary>
+public class SilkGrenade
+{
+    /// <summary>
+    /// What does this do?!?
+    /// </summary>
+    public void RedButton();
+    /// <summary>
+    /// Detonates the grenade.
+    /// </summary>
+    protected void Kaboom();
+    /// <summary>
+    /// Gets or sets whether this grenade is disarmed or not.
+    /// </summary>
+    public bool Disarmed { get; set; }
+}
+```
--- a/documentation/proposals/Proposal - Multi-Backend Input.md
+++ b/documentation/proposals/Proposal - Multi-Backend Input.md
@@ -20,6 +20,7 @@ Proposal API for backend-agnostic, refactored Input via keyboards, mice, and con
 - This proposal also assumes knowledge of 2.0's Input APIs as it is only a refactor and not a complete redesign of the API as in other proposals. The most noticable differences between the design of Input 2.0 and Input 3.0 are:
     - Input no longer has a hard bond to Windowing. The integration will remain the same to the end user but will use Source Generators, more on that later. 
     - Input contexts are no longer interfaces, they are instead classes which contain input backends.
+- Unlike 1.0 and 2.0, all device lists (including those on `InputContext`) will return both connected and disconnected devices when indexed or iterated.
 
 ## Reference Implementation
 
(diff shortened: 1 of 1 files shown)


Perksey:
This is ready for pre-meeting review.

HurricanKai:
Please rebase proposal/3.0-input-llapi to not include all the 2.7 changes.

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -0,0 +1,184 @@
...
+    - Input no longer has a hard bond to Windowing. The integration will remain the same to the end user but will use Source Generators, more on that later. 
+    - Input contexts are no longer interfaces, they are instead classes which contain input backends.
+
+## Reference Implementation
+
+Similar to Windowing 3.0, a reference implementation will be included in the main `Silk.NET.Input` package which uses the same API or family of APIs as Windowing. This will be exposed via the `InputBackend` class. The `InputBackend` class contains a static `Create` method which accepts the native handle for the input context on the particular underlying API i.e. on GLFW this handle represents a `GLFWwindow*`. This may be subject to change.
+
+## Source Generator
+
+Alongside the input package there will be a source generator (shipped in a `Silk.NET.Input.Roslyn` NuGet package, which is referenced by the main input package). This source generator will be very small, but it allows us to create a link between windowing and input at compile time without a hard reference.
```
HurricanKai, line 30: Why in a separate package? at this point I'm pretty confident one can ship both in one package.

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -0,0 +1,184 @@
...
+{
+    internal static class InputSurfaceExtensions // internal to avoid conflicts with other assemblies
+    {
+         public static InputContext CreateInput(this ISurface surface);
+    }
+}
+```
+
+The `CreateInput` method will use the surface, obtain its platform data (i.e. by using something like `IGlfwPlatformData`), and then feed that into `InputBackend.Create`. The returned `InputContext` will be configured to have the returned `IInputBackend` as a backend. This method will also appropriately bind the `ISurface.Update` callback:
+- The delegate passed in the `onActivated` parameter of `InputBackend.Create` will bind the `IInputBackend.Update` method to run in the `ISurface.Update` callback.
```
HurricanKai, line 44: the proposal hasn't yet outlined what `InputBackend.Create`s signature is.

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -0,0 +1,184 @@
...
+- The delegate passed in the `onDeactivated` parameter of `InputBackend.Create` will unbind the `ISurface.Update` callback.
+- These delegates must be called by the reference implementation in its `IInputBackend.Activate` and `IInputBackend.Deactivate` methods (respectively)
```
HurricanKai, line 46: IInputBackend? What is that? Activate? Deactivate? What is that?

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -0,0 +1,184 @@
...
+- The `IInputBackend.Activate` and `IInputBackend.Deactivate` methods must be called by the `InputContext.Add` and `InputContext.Remove` methods (respectively)
```
HurricanKai, line 47: InputContext? what is that?

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -0,0 +1,184 @@
...
+```
+
+```diff
+namespace Silk.NET.Input
+{
++   /// <summary>
++   /// Encapsulates input devices sourced from input backends.
++   /// </summary>
+-   public interface IInputContext : IDisposable
++   public class InputContext : IDisposable
```
HurricanKai, line 71: abstract/sealed? I think this is intended to be abstract with `Backends` abstract as well?

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -0,0 +1,184 @@
...
+```
+
+```diff
+namespace Silk.NET.Input
+{
++   /// <summary>
++   /// Represents the out-of-the-box input backend, which uses a native API to retrieve input backends using a native handle.
++   /// </summary>
++   /// <remarks>
++   /// On desktop, this uses GLFW.
```
HurricanKai, line 166: I thought we were multi-backend now?

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -0,0 +1,184 @@
...
++   /// </remarks>
++   public static class InputBackend
++   {
++       /// <summary>
++       /// Creates an instance of the out-of-the-box input backend, which uses a native API to retrieve input backends using
++       /// the given native handle, optionally calling the given delegates when the input backend is activated and deactivated.
++       /// </summary>
++       /// <param name="handle">The native handle created/sourced from the underlying native API used by this input backend.</param>
++       /// <param name="onActivate">A delegate to be called when the context is activated. May be null.</param>
++       /// <param name="onDeactivate">A delegate to be called when the context is deactivated. May be null.</param>
```
HurricanKai, line 176: Why are these designed this way, why not events?

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -0,0 +1,184 @@
...
++       /// <remarks>
++       /// This method is implicitly called by the Windowing-Input integration. <br />
++       /// On desktop, this uses GLFW.
++       /// </remarks>
++       public static unsafe IInputBackend Create(void* handle, Action? onActivate = null, Action? onDeactivate = null);
```
HurricanKai, line 181: This is just very very type-unsafe. How is anyone supposed to know what to pass as `handle` especially considering the only documentation is "On desktop this uses GLFW"?

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -0,0 +1,184 @@
...
++       /// <summary>
++       /// Activates this input backend and runs any necessary prerequisites.
++       /// </summary>
++       /// <remarks>
++       /// Called by <see cref="InputContext.Add" />
++       /// </remarks>
++       void Activate();
++
++       /// <summary>
++       /// Deactivates this input backend and tears down any resources created.
```
HurricanKai, line 139: > tears down any resources created
why not `Dispose` + `IDisposable`?
Is this supposed to be de- and re-activated at will? That seems like a pointless complication.
Why not activate -> ctor and deactivate -> `Dispose`, the cost of recreating an instance seems very low compared to how often this should be done.

HurricanKai (changes requested):
See comments.
Also, please elaborate a bit on how these new types are supposed to work. I'm just massively confused what the interactions are and what the design is here.

documentation/proposals/Proposal - Multi-Backend Input.md
Perksey: Yes it has, look at the bottom of the page in the Proposed API

documentation/proposals/Proposal - Multi-Backend Input.md
Perksey: Look at the bottom of the page in the Proposed API

documentation/proposals/Proposal - Multi-Backend Input.md
Perksey: Look at the bottom of the page in the Proposed API

documentation/proposals/Proposal - Multi-Backend Input.md
HurricanKai: Fair enough, I just ment in the written section. I usually like to read the proposals top-to-bottom without having to scroll 1000x times to understand what is going on.

documentation/proposals/Proposal - Multi-Backend Input.md
Perksey: This should be sealed.
```suggestion
+   public sealed class InputContext : IDisposable
```

documentation/proposals/Proposal - Multi-Backend Input.md
Perksey: This isn't really intended for general consumption. I'm personally not a fan of changing the public API depending on TFM, but then again this is also not intended for general consumption so I guess I can adjust this.

documentation/proposals/Proposal - Multi-Backend Input.md
Perksey: It is intended that backends can be passed between contexts. Will update docs to indicate my intentions.

documentation/proposals/Proposal - Multi-Backend Input.md
HurricanKai: I don't want to change the public API depending on TFM either, but I'd like some documentation on what it does. I think right now you are the only one that knows😅

documentation/proposals/Proposal - Multi-Backend Input.md
Perksey: The event is invoked by the input context and relate to the input context, which isn't really a great fit for events seeing as they can't be (and are not designed to be) invoked outside of the object containing the event. I also don't think we should force use of events like this, as this is really only needed by the reference implementation.

documentation/proposals/Proposal - Multi-Backend Input.md
Perksey: We are but I'm not in a position to comment on other backends right now.

HurricanKai: Ok but then why the documentation notice? Does the user care that this uses GLFW? Should they?

Perksey: No, but this is a remark i.e. informative text

documentation/proposals/Proposal - Multi-Backend Input.md
Perksey: Makes sense.

documentation/proposals/Proposal - Multi-Backend Input.md
Perksey: I'm not entirely convinced this is possible, they're structured differently and even if we can NuGet will not be able to distinguish dependencies.

documentation/proposals/Proposal - Multi-Backend Input.md
Perksey: Updated docs.

documentation/proposals/Proposal - Multi-Backend Input.md
Perksey: Put a sentence in saying that the api surface is defined later.

documentation/proposals/Proposal - Multi-Backend Input.md
Perksey: Put a sentence in saying that the api surface is defined later.

documentation/proposals/Proposal - Multi-Backend Input.md
Perksey: Put a sentence in saying that the api surface is defined later.

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -0,0 +1,190 @@
...
+- The delegate passed in the `onActivated` parameter of `InputBackend.Create` will bind the `IInputBackend.Update` method to run in the `ISurface.Update` callback.
+- The delegate passed in the `onDeactivated` parameter of `InputBackend.Create` will unbind the `ISurface.Update` callback.
+- These delegates must be called by the reference implementation in its `IInputBackend.Activate` and `IInputBackend.Deactivate` methods (respectively)
+- The `IInputBackend.Activate` and `IInputBackend.Deactivate` methods must be called by the `InputContext.Add` and `InputContext.Remove` methods (respectively)
+
+The API surface for this is defined later in the Proposed API section.
+
+One instance of an `IInputBackend` can only belong to one input context at a time. If multiple input contexts are in use, multiple backend instances should also be used. An input context is set as active/in use by a context using the `Activate` and `Deactivate` methods. `Activate` should throw an exception if already activated, and `Deactivate` should throw an exception if not already activated. Once deactivated, an input backend can be reactivated.
```
ThomasMiz, line 51: > An input context is set as active/in use by a context using the `Activate` and `Deactivate` methods
Do you mean an IInputBackend? InputContext doesn't have activate/deactivate methods

It's very unclear to me what the InputContext.Add and Remove do and how it behaves with the backend's activate/deactivate. What happens if I add two backends to an InputContext and activate both, or deactivate both?

Perksey: So the idea is that the average user doesn't actually use `Activate` and `Deactivate`, and that `InputContext.Add` and `InputContext.Remove` implicitly call this. If you try to add an already-activated input context, you can expect an exception. The idea is that an input backend can only be "activated for" one input context at a given time.

Perksey:
Can I get a follow-up on this and resolve discussions if you think they're fine?

documentation/proposals/Proposal - Multi-Backend Input.md
Perksey: Added a description of what the handle is on GLFW. Other platforms TBD.

documentation/proposals/Proposal - Multi-Backend Input.md
ThomasMiz: When would these get called from outside the containing object? From what I understand these should only get invoked by the containing ```IInputBackend``` when it is activated/deactivated

documentation/proposals/Proposal - Multi-Backend Input.md
HurricanKai: This is my main concern with this right now - I don't think this Activate/Deactivate makes a lot of sense.
When enquiring about this in a different place you said
> It is intended that backends can be passed between contexts. Will update docs to indicate my intentions.
Apart from me being too stupid to find where you updated this, I don't think that makes sense. Clearly the input context "owns" the backend for the time between Add and Remove.
So why have this? Why not make Add/Remove generic and pass in a type (I'm assuming only one of each type can be activated at once) and instanciate it inside the InputContext and dipose it when InputContext is disposed or the backend is removed. The Activate/Deactivate semantics seem like an unnecessary complication to me.

HurricanKai: great. can't have two conversations on one line

documentation/proposals/Proposal - Multi-Backend Input.md
Perksey: 3 days later and, yes this makes sense as an event. Changing.

documentation/proposals/Proposal - Multi-Backend Input.md
Perksey: The main reason for Activate and Deactivate is for external control i.e. you have one callback which, when called, you can unbind unrelated resources which will potentially outlive your backend (such as a `ISurface`)

The same could be achieved by exposing appropriate events on `InputContext`, but to use the `ISurface` example once again, we'd need to bind said event to bind the `Update` event when it's added, and then bind the removal event to both remove the `Update` event binding as well as unbind the callback from within the callback, which seems messy. The ownership of a backend is a backend-level construct, not a context-level construct, and we'd have to make a lot of logical compromises to have it defined at the latter.

> Apart from me being too stupid to find where you updated this

iirc we're talking on the line which was inserted to update it.

HurricanKai: > The main reason for Activate and Deactivate is for external control i.e. you have one callback which, when called, you can unbind unrelated resources which will potentially outlive your backend (such as a ISurface)

I don't understand what this is trying to say?
No one can ever call Activate/Deactivate when using the backend with InputContext. So what is meant with external control?

ThomasMiz: If there can only be one backend in use at a time for a given input context, why not just make it a property?

Otherwise if you add many backends to a context, how would you change which one is active between them? Wouldn't we also have to change the IKeyboard/IMouse/IGamepad/etc lists when the backend is changed?

Perksey: > If there can only be one backend in use at a time for a given input context, why not just make it a property?

The whole idea of this proposal is that you can have multiple backends on an input context, but just one context for an input backend [object].

> I don't understand what this is trying to say?
No one can ever call Activate/Deactivate when using the backend with InputContext. So what is meant with external control?

We need some way of detecting when the backend is no longer connected to an input context, so that we can remove it from the `ISurface` update loop and stop receiving events if we're doing nothing with them. Note that I never said "external control", but referred to usage of "external resources". `ISurface`, here, is an "external resource" which the input backend will be using because of what we do in the `Activated` callback.

HurricanKai: Hate to nitpick, but you did say external control, but I understand what you are saying now I think.
"external resources" are the typical case for `IDispose`, so what is the advantage of having `Activate` instead of a constructor and `Decativate` instead of `Dispose`?

ThomasMiz: > The whole idea of this proposal is that you can have multiple backends on an input context, but just one context for an input backend [object].
If the current backend in use were a property, the user could set it to change it and since it's a property the ```InputContext``` can react to that change, plus it would make it implicitly clear you can only use one backend at a time. Currently you can add multiple backends, but how do you choose which one is in use? How would input devices be handled? Do these lists change completely when the backend is changed? If so, it completely defeats the purpose of this refactor.

Perksey: > Currently you can add multiple backends, but how do you choose which one is in use? 

All of them are in use, that's the idea. `InputContext.Keyboards` would be all of the `IInputBackend.Keyboards` of the configured backends concatenated together.

> Hate to nitpick, but you did say external control,

Damn well okay that's my bad, I really shouldn't have!

> "external resources" are the typical case for IDispose

Because the lifetimes of the resources in question are not in sync, we're not "cleaning up" after ourselves we're just detaching ourselves of responsibility with particular resources.

Moreover, the input backend actually has no idea what resources it might be bonded do, remember the whole `ISurface` tie is done by a dedicated Windowing-Input integration. The whole point of `Activated` and `Deactivated` (the events, not the methods) is so that _user code_ (or something that represents/helps user code like `surface.CreateInput`) can detect when a backend is in action, so that they can start telling the backend to do stuff (i.e. `Update` its input data)

In something that isn't the Windowing-Input integration, the user for example could listen to `Activated` and, when activated, start a thread for updating the backend. We don't want that thread to be active needlessly, so `Deactivated` tells it to shutdown. The cornerstone of this proposal is that you can morph an input context to contain multiple backends as and when you see fit.

<details>
<summary>A further explainer I wrote before you said you understand what I'm saying, but figured I'd keep it in anyway</summary>

Further explainer: `Activate` is `InputContext`'s way of saying "alright backend, boss is coming look busy" and telling the `IInputBackend` to get its act together and start doing meaningful things (like updating input data). `Activate` (and by extension `Activated`) is a mechanism for telling the backend that we care about it now, which is why it's only at this point that we bond the context to the `ISurface`. `ISurface` will, in all likelihood, outlive the input context (even if by just a couple of nanoseconds in most cases) so we need to be able to detect when `InputContext` no longer cares about a particular backend.

</details>

ThomasMiz: > All of them are in use, that's the idea. ```InputContext.Keyboards``` would be all of the ```IInputBackend.Keyboards``` of the configured backends concatenated together.

This makes no sense to me. What's the point of being able to add/remove backends at will if you'd also need to change all the input device instances you're using? How would you even know which IKeyboard instance to use when you have multiple backends in one context?

Perksey: Multiple keyboards was probably a really poor example here, but the main motivator for this proposal is you'll have different backends for different kinds of input devices. For instance, I plan on adding (at some point in 3.X) a wrapper around OpenXR's input mechanisms. This means that we'd have one input backend to receive traditional keyboard and mice inputs, and then one input backend for VR hands, both of which will be accessible from one object (i.e. in the future we'll have `InputContext.Hands` as well as `InputContext.Keyboards`)

> change all the input device instances you're using?

Given this proposal knows about the original Silk.NET 1.0 input proposal, the same requirement applies:

```
In implementations, once an IJoystick, IGamepad, or other IInputDevice object has been created, it must not be destroyed until its IInputContext is disposed. This is to preserve event bindings.
```

Of course, here `IInputContext` is substituted with `IInputBackend`. 

If the user chooses to remove a backend they've bound events to, that's on them. It'll deactivate, which means all of their event bindings are useless but they knowingly removed the backend. Silk.NET itself likely won't ever remove a backend, only add backends, as I don't see any reason why we'd need to yank one out.

Perksey: @HurricanKai After giving it some thought, I appreciate the merit in using disposables. I'll change the proposal to use `IDisposable`, scrap the whole passing around thing, scrap `Activate` and `Activated` and assume that you'll just need to do any preinit before you add the backend, and add a `Disposing` event. Not the cleanest, but not the dirtiest either.

ThomasMiz: Wait, so one input context can have multiple backends active at the same time? Now this makes more sense.

I'll still bring this up again:

> An input context is set as active/in use by a context using the `Activate` and `Deactivate` methods.

I think you meant to write "An input backend is set..."

Perksey: Yes, that was meant to be backend not context. Sorry.

However, any reference to `Activate` and `Deactivate` has now been removed. Please read the revised lifetime rules introduced in the last two commits.

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -41,14 +41,12 @@ namespace Silk.NET.Input
...
 
 The API surface for this is defined later in the Proposed API section.
 
-One instance of an `IInputBackend` can only belong to one input context at a time. If multiple input contexts are in use, multiple backend instances should also be used. An input context is set as active/in use by a context using the `Activate` and `Deactivate` methods. `Activate` should throw an exception if already activated, and `Deactivate` should throw an exception if not already activated. Once deactivated, an input backend can be reactivated.
+One instance of an `IInputBackend` can only belong to one `InputContext` throughout the entirety of the `IInputBackend`'s lifetime. Once a `IInputBackend` is `Add`ed to an `InputContext`, the `InputContext` takes full ownership of the `IInputBackend` and, as a result, will `Dispose` the `IInputBackend` when it is `Remove`d or the context itself is `Dispose`d. If multiple `InputContext`s are in use, multiple `IInputBackend` instances must also be used. `IInputBackend` should throw an exception if its APIs are used after the backend has been disposed.
```
HurricanKai, line 49:
```suggestion
One instance of an `IInputBackend` can only belong to one `InputContext` throughout the entirety of the `IInputBackend`'s lifetime. Once a `IInputBackend` is `Add`ed to an `InputContext`, the `InputContext` takes full ownership of the `IInputBackend` and, as a result, will `Dispose` the `IInputBackend` when it is `Remove`d or the context itself is `Dispose`d. If multiple `InputContext`s are in use, multiple `IInputBackend` instances must also be used. `IInputBackend` will throw an exception if its APIs are used after the backend has been disposed.
```

HurricanKai (commented):
Looking pretty good now. I'm not fully convinced it's good to have Add as a normal method still, I think the InputContext also calling the constructor might be easier to use.

Perksey:
> I think the InputContext also calling the constructor might be easier to use.

Unfortunately wouldn't be possible as we need to capture the API object and other native handles when creating the backends (won't just be parameterless so we can't use the `T : new()` constraint)

Perksey:
@HurricanKai @ThomasMiz are you happy with this? Please approve on GitHub if you are.

HurricanKai (approved):

ThomasMiz (approved):

thargy:
Hi @Perksey & @HurricanKai,

The more I look into this the more my spidey-sense is tingling.  The source generator requirement breaks KISS and produces non-intuitive behaviour for a consumer of the libraries.  Further, it produces a default behaviour for the backends we know about, that isn't reflected for those we don't know about (which is bad design).  Every time a solution like this has appeared in the past I've eventually realised there was a much simpler approach.

I get we don't want a dependence from Input -> Backend, for example, when using VR we won't want the Window backend, etc.

The case for why we don't want backends dependent on input is far less clear though.  I suspect the main justification is that you might wish to use a separate input system.  This is countered by 2 approaches, have the backend implement functionality for input by implementing a thin set of functionality atop of the base classes and interfaces in input, but ensure that initialisation/etc. is optional (classic approach); or have a third package that is dependent on both the Backend and Input (this is overkill).

Perksey:
One way we have countered this in the past is having an interchange type in Silk.NET.Core. For Windowing, this was `INativeWindow` - an interface that just contains the handles we need and nothing more. We're trying to avoid using Silk.NET.Core as much as possible, as in 2.X this became a massive dumping ground for anything that makes interoperability with any of our packages better.

In Silk.NET 1.X, we did have a package which literally only existed to bond Windowing and Input, but I don't think that was a good solution. 2.X is structured such that each backend has its own package. We could do this again (i.e. `Silk.NET.Input.Windowing` and `Silk.NET.Input.Common`) and reintroduce the `Silk.NET.Input` "metapackage", but this deviates from the Windowing redesign which mandates that one "reference implementation" is bundled with the main package. Personally if we do something for Input in terms of package structure I'd like it to be reflected in Windowing as well for consistency.

HurricanKai:
I'm not a big fan of using source generators for this either, but it's a bit unclear to me what other options there are? We don't want to use reflection, and we don't want to have the user deal with what backend they use.

HurricanKai:
I really like the idea of the device just holding the state, the backend updating that state as needed and then the user just querying that state.
Not sure how I feel about consolidating Window and Input into one package, I mean given we already consolidate extensions and core packages into one package in 3.0, and therefore already rely on .NET 6 linking a lot it could be worthwhile and would simplify stuff a lot.
This idea would also tie in very well into the new `ISurface` APIs, so that is a plus point.

Perksey:
Windowing and Input I want to keep separate. I will write up the rest of my opinions later.

thargy:
Worth noting that, at the moment, the design only uses Context as a convenient Backend grouper.  You could drop Context entirely and have the `FindDevice`/`GetState` methods implemented as extension methods on a single and/or multiple backends if you're not to bothered about grouping the events.  This makes it even easier to use in the common case that you just want a singular backend to supply you with devices.

Perksey:
> In terms of keeping input out of windowing, then the normal approach is the shim NuGet approach. This is what logging libraries usually do (amongst many others). It does create NuGet bloat, but it's not a massive deal. That's still preferable to magic code injection (IMHO) which is non-intuitive to a consumer/contributor. Placing Window logic in Input is also worrisome as it's effectively backwards and makes Window input behave differently to a 3rd-party backend.

Agree, I think a `Silk.NET.Input.Abstractions` (_not_ using `.Common` to avoid confusion) and `Silk.NET.Input.Windowing` (maybe not that name?) split is as good as we can get without source generators, and is something I can deal with as much as I hate introducing packages.

<blockquote>
<details><summary>Abridged proposal containing only surface API</summary>

    // Prototype implementation
    // Note - thread-safety and performance are not implemented.
    namespace Silk.NET.Input
    {
        /// <summary>
        /// Represents an input backend from which input devices can be retrieved.
        /// </summary>
        /// <remarks>
        /// This interface is to be implemented by backend libraries, e.g. Windowing.
        /// </remarks>
        public interface IInputBackend
        {
            /// <summary>
            /// Collection of all devices, supplied by this backend.
            /// </summary>
            IReadOnlyList<InputDevice> Devices { get; }

            /// <summary>
            /// Raised whenever devices are added to the backend.
            /// </summary>
            event Action<IReadOnlyList<InputDevice>>? DevicesAdded;

            /// <summary>
            /// Raised whenever devices are removed from the backend.
            /// </summary>
            event Action<IReadOnlyList<InputDevice>>? DevicesRemoved;
        }

        /// <summary>
        /// Groups together all <see cref="IInputBackend">backends</see> into a single context.
        /// </summary>
        public class InputContext : IDisposable
        {
            public IReadOnlyList<IInputBackend> Backends { get; }

            /// <summary>
            /// Returns all devices from all backends.
            /// </summary>
            public IReadOnlyList<InputDevice> AllDevices { get; }

            /// <summary>
            /// Gets all devices of the given type.
            /// </summary>
            /// <remarks>
            /// The <typeparamref name="T" /> type parameter must be the last interface in the inheritance heirarchy (the "device type")
            /// i.e. it must be a device type like <see cref="IJoystick" /> and not something like <see cref="IInputDevice" />.
            /// The behaviour of using a T that isn't a device type recognised by the backend is undefined.<br />
            /// <br />
            /// This method is not intended for general consumption. Consider <see cref="InputContext" /> instead.
            /// </remarks>
            public IReadOnlyList<T> GetDevices<T>() where T : InputDevice;

            /// <summary>
            /// Find the first device that matches the optional filter criteria.
            /// </summary>
            /// <param name="isConnected">Whether the device must be connected.</param>
            /// <param name="backend">The backend.</param>
            /// <param name="deviceId">The device ID.</param>
            /// <param name="name">The device name.</param>
            /// <param name="comparisonType">The string comparison type to use for device names</param>
            /// <typeparam name="T">The device type.</typeparam>
            /// <returns></returns>
            public T? FindDevice<T>(
                bool? isConnected = null,
                IInputBackend? backend = null,
                int? deviceId = null,
                string? name = null,
                StringComparison comparisonType = StringComparison.Ordinal) where T : InputDevice;

            /// <summary>
            /// Gets the state of the first connected device that matches the optional filter criteria.
            /// </summary>
            /// <param name="backend">The backend.</param>
            /// <param name="deviceId">The device ID.</param>
            /// <param name="deviceType">The device type.</param>
            /// <param name="name">The device name.</param>
            /// <param name="comparisonType">The string comparison type to use for device names</param>
            /// <typeparam name="T">The device type.</typeparam>
            /// <returns></returns>
            public T? GetState<T>(
                IInputBackend? backend = null,
                int? deviceId = null,
                Type? deviceType = null,
                string? name = null,
                StringComparison comparisonType = StringComparison.Ordinal)
                where T : InputDeviceState;

            public InputContext();

            /// <summary>
            /// Initializes a new context with the provided backends.
            /// </summary>
            /// <param name="backends">The backends.</param>
            public InputContext(params IInputBackend[] backends);

            /// <summary>
            /// Initializes a new context with the provided backends.
            /// </summary>
            /// <param name="backends">The backends.</param>
            public InputContext(IEnumerable<IInputBackend> backends);

            /// <summary>
            /// Adds the <paramref name="backends"/> to this context.
            /// </summary>
            /// <param name="backends">The backends to add</param>
            /// <returns><see langword="true"/> if any backends were added; <see langword="false"/> otherwise.</returns>
            public bool Add(params IInputBackend[] backends) => Add((IEnumerable<IInputBackend>) backends);

            /// <summary>
            /// Adds the <paramref name="backends"/> to this context.
            /// </summary>
            /// <param name="backends">The backends to add</param>
            /// <returns><see langword="true"/> if any backends were added; <see langword="false"/> otherwise.</returns>
            public bool Add(IEnumerable<IInputBackend> backends);

            /// <summary>
            /// Remove the <paramref name="backends"/> from this context.
            /// </summary>
            /// <param name="backends">The backends to remove</param>
            /// <returns><see langword="true"/> if any backends were removed; <see langword="false"/> otherwise.</returns>
            public bool Remove(params IInputBackend[] backends);

            /// <summary>
            /// Remove the <paramref name="backends"/> from this context.
            /// </summary>
            /// <param name="backends">The backends to remove</param>
            /// <returns><see langword="true"/> if any backends were removed; <see langword="false"/> otherwise.</returns>
            public bool Remove(IEnumerable<IInputBackend> backends);

            /// <summary>
            /// Removes all backends from the context.
            /// </summary>
            /// <returns><see langword="true"/> if any backends were removed; <see langword="false"/> otherwise.</returns>
            public bool Clear();

            /// <summary>
            /// Attaches to backend events
            /// </summary>
            /// <param name="backends">Enumeration of backends</param>
            protected virtual void Attach(IEnumerable<IInputBackend> backends);

            /// <summary>
            /// Detaches from backend events
            /// </summary>
            /// <param name="backends">Enumeration of backends</param>
            protected virtual void Detach(IEnumerable<IInputBackend> backends);

            /// <summary>
            /// Called whenever devices are added.
            /// </summary>
            /// <param name="devices"></param>
            protected virtual void OnDevicesAdded(IReadOnlyList<InputDevice> devices);

            /// <summary>
            /// Called whenever devices are removed.
            /// </summary>
            /// <param name="devices"></param>
            protected virtual void OnDevicesRemoved(IReadOnlyList<InputDevice> devices);

            /// <summary>
            /// Called whenever a device's state is changed. 
            /// </summary>
            /// <param name="stateChanges"></param>
            protected virtual void OnStateChange(InputDevice device,
                IReadOnlyCollection<InputDeviceStateChange> stateChanges)
                => StateChanges?.Invoke(device, stateChanges);

            /// <summary>
            /// Raised whenever devices are added to the context.
            /// </summary>
            public event Action<IReadOnlyList<InputDevice>>? DevicesAdded;

            /// <summary>
            /// Raised whenever devices are removed from the context.
            /// </summary>
            public event Action<IReadOnlyList<InputDevice>>? DevicesRemoved;

            /// <summary>
            /// Fired whenever any state changes occur.
            /// </summary>
            private event Action<InputDevice, IReadOnlyCollection<InputDeviceStateChange>>? StateChanges;

            /// <summary>
            /// Disposes this context, detaching from all backends.
            /// </summary>
            public void Dispose();
        }

        /// <summary>
        /// Base class for all input devices, is provided by an <see cref="IInputBackend"/> and holds a current state.
        /// </summary>
        /// <remarks>
        /// <para>The device's current state is represented as a dictionary of state keys with long values.</para>
        /// </remarks>
        public abstract class InputDevice
        {
            /// <summary>
            /// Creates a new input device
            /// </summary>
            /// <param name="backend">The backend.</param>
            /// <param name="deviceId">The device ID.</param>
            /// <param name="name">An optional friendly name.</param>
            /// <param name="initialState">The initial state.</param>
            public InputDevice(IInputBackend backend, int deviceId, string? name = null,
                IEnumerable<KeyValuePair<string, long>>? initialState = null);

            /// <summary>
            /// Whether the device is currently connected
            /// </summary>
            /// <remarks>
            /// You should not hold on to a disconnected device.</remarks>
            public bool IsConnected { get; protected set; }

            /// <summary>
            /// The backend that supplies this device
            /// </summary>
            public IInputBackend Backend { get; init; }

            /// <summary>
            /// The backend-specific, device-type-specific identifier for this device.
            /// </summary>
            /// <remarks>
            /// For example, no gamepad may share a DeviceId with another gamepad (same applies for all other device types) on that individual backend.
            /// This property should not be used as a globally unique identifier.
            /// 1 2 3 4 5
            /// 1 3 
            /// 5
            /// </remarks>
            public int DeviceId { get; init; }

            /// <summary>
            /// An optional device friendly name (e.g. "Mouse", "Keyboard", does not need to be unique), this often comes from the USB identifier string.
            /// </summary>
            public string Name { get; init; }

            /// <summary>
            /// Fired whenever any state changes occur.
            /// </summary>
            public event Action<InputDevice, IReadOnlyCollection<InputDeviceStateChange>>? StateChanges;

            /// <summary>
            /// Updates the current state.
            /// </summary>
            /// <param name="state">The new state keys.</param>
            /// <returns><see langword="true"/> if any state change was detected; <see langword="false"/> otherwise.</returns>
            protected IReadOnlyList<InputDeviceStateChange> SetState(IEnumerable<KeyValuePair<string, long>> state);

            protected virtual void OnStateChanges(IReadOnlyList<InputDeviceStateChange> changes);

            /// <summary>
            /// Gets a snapshot of the current device state.
            /// </summary>
            public InputDeviceState State => new InputDeviceState(this, _state);
        }

        /// <summary>
        /// A snapshot of an input device's current state
        /// </summary>
        public class InputDeviceState : IReadOnlyDictionary<string, long>
        {
            public InputDevice Device { get; init; }

            protected internal InputDeviceState(InputDevice device, IReadOnlyDictionary<string, long> state);

            /// <summary>
            /// Whether the device was connected when the state was snapshot. 
            /// </summary>
            public bool WasConnected => _state.TryGetValue(StateKey_Connected, out var value) && value != 0L;
        }

        /// <summary>
        /// A simple generic input device.
        /// </summary>
        /// <remarks>
        /// This is best used for testing purposes, in practice a backend should implement a concrete type that provides
        /// functionality to interpret it's state more readily.
        /// </remarks>
        public class GenericInputDevice : InputDevice
        {
            public GenericInputDevice(IInputBackend backend, int deviceId, string? name = null,
                IEnumerable<KeyValuePair<string, long>>? initialState = null)
                : base(backend, deviceId, name, initialState);

            protected IReadOnlyList<InputDeviceStateChange> Update(IEnumerable<KeyValuePair<string, long>> state)
                => SetState(state);
        }

        /// <summary>
        /// Interface for state changes on a device, lists state keys that have changed.
        /// </summary>
        public readonly struct InputDeviceStateChange
        {
            /// <summary>
            /// The state key that changed.
            /// </summary>
            public readonly string Key;

            /// <summary>
            /// The old state value (if any); otherwise <see langword = "null"/> indicates the state was added.
            /// </summary>
            public readonly long? OldValue;

            /// <summary>
            /// The new state value (if any); otherwise <see langword = "null"/> indicates the state was removed.
            /// </summary>
            public readonly long? NewValue;

            public InputDeviceStateChange(string key, long? oldValue, long? newValue);
        }

        /***
         * OPTIONAL: Input can provide some common 'strongly-typed' overloads for convenience. i.e. mouse, keyboard,
         * joystick, etc.  
         *
         * These would still be implemented by the backend, but would be easier to implement
         *
         * This example shows a generic mouse implementation, with state keys for X,Y and buttons.
         *
         * A keyboard would have states for modifiers, and would add/remove state keys for each key pressed. (The state
         * system tracks addition/removal of keys too, so keys can be 'permanent' or 'transient').
         *
         * Note that transitions to/from states are supported so it is possible to see the current state as well
         * as use events to detect transitions.  Handling key auto-repeats however is more complex, but can still be done
         * by the device implementation for a generic keyboard by triggering an additional event every x-ms whilst a
         * key is held down. 
         ***/
        
        /// <summary>
        /// Generic mouse device implementation
        /// </summary>
        public abstract class MouseDevice : InputDevice
        {
            protected MouseDevice(
                IInputBackend backend,
                int deviceId,
                string? name = null,
                IEnumerable<KeyValuePair<string, long>>? initialState = null)
                : base(backend, deviceId, name ?? "Generic mouse", initialState);

            protected override void OnStateChanges(IReadOnlyList<InputDeviceStateChange> changes);

            /// <summary>
            /// Get's the latest mouse state.
            /// </summary>
            public new MouseState State => new MouseState(this, base.State);

            public event Action<MouseMoveEvent> MouseMove;
            public event Action<MouseButtonEvent> ButtonsChanged;
        }

        public class MouseState : InputDeviceState
        {
            public MouseDevice Device { get; init; }

            // State encapsulation field
            private IReadOnlyDictionary<string, long> _state;

            public MouseState(MouseDevice device, IReadOnlyDictionary<string, long> state)
                : base(device, state)
            {
            }

            #region Mouse State helpers

            public long X => TryGetValue(StateKey_X, out var value) ? value : 0L;

            public long Y => TryGetValue(StateKey_Y, out var value) ? value : 0L;

            public bool LeftClicked => TryGetValue(StateKey_Left_Click, out var value) && value != 0L;

            public bool RightClicked => TryGetValue(StateKey_Right_Click, out var value) && value != 0L;

            public bool MiddleClicked => TryGetValue(StateKey_Middle_Click, out var value) && value != 0L;

            public bool ButtonClicked(int number) =>
                TryGetValue($"{StateKey_Button_Prefix} {number}", out var value) && value != 0L;

            #endregion
        }

        public readonly struct MouseMoveEvent
        {
            public readonly MouseDevice Device;
            public readonly Vector2 Position;
            public readonly Vector2 Delta;

            public MouseMoveEvent(MouseDevice device, InputDeviceStateChange? xChange, InputDeviceStateChange? yChange)
            {
                Device = device;
                Position = new Vector2(xChange?.NewValue ?? 0L, yChange?.NewValue ?? 0L);
                Delta = new Vector2(Position.X - (xChange?.OldValue ?? 0L),
                    Position.Y - (yChange?.OldValue ?? 0L));
            }
        }

        public readonly struct MouseButtonEvent
        {
            public readonly MouseDevice Device;

            /// <summary>
            /// Button indices that changed. <see langword="true"/> if button was pressed; <see langword="false"/>
            /// otherwise.
            /// </summary>
            private readonly IReadOnlyDictionary<int, bool> _buttonChanges;

            public MouseButtonEvent(MouseDevice device, IEnumerable<InputDeviceStateChange> buttonChanges)
            {
                Device = device;
                var changes = new Dictionary<int, bool>();
                foreach (var change in buttonChanges)
                {
                    // Get the button number from the key
                    if (!int.TryParse(change.Key[MouseState.StateKey_Button_Prefix.Length..], out var index))
                        continue;

                    var oldValue = change.OldValue ?? 0L;
                    var newValue = change.NewValue ?? 0L;

                    // Sanity check
                    if (oldValue == newValue) continue;

                    changes[index] = newValue != 0L;
                }

                _buttonChanges = changes;
            }

            public bool LeftClicked => _buttonChanges.TryGetValue(MouseState.Button_Left, out var value) && value;

            public bool RightClicked => _buttonChanges.TryGetValue(MouseState.Button_Right, out var value) && value;

            public bool MiddleClicked => _buttonChanges.TryGetValue(MouseState.Button_Middle, out var value) && value;
            public bool LeftReleased => _buttonChanges.TryGetValue(MouseState.Button_Left, out var value) && !value;

            public bool RightReleased => _buttonChanges.TryGetValue(MouseState.Button_Right, out var value) && !value;

            public bool MiddleReleased => _buttonChanges.TryGetValue(MouseState.Button_Middle, out var value) && !value;
            
            public bool ButtonClicked(int number) => _buttonChanges.TryGetValue(number, out var value) && value;
            public bool ButtonReleased(int number) => _buttonChanges.TryGetValue(number, out var value) && !value;
        }
        
        /*
         * Example backend implementation, this is dumb and just creates a single mouse device.
         * 
         */
        public class Window : IInputBackend
        {
            private readonly List<InputDevice> _devices;

            public Window()
            {
                _devices = new List<InputDevice> {new WindowMouseDevice(this, 0)};
            }

            public IReadOnlyList<InputDevice> Devices => _devices;
            public event Action<IReadOnlyList<InputDevice>>? DevicesAdded;
            public event Action<IReadOnlyList<InputDevice>>? DevicesRemoved;
        }
    }
}
</details>
</blockquote>

Most of my irks were with the implementation rather than the API, so I've ignored that in the name of the proposal.

- I agree with the philosophy of the proposal, as well as the surface API of the state, device, and context classes, except:
    - Why all the dictionary stuff? We don't need this, we can just add more fields to the structs - no need to abstract away the state storage. This is thanks to (from the SDP):
    > Users are expected to keep all of the versions of all Silk.NET packages they are using in-sync.
    - Where are MouseDown and friends? Can we replace the generic event on InputContext with those specialized events? Also I'm not sure about the whole generic event thing, maybe something like `bool InputDevice.Subscribe<TState>(Action<TState> subscriber);` and a matching `Unsubscribe` method would make sense so we can receive `MouseButtonEvent` (from the Enhanced Input Events proposal) directly from the backend.
    - I don't think InputContext should be abstract, in fact my gut feel is that it should be sealed.
    - But yeah TLDR I love this, just the "state" and "events" are a bit too abstracted away 
- Naming is top notch as it definitely conveys that state represents the mouse at any point in time

thargy:
> Most of my irks were with the implementation rather than the API, so I've ignored that in the name of the proposal.

Yeah, this was thrown together to make the API compile and demonstrate the concepts a little more readily.  It's not production ready.

>   * Why all the dictionary stuff? We don't need this, we can just add more fields to the structs - no need to abstract away the state storage. This is thanks to (from the SDP):

By having state represented as a simple Dictionary it is using a simple property-bag model - this is something I've come across repeatedly and it has a number of advantages over using unknowable extensible structs.  This allows the code in the abstract device base class to handle state change logic (effectively diff'ing), event propagation and thread-synchronisation logic (not shown in the example), for all backends without backends needing to re-invent the wheel.  Further, diffing is faster (you can do it with unknown structs using reflection, but that has a start-up performance at least and requires a lot of code to create compilable lambdas).  It allows that without the input engine having any idea what a backend actually does.

>   * Where are MouseDown and friends?

`MouseDevice.ButtonsChanged` is fired whenever the mouse button states changes.  The `MouseButtonEvent` contains all the buttons that changed state.  Unlike the other API, this API is heavily focussed on batching events for performance, whenever possible.  As multiple buttons can technically change state at the same time (due to the way USB, or a polling backend may work, and due to a single `SetState` call) then we only want to throw one event when the state changes.  For example, if someone clicks both Left and Right mouse buttons, and releases Button 5 (on a multi-button mouse) then `MouseButtonEvent._buttonChanges` (which should probably be exposed through a public read-only property) contains ([1,true],[2,true],[5,false]), as such calling  `LeftClicked`, `RightClicked` and `ButtonReleased(5)` will all return `true`.  The properties can be renamed `LeftDown`, `LeftUp`, etc. if you prefer.  Importantly, there is no limit to how many buttons the mouse can expose, but we name the most common one for convenience (under the hood they are numbered though).


> Can we replace the generic event on InputContext with those specialized events?

The generic event is called for all state changes because we know about all state changes at the base.  This is convenient for logging and debugging as it allows you to see changes being raised by devices you're learning about, also useful during backend development, as you can quickly throw something together before adding all the extra event and state wrappers.  However, descendent Devices can (and should) add more specialised events that are more relevant, at no additional performance cost (I won't go into it now, but you can reduce the performance impact by only doing stuff when events are listened to, I didn't implement that for clarity).  The `MouseDevice` demonstrates this with its two additional example events:
```csharp
        public event Action<MouseMoveEvent> MouseMove;
        public event Action<MouseButtonEvent> ButtonsChanged;
```

This approach gives you the best of both worlds IMHO.

> Also I'm not sure about the whole generic event thing, maybe something like `bool InputDevice.Subscribe<TState>(Action<TState> subscriber);` and a matching `Unsubscribe` method would make sense so we can receive `MouseButtonEvent` (from the Enhanced Input Events proposal) directly from the backend.

Not quite sure where you're going with that one (look a bit RX like).  Under this proposal the backend only needs to override one of the standard abstract devices (like MouseDevice, KeyboardDevice) and implement calls to `SetState` and be done, removing the majority of boilerplate.  Where a backend wants to introduce an entirely new kind of device, then they have the ability to do that, but it won't be that common, as the end consumer would need to also understand that new device type to get the most benefit.

Another reason for having the base `StateChanges` event is that it allows a backend to expose other state changes beyond the 'norm'.  For example, let's say we have a special gaming keyboard that allows you to set its backlight RGB.  We might create a backend that knows about it and extends the `KeyboardDevice` from the `Input` Nuget (not shown yet!) with a `BacklitKeyboardDevice`, which has the additional members:

```csharp
	public class BacklitKeyboardDevice : KeyboardDevice
	{
		...
        public event Action<BackLightEvent> BacklightChanged;
		public void SetBacklight(Vector3 color) { ... }
	}


    public class BacklitKeyboardState : KeyboardDeviceState
    {
        #region State Keys used by Device

        public const string StateKey_BacklightColor = "BacklightRGB";

        #endregion

        public BacklitKeyboardDevice Device { get; init; }

...

        #region Mouse State helpers

        public Vector3 Color => TryGetValue(StateKey_BacklightColor, out var value) ? new Vector3(/* Convert long */...) : Vector3.Zero;

        #endregion
    }

	public readonly struct BackLightEvent
	{
            public readonly BacklitKeyboardDevice Device;
            public readonly Vector3 OldColor;
            public readonly Vector2 NewColor;

			...
	}
```

The beauty of this is that any code written to expect just a KeyboardDevice will continue to work as normal.  And if you have Debug code writing out events from the keyboard you will see state changes referring to `StateKey_BacklightColor` whenever the colour changes (or you grab a state snapsot).  However, you can also then add code to make use of the actual `BacklitKeyboardDevice` which not only now exposes a specific event for backlight changes, but gives you a method to call to actually change the backlight.  Note that calling the method would send a message to the keyboard to make the change, and then, at a later time, the state would change to reflect the new color (easy asynchrony).  This was something we discussed in the working group - with a `MouseDevice` (being able to 'set' the cursor position), though I didn't show that in the example implementation for brevity.  (In general, methods that can change a device should always be method calls, and we should resist them as this is an 'input' library).


>   * I don't think InputContext should be abstract, in fact my gut feel is that it should be sealed.

Yeah, that was a toss-up, I made it abstract as it's somewhat limited usefulness at the moment, it merely groups together backends.  However, being a backend grouper then it's also hard to see the use case for a backend extending it...

Perksey:
> By having state represented as a simple Dictionary it is using a simple property-bag model - this is something I've come across repeatedly and it has a number of advantages over using unknowable extensible structs.

This is just a struct wrapper around a dictionary though, which I don't think is needed. Because it's a struct, it isn't really any more extensible than if they were regular fields (and I'm not saying make it more extensible, I'm saying we don't need this).

Dictionaries will lead user code to have to do roundabout hash table lookup in use, something which is an undesirable expense given this will likely be used on a hot path. The best summary of my train of thought is that I just don't want a bag, I want pigeon holes where everything has its place fixed. The dictionaries make me uncomfortable.

For diffing I think we should do this manually or if this proves to be too much of a small make a private (i.e. not seen by the end user) source generator which does this for us. If we really wanted to get performance nuts about this, more theoretical performance is possible by doing it this way if we wanted to use intrinsics in the diffs (but I doubt we'd need to)

> Unlike the other API, this API is heavily focussed on batching events for performance, whenever possible.

This will decrease the number of `callvirt`s but I think the performance difference overall will be immeasurable or worse. If we needed to batch events, which I don't think is a bad idea, we can investigate altering Enhanced Input Events such that the actual button data is a `Span<T>` where T is button-specific data, the event structures `ref struct`s, and the delegates named to get around the `ref struct` generics restrictions.

> MouseDevice.ButtonsChanged is fired whenever the mouse button states changes.

Unified mouse events were rejected in the initial drafts of Enhanced Input Events, so I'd rather not do that. It's going out of our way to remove all familiarity for little benefit, and it makes it awkward to use. I'd imagine most of our users don't even care about changing the state, they just want to know when things happen and having dedicated events with the event structs proposed in Enhanced Input Events exposes this in a clean way.

> Yeah, that was a toss-up, I made it abstract as it's somewhat limited usefulness at the moment, it merely groups together backends. However, being a backend grouper then it's also hard to see the use case for a backend extending it...

Eeeeeeeehhhhhhhhhhhhhh come back to me.

> Not quite sure where you're going with that one (look a bit RX like)

The idea is that the Input Context subscribes to a specific kind of state change individually, and routes this to the events it exposes.

Perksey:
I think the main thing is the new events are too fluid. I would rather have rigid events with a fluid way of subscribing to them rather than fluid events with a rigid way of subscribing to them. This would what my proposed `Subscribe` and `Unsubscribe` would effectively do. The state should be rigidly structured and rigidly defined.

thargy:
I've gone down the rabbit hole before of building type differs. In fact, I've gone back and forth between dictionaries and type members (using reflection or T4 source gen) so many times, I've lost count (check out the horrifying decade long history on the [core libraries](https://github.com/webappsuk/CoreLibraries/tree/master/Utilities), which I wrote most of between 2007-2018, for just a few examples of the back and forth).  

It has therefore been my bitter experience that, as the adage goes, performance is an engineering problem, not a design one.  That is you shouldn't optimise it unless you prove you have too. I have used dictionaries on hot paths of very large scale applications, and they are fast (being hashtable lookups).  But I accept it involves heap allocations, etc.  Either way, using source gen or reflection both require a lot more work, so you really should establish the problem exists before exploring that path.  At least that's been my experience.  

In a similar way, event batching just makes sense with a batch state setter, and though event invocation has improved over the years (compared to the bad ol' days) it's still best minimised.  Ultimately, events are multicast delegates and are therefore slower than simple method invocations.

In terms of rejecting unified mouse events, I think it's more than that.  A mouse, like a keyboard, is capable of reporting multiple buttons being pressed/released at the same time, having them grouped makes it much easier to code for detecting, for example, left & right mouse buttons pressed together; just as you would do with multiple keys (which I know was a criticism levelled against the old API).

However, I'm not precious, I wouldn't use the event interface in practice anyway, preferring to grab the state on each game loop, but it's there for those who prevent event-driven programming.  But, the more events exposed means the more checks and invocations required every time a state changes, I see that as a bigger overhead than dictionary lookups.

I'm really not sure what you're getting at with the Subscribe/Unsubscribe proposal.  It looks, exactly like a slower re-implementation of multicast delegates/events by another name?  As for adding concrete events (like MouseMove) to the Context, that blocks extensibility and future-proofing as you can't add additional ones in future when implementing a new Device (like the BaclitKeyboard example I gave).

I'd try to implement what you're suggesting, but I don't think I've followed you properly, could you work up a practical example?

Perksey:
> performance is an engineering problem, not a design one.

Right but I don't think we should allow ourselves to potentially be bottlenecked by a design which, if approved now, we'd be locked into until 2023 when Silk.NET 4.0 releases. By the time we've proven that this is a problem, there's nothing we can do about it because the dictionary stuff is part of the public API. At least if it were an implementation detail we can choose not to use it. Moreover it's our job to think of performance as much as possible, as Silk.NET can never afford to be even remotely a bottleneck.

> they are fast (being hashtable lookups)

Not as fast as regular field retrievals.

> Ultimately, events are multicast delegates and are therefore slower than simple method invocations.

They're not that much slower on a hot path. Most modern CPUs will cache the VTable resolved function's instructions if called lots (within reason), and this is why our indirect calls don't prove to be that slow in practice either. 

> I'm really not sure what you're getting at with the Subscribe/Unsubscribe proposal. It looks, exactly like a slower re-implementation of multicast delegates/events by another name?

Yes it's literally just to get around that we can't do
```cs
public event Action<T> StateChanged<T>;
```

I agree it's not great, but IInputBackend was never intended for general consumption - InputContext remains the main user-facing API.

> I'd try to implement what you're suggesting, but I don't think I've followed you properly, could you work up a practical example?

Will have a play this afternoon.

> As for adding concrete events (like MouseMove) to the Context, that blocks extensibility and future-proofing as you can't add additional ones in future when implementing a new Device (like the BaclitKeyboard example I gave).

Why would this be the case? We want to be able to change the Context class however frequently we want without affecting the backend.

thargy:
> > performance is an engineering problem, not a design one.
> 
> Right but I don't think we should allow ourselves to potentially be bottlenecked by a design which, if approved now, we'd be locked into until 2023 when Silk.NET 4.0 releases. By the time we've proven that this is a problem, there's nothing we can do about it because the dictionary stuff is part of the public API. At least if it were an implementation detail we can choose not to use it. Moreover it's our job to think of performance as much as possible, as Silk.NET can never afford to be even remotely a bottleneck.

Agreed, sorry I meant that, for performance-critical scenarios, sometimes we need to prototype options and benchmark prior to settling the design, not that we develop and release a less performant version and wait for the next major milestone.  You see this approach a lot in the dotnet team's own approach when debating designs of performance-critical code (I've had to submit BenchmarkDotNet samples, to show relative performance in terms of speed and allocs).

> > they are fast (being hashtable lookups)
> 
> Not as fast as regular field retrievals.

Yes, obviously not! However, the real-world impact still comes down to how often you do those ~5ns dictionary lookups each game loop compared to everything else you do, and what you gain/lose in return.

> > Ultimately, events are multicast delegates and are therefore slower than simple method invocations.
> 
> They're not that much slower on a hot path. Most modern CPUs will cache the VTable resolved function's instructions if called lots (within reason), and this is why our indirect calls don't prove to be that slow in practice either.

Agreed, this is the problem with performance during design is that these slower/not slower things are hard to predict the impact of without an actual real-world benchmark. I only said they were slower, not by how much.  Indeed, even after years of writing performant code, I get surprised when something I thought would be slower is actually faster, etc.  That's particularly true as the compiler team continue to optimise (cf the new record struct performance compared to plain old structs, they add cool functionality and are faster).  For example, we used to have to avoid `Enum.HasFlag` like the plague, until the compiler team made it as fast (in some cases faster) than hand-coding the logic, and then we had to go back and rewrite.

> > I'm really not sure what you're getting at with the Subscribe/Unsubscribe proposal. It looks, exactly like a slower re-implementation of multicast delegates/events by another name?
> 
> Yes it's literally just to get around that we can't do
> 
> ```cs
> public event Action<T> StateChanged<T>;
> ```
> 
> I agree it's not great, but IInputBackend was never intended for general consumption - InputContext remains the main user-facing API.
> 
> > I'd try to implement what you're suggesting, but I don't think I've followed you properly, could you work up a practical example?
> 
> Will have a play this afternoon.
> 
> > As for adding concrete events (like MouseMove) to the Context, that blocks extensibility and future-proofing as you can't add additional ones in future when implementing a new Device (like the BaclitKeyboard example I gave).
> 
> Why would this be the case? We want to be able to change the Context class however frequently we want without affecting the backend.

Ah! :bulb:

So in my proposal the separation of concerns is this:

| Class | Role | Location |
| --- | :-- | --: |
| `IInputBackend` | Requires a backend to provide a collection of devices and notify changes | Input |
| `InputContext` | Groups backends for accessing all backends and devices together, grouping device attach/detach events, and all device events, if desired, which is useful for logging. | Input |
| `InputDevice` | Base class of all devices, hold device state in a generic way, detects changes and raises an event indicating what has changed and how | Input |
| `InputDeviceState` | Base class of all device state snapshots giving 'friendly' access to state | Input |
| `[Mouse\|Keyboard\|Gamepad\|Joystick\|...]Device` | Extends `InputDevice` and provides a common surface area for an abstract version of a device type, but _doesn't actually implement the device fully_. | Input |
| `[Mouse\|Keyboard\|Gamepad\|Joystick\|...]State` | Extends `InputDeviceState` and provides relevant state accessors for that abstract device type | Input |
| `[Mouse\|Keyboard\|Gamepad\|Joystick\|...]Event` | Creates event implementations for that abstract device type | Input |
| `*Device` | Concrete device implementation that can extend `InputDevice` directly if a unique device type, or one of the provided `[Mouse\|Keyboard\|Gamepad\|Joystick\|...]Device`. It's main purpose is to call `SetState` on the base classes whenever a device update occurs, and/or implement any abstract base methods that are for setting device state (rare as discussed above).  Where no new functionality is added these may be internal, leaving only the abstract common base class exposed (e.g. `MouseDevice`) | Backend |
| `*State` | Optional concrete device state implementation extends `InputDeviceState` and provides relevant state accessors for that device type (normally only needed for a unique device type) | Backend |
| `*Event` | Optional concrete event implementations for that device type  (normally only needed for a unique device type) | Backend |
| `*Backend` | Concrete backend implementation provides it's specific devices | Backend |

In the proposal:
* A 'device' is a provider of device-specific functionality (a mouse, a keyboard, etc.), only a mouse device would expose a `MouseMove` event, which makes no sense for a keyboard, etc.  The full implementation belongs in a backend (which may be 3rd-party and unknown to us).  `MouseDevice` (et al) are partial implementations to provide a common surface area so that any consumer of our library can make use of mouse-like devices without necessarily caring or worrying about the device's 'full' functionality as provided by the backend.  The `BaclkitKeyboardDevice` example illustrates this, you can drop in the backend that supports it and existing code will continue to treat it like a `KeyboardDevice`, however, if you care about the extra functionality you can check for the `BaclkitKeyboardDevice` and use it explicitly.  **NB** _The input library itself has no concrete implementations of any device_.
* A 'backend' is a provider of supported devices (e.g. a 'Windowing' backend may supply mouse, keyboard, gamepad and joystick implementations), it provides an implementation of `IInputBackend`, and concrete implementations for the devices it supports.
* A 'context' groups backends together, allowing it to expose all devices from all backends in one place.  This makes finding devices easier.

To use the system you add `backends` to a `context` then query the `context` for the device you want.  You can ask for a `MouseDevice` and you will get any concrete implementation that extends a `MouseDevice`, or you can ask for a specific concrete implementation like `BacklitKeyboardDevice`, etc.  Once you have a device, you can access its unique events (like `MouseMove` on a `MouseDevice`).

When I say this design is future proof, consider the following example. A 3rd party consumer wants to make use of our library, but has a custom device they want to work with; that device is a gaming keyboard with a backlight.  They don't want to re-implement the wheel for gaming keyboards, and in fact, want their gaming keyboard to be useable by existing games with minimal effort.  As such, they implement a new backend (`IInputBackend`) that provides a new `BacklitKeyboardDevice` that extends `KeyboardDevice`.  They only need to add the extra functionality for controlling and monitoring the state of the backlight (see example above), which is trivial, and providing the state information for their keyboard (i.e. which keys/modifiers are pressed) via `SetState`.  In their own code, they request a `BacklitKeyboardDevice` and can make use of the existing and new functionality.

They don't have to worry about the state thread-safety, or change tracking and event propagation, the InputDevice base class has done all that work for them already.  Nor do they have to request we add something to `InputContext` to make it work.  For this reason, a concrete event like `MouseMove` or `BacklightChanged` can't be exposed directly on `InputContext` as to do so would require all supported devices to be known when we build `InputContext` (the opposite of future proof).

Your 'Subscribe'\'Unsubscribe' model, however, does allow you a way to get close whilst remaining future-proof.  If a device, instead of exposing events, calls a generic base method on `InputDevice` with a generic 'event payload', then `InputDevice` can raise those 'events' using the `Subscribe\Unsubscribe` methodology, these can then be subscribed to by the `InputContext` and exposed there.  However, that's exactly what `IObservable<T>` is for, along with Rx (Reactive Extensions).  It is what I use in my own [HID library](https://github.com/DevDecoder/HIDDevices) for that reason, but if you're going for hyper-performance (Rx has overhead, which is something I'm intimately familiar with), then implementing a Subsribe\Unsubscribe model will require dictionaries and lookups, etc.

Based on SOLID, the 'context' should not contain events like `MouseMove`, only a Mouse device should.  The `StateChanges` event I exposed there is SOLID as a context can know that a device has changed state (by subscribing to it) even if it doesn't understand the state change itself, it is generic (like your proposal) and abstract, and so is primarily only useful for logging/debugging.  It is low overhead being an event.  I don't think we should be worried about the 'inconvenience' of getting the kind of device you want and then getting its events.

However, if you are concerned about, for example, getting all keyboard events from multiple keyboards, then this is where my proposal is that we forget about 'context' and instead think more about device grouping.  My current proposal treats context as a device group of all devices provided by a group of backends, that concept can be extended to a group by type (all keyboards from all backends), etc.  There are complexities though, about how you group together state (effectively overwriting in specified priority order).  But that's a point for a broader discussion and breaks KISS.

Usually, a consumer wants to keep separate events from individual devices (e.g. gamepads) not group them all together.  As such, the proposed design of finding a connected device instance from a group of backends and then querying it about its state (or subscribing to its events) seems like the right approach.  Indeed I have no objection to dropping the `StateChanges` event from `InputContext` entirely if its objectionable, it's only there because it's essentially free and readily available (and useful for debugging/logging as mentioned).

PR #561, opened 2021-08-06
"Winapi" support on NS20 and .NET 5

Perksey:
2.7 is due today, to get this fix in we'll delay release to tomorrow.

This implements support for platform-default P/Invoke calling conventions instead of blindly using Cdecl (which is incorrect!) This is something that was pointed out when SilkTouch development first started, but was not implemented.

On .NET Standard 2.0, seeing as there's no true way to specify "platform default" we use a "best effort" implementation to cover known cases. Right now, this is just 32-bit Windows being not Cdecl - there are no other concerns we know of today.

Codegen:
- [Assimp.WinapiNET5.gen.txt](https://github.com/dotnet/Silk.NET/files/6948095/Assimp.WinapiNET5.gen.txt)
- [Assimp.WinapiNS20.gen.txt](https://github.com/dotnet/Silk.NET/files/6948096/Assimp.WinapiNS20.gen.txt)

### DIFF
--- a/src/Assimp/Silk.NET.Assimp/Assimp.cs
+++ b/src/Assimp/Silk.NET.Assimp/Assimp.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Runtime.InteropServices;
 using Silk.NET.Core.Contexts;
 using Silk.NET.Core.Loader;
 using Silk.NET.Core.Native;
@@ -7,6 +8,7 @@ using Silk.NET.Core.Native;
 
 namespace Silk.NET.Assimp
 {
+    [NativeApi(Convention = CallingConvention.Winapi)]
     public partial class Assimp
     {
         public static Assimp GetApi()
--- a/src/Core/Silk.NET.Core/Native/NativeApiAttribute.cs
+++ b/src/Core/Silk.NET.Core/Native/NativeApiAttribute.cs
@@ -45,7 +45,7 @@ namespace Silk.NET.Core.Native
 
         public static CallingConvention GetCallingConvention(NativeApiAttribute attr, NativeApiAttribute parent)
         {
-            return attr?._actualConvention ?? parent?._actualConvention ?? CallingConvention.Winapi;
+            return attr?._actualConvention ?? parent?._actualConvention ?? CallingConvention.Cdecl;
         }
     }
 }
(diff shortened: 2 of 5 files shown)


HurricanKai (changes requested):
SilkTouch is run once per target.
Instead of delegating this to a runtime check that checks the Silk.NET.Core framework we built against, just put the #IF NET5_0 .... #else ... #endif right in the generated code (well not actually, just check in SilkTouch what target we are building).
Also, instead of having Winapi be the new default, just set it to be the default only for Assimp, that way we only have to multi-target Assimp to NET5 and NS.

Perksey:
> Instead of delegating this to a runtime check that checks the Silk.NET.Core framework we built against, just put the #IF NET5_0 .... #else ... #endif right in the generated code.

This uglies the code, both generated and generator and has zero gain. Not reworking just for this.

> Also, instead of having Winapi be the new default, just set it to be the default only for Assimp, that way we only have to multi-target Assimp to NET5 and NS.

This is a deviation from DllImport behaviour and expected behaviour, but fine.

HurricanKai:
> This uglies the code, both generated and generator and has zero gain. Not reworking just for this.
No it doesn't. It means there is zero code diff on .NET 5, and an extra if check on other platforms.

Perksey:
> No it doesn't. It means there is zero code diff on .NET 5, and an extra if check on other platforms.

There already is zero generated code diff on .NET 5.

Perksey:
Unless you mean diff between the outputs of the two different targets, in which case I'm not reworking this for that as it's absolutely zero gain.

Perksey:
Is there anything left to do here?

HurricanKai (approved):
Not very happy with this, but I'll merge it for 2.7

Perksey:
For the record doing your ifdef in the generated code tactic has no real positives and quite a few negatives for example at this scale it will decrease performance doing this check at reparse time rather than generation time in addition to all of my reasons outlined above. You would be going out of your way to increase the compiler's load purely because you want the output to be the same between the target frameworks, which it effectively isn't anyway given disabled preprocessor branches are just text which the compiler discards - so really it's just a courtesy for... I have no idea who?

PR #587, opened 2021-08-19
Linux fix for Silk.NET Vulkan Triangle experiment.

usertoroot:
# Summary of the PR
The validation layer on all my computers is called `VK_LAYER_LUNARG_standard_validation` and not `VK_LAYER_KHRONOS_validation`. Changing this will allow the experiment application to run on my computers.

# Further Comments
As I understand from the Vulkan docs though, `VK_LAYER_LUNARG_standard_validation` is deprecated and `VK_LAYER_KHRONOS_validation` is recommended. However, this does not exist on any of my computers. Therefore the application will exit.
https://vulkan.lunarg.com/doc/view/1.1.114.0/windows/validation_layers.html

### DIFF
--- a/src/Lab/Experiments/VulkanTriangle/HelloTriangleApplication.cs
+++ b/src/Lab/Experiments/VulkanTriangle/HelloTriangleApplication.cs
@@ -71,7 +71,7 @@ namespace VulkanTriangle
         private KhrSurface _vkSurface;
         private KhrSwapchain _vkSwapchain;
         private ExtDebugUtils _debugUtils;
-        private string[] _validationLayers = { "VK_LAYER_KHRONOS_validation" };
+        private string[] _validationLayers = { "VK_LAYER_LUNARG_standard_validation" };
         private string[] _instanceExtensions = { ExtDebugUtils.ExtensionName };
         private string[] _deviceExtensions = { KhrSwapchain.ExtensionName };
 


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/Silk.NET?pullRequest=587) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/Silk.NET?pullRequest=587) before we can accept your contribution.<br/><br/>:x: usertoroot [sign now](https://cla.dotnetfoundation.org/dotnet/Silk.NET?pullRequest=587)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/Silk.NET?pullRequest=587) it.</sub>

Perksey:
Have you updated your Vulkan SDK and tried again? Not a fan of using a deprecated validation layer because it doesn't work on one user's machine...

usertoroot:
I am using linux, vulkan 1.2.131.2-1 standard with linux mint distro. Doubt i would be the only one affected. Seems to be post deprecation notice.

abenedik:
Because Vulkan is cross-platform, we cannot assume that some specific layers are supported. Therefore the proper way of using validation layers is to first check which are available and then enable one of those.

The following code can be used for that (taken from my engine):

`        
            string[]? layersToEnable;

            if (enableStandardValidation)
            {
                var availableInstanceLayerNames = VulkanInstance.GetAllAvailableInstanceLayerNames();
                layersToEnable = GetInstanceValidationLayerNames(availableInstanceLayerNames);
            }
            else
            {
                layersToEnable = null;
            }

private static string[]? GetInstanceValidationLayerNames(string[]? availableInstanceLayerNames)
        {
            if (availableInstanceLayerNames == null)
                return null;


            string[]? layersToEnable;

            if (Array.IndexOf(availableInstanceLayerNames, "VK_LAYER_KHRONOS_validation") != -1)
            {
                // new validation layer that is used instead of all previous validation layers
                layersToEnable = new[] { "VK_LAYER_KHRONOS_validation" };
            }
            else if (Array.IndexOf(availableInstanceLayerNames, "VK_LAYER_LUNARG_standard_validation") != -1)
            {
                // if new validation layer is not available, then try to use the old validations (standard validation is used for all other validations)
                layersToEnable = new[] { "VK_LAYER_LUNARG_standard_validation" };
            }
            else if (Array.IndexOf(availableInstanceLayerNames, "VK_LAYER_LUNARG_core_validation") != -1 &&
                     Array.IndexOf(availableInstanceLayerNames, "VK_LAYER_LUNARG_parameter_validation") != -1)
            {
                // if new validation layer is not available, then try to use the old validations (select two most important)
                layersToEnable = new[] { "VK_LAYER_LUNARG_core_validation", "VK_LAYER_LUNARG_parameter_validation" };
            }
            else
            {
                layersToEnable = null;
            }

            return layersToEnable;
        }

        public static unsafe string[]? GetAllAvailableInstanceLayerNames()
        {
            uint count = 0;
            Vk.EnumerateInstanceLayerProperties(&count, null).LogAndCheckResult(Logger, "vkEnumerateInstanceLayerProperties");

            if (count == 0)
                return null;

            var layerPropertiesLocal = stackalloc LayerProperties[(int)count];
            Vk.EnumerateInstanceLayerProperties(&count, layerPropertiesLocal).LogAndCheckResult(Logger, "vkEnumerateInstanceLayerProperties");


            var layerProperties = new string[(int)count];
            for (int i = 0; i < count; i++)
                layerProperties[i] = Marshal.PtrToStringAnsi((nint)layerPropertiesLocal[i].LayerName) ?? "";


            return layerProperties;
        }`

usertoroot:
I like the solution, prioritizing the non-deprecated stuff, however, given that this is supposed to be a minimal example (I guess?), it might not be an ideal solution to have such a large amount of code.

abenedik:
Yes this is surely not for a minimalistic example. 

But still the proposed PR is not correct - trying to enable the "VK_LAYER_LUNARG_standard_validation" will throw an exception when creating Vulkan instance because this validation is not supported on systems with newer Vulkan API. So if a fixed validation name needs to be chosen, then "VK_LAYER_KHRONOS_validation" is a better option as this will run on much more systems.

usertoroot:
"will throw an exception when creating Vulkan instance because this validation is not supported on systems with newer Vulkan API" is this confirmed?

abenedik:
I have just tried to enable VK_LAYER_LUNARG_standard_validation on Windows and I got ErrorLayerNotPresent error result code from vkCreateInstance call.

usertoroot:
Interesting! Good to know, so neither alone is a good solution then. Your suggestion if then the only good solution.

https://github.com/KhronosGroup/Vulkan-Samples/blob/96dc1b3d729a743b7bb1caf5b9a631a5e684538b/samples/api/hpp_hello_triangle/hpp_hello_triangle.cpp#L135

Seems like the triangle example by KhronosGroup has also been updated using a similar mechanism.

usertoroot:
I have nothing more to add for now. Will not have the time to implement a proper solution for that if it is even desired. Pull request can be closed as far as I'm concerned.

abenedik:
Maybe you should try again to update the Vulkan API. I am having the VK_LAYER_KHRONOS_validation available on Ubuntu.

usertoroot:
Well it's not that there's no solution, I just hoped that the deprecated one was available on old and new. Which does not seem to be the case (thank you for confirming). Updating Vulkan is not the point, I am on the latest version my distro has available (Linux Mint 20.1), which is a fairly widely used distro. 

Having examples work out of the box would have been good for Silk.NET so people would not have to dig.

HurricanKai:
I believe we should simply remove the validation from this. I know Vulkan-Tutorial.com does this, but in practice using vkconfig (which is a tool shipped with the SDK to configure validation externally) is much better and easier to use then manually enabling the layer.

Perksey:
Happy with either solution proposed - this is mostly just an experiment to only test our Vulkan bindings and isn't (or shouldn't be) treated as a concrete example because it likely has horrible practices, memory leaks, inefficiencies, or all of the above in it.

Perksey:
Closing due to:
- CLA not being signed
- A larger fix is required which the original author has indicated is out-of-scope

Issue #574, opened 2021-08-11
Add TryGetExtension methods AL & ALContext matching those BuildTools generates for OpenGL (et al)


Perksey:
Because the OpenAL bindings are handwritten (and the first set of bindings written for Silk.NET) they do not match our other bindings. One of the most notable differences is the lack of the TryGetExtension method in our OpenAL bindings, which is present in all of our other bindings.

This issue is to add a TryGetExtension method to the `AL` and `ALContext` classes. This method is simple (see [here](https://github.com/dotnet/Silk.NET/blob/main/src/OpenGL/Silk.NET.OpenGL/GL.cs#L87-L94)) and should be able to be just copied and pasted cleanly over.

Community contributions welcome.

Perksey:
This has been fixed in #598.

PR #598, opened 2021-09-01
Add TryGetExtension methods to AL & ALContext

DmitryGolubenkov:
# Summary of the PR
This PR adds `TryGetExtension` methods to `AL` and `ALContext` classes in `Silk.NET.OpenAL`

https://github.com/dotnet/Silk.NET/issues/574

# Further Comments
`TryGetExtension` method in `AL` is nearly the same as in OpenGL bindings. It was trivial to copy-paste and then compare to `GetExtension` to verify. The XML documentation was also taken from `GL` and modified.

However, `GetExtension` method in `ALContext` is different and also it doesn't use `ExtensionAttribute`. Also it requires a pointer to a `Device `to operate. So I am not sure if "copy-paste" approach would work. Should I convert `ALContext.GetExtension` to `TryGetExtension`?
Or maybe I am wrong, and it will be enough to copy existing method and import  `Silk.NET.Core.Attributes` namespace?
Just want to be sure.

### DIFF
--- a/src/OpenAL/Silk.NET.OpenAL/AL/AL.cs
+++ b/src/OpenAL/Silk.NET.OpenAL/AL/AL.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Numerics;
-using Microsoft.Extensions.DependencyModel;
 using Silk.NET.Core.Attributes;
 using Silk.NET.Core.Contexts;
 using Silk.NET.Core.Loader;
@@ -348,6 +347,27 @@ namespace Silk.NET.OpenAL
             ctx.Contexts[1] = new LamdaNativeContext
                 (x => x.EndsWith("GetProcAddress") ? default : ret.GetProcAddress(x));
             return ret;
+        }
+
+        /// <summary>
+        ///     Attempts to load a native OpenAL extension of type <typeparamref name="T" />.
+        /// </summary>
+        /// <param name="ext">
+        ///     The loaded extension.
+        /// </param>
+        /// <typeparam name="T">
+        ///     Type of <see cref="NativeExtension{T}" /> to load.
+        /// </typeparam>
+        /// <returns>
+        ///     <c>True</c> if the extension was loaded, otherwise <c>False</c>.
+        /// </returns>
+        public bool TryGetExtension<T>(out T ext)
+            where T : NativeExtension<AL>
+        {
+            ext = IsExtensionPresent(ExtensionAttribute.GetExtensionAttribute(typeof(T)).Name)
+                ? (T) Activator.CreateInstance(typeof(T), Context)
+                : null;
+            return ext != null;
         }
 
         /// <summary>
--- a/src/OpenAL/Silk.NET.OpenAL/AL/AL.cs
+++ b/src/OpenAL/Silk.NET.OpenAL/AL/AL.cs
@@ -375,6 +375,7 @@ namespace Silk.NET.OpenAL
         /// </summary>
         /// <typeparam name="TExtension">The extension type.</typeparam>
         /// <returns>The extension.</returns>
+        [Obsolete("This method has been deprecated and will be removed in Silk.NET 3.0. Please use TryGetExtension instead.")]
         public TExtension GetExtension<TExtension>()
             where TExtension : NativeExtension<AL>
         {
--- a/src/OpenAL/Silk.NET.OpenAL/AL/AL.cs
+++ b/src/OpenAL/Silk.NET.OpenAL/AL/AL.cs
@@ -350,24 +350,18 @@ namespace Silk.NET.OpenAL
         }
 
         /// <summary>
-        ///     Attempts to load a native OpenAL extension of type <typeparamref name="T" />.
+        /// Attempts to load a native OpenAL extension of type <typeparamref name="T" />.
         /// </summary>
-        /// <param name="ext">
-        ///     The loaded extension.
-        /// </param>
-        /// <typeparam name="T">
-        ///     Type of <see cref="NativeExtension{T}" /> to load.
-        /// </typeparam>
-        /// <returns>
-        ///     <c>True</c> if the extension was loaded, otherwise <c>False</c>.
-        /// </returns>
+        /// <param name="ext">The loaded extension.</param>
+        /// <typeparam name="T">Type of <see cref="NativeExtension{T}" /> to load.</typeparam>
+        /// <returns><c>true</c> if the extension was loaded, otherwise <c>false</c>.</returns>
         public bool TryGetExtension<T>(out T ext)
             where T : NativeExtension<AL>
         {
             ext = IsExtensionPresent(ExtensionAttribute.GetExtensionAttribute(typeof(T)).Name)
                 ? (T) Activator.CreateInstance(typeof(T), Context)
                 : null;
-            return ext != null;
+            return ext is not null;
         }
 
         /// <summary>
@@ -375,7 +369,11 @@ namespace Silk.NET.OpenAL
         /// </summary>
         /// <typeparam name="TExtension">The extension type.</typeparam>
         /// <returns>The extension.</returns>
-        [Obsolete("This method has been deprecated and will be removed in Silk.NET 3.0. Please use TryGetExtension instead.")]
+        [Obsolete
+        (
+            "This method has been deprecated and will be removed in Silk.NET 3.0. " +
+            "Please use TryGetExtension instead."
+        )]
         public TExtension GetExtension<TExtension>()
             where TExtension : NativeExtension<AL>
         {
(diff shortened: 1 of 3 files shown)


Perksey:
The ALContext methods I’d expect to mirror the ones in Vk and XR. All extension-related methods should use ExtensionAttribute. Hope this helps!

DmitryGolubenkov:
Thanks, that helped a bit. I did some research about the codebase and also looked into documentation to better understand what is going on.  And now I agree that `TryGetExtension` would nearly mirror the methods in `Vk` and `XR`. 

Anyway, I noticed that `GetExtension` in `ALContext` requires a pointer to a device, but the parameter is not used.  At least it is not obvious. Is there any reason for that?

You mentioned an extra parameter. There is no separate `Instance` in OpenAL wrappers, and the `Device` struct is empty and is used as a pointer. So I don't understand what that parameter is.

I'll make a commit so that you could check if the method is correct. Then I'll add XML documentation

src/OpenAL/Silk.NET.OpenAL/ALC/ALContext.cs
```diff
@@ -110,6 +111,12 @@ public static unsafe ALContext GetApi(bool soft = false)
             return ret;
         }
 
+        public bool TryGetExtension<T>
+            (out T ext) where T : NativeExtension<ALContext> => 
+            !((ext = IsExtensionPresent(ExtensionAttribute.GetExtensionAttribute(typeof(T)).Name)
+                ? (T) Activator.CreateInstance(typeof(T), Context)
+                : null) is null);
+
```
Perksey, line 119: This method should have a `Device*` parameter which you feed in as the first parameter of `IsExtensionPresent`. It's awful we weren't using that in `GetExtension` - we really should have been!

Perksey (commented):
As requested, an initial lookover (not a full review)

Some comments:
- If you wanted to add `[Obsolete("This method has been deprecated and will be removed in Silk.NET 3.0. Please use TryGetExtension instead.")]` to `GetExtension` on both classes, that'd be much appreciated.
- If you'd like to add `[Obsolete("This class is deprecated and will be removed in Silk.NET 3.0. Please use the TryGetExtension method on ALContext.")]` to `ALExtensionLoader` again would be much appreciated

src/OpenAL/Silk.NET.OpenAL/ALC/ALContext.cs
DmitryGolubenkov: Should XML documentation contain a remark `This function doesn't check that the extension is enabled - you will get an error later on if you attempt to call an extension function from an extension that isn't loaded.`? If I understand correctly, `ALContext.TryGetExtension` also doesn't check if it is enabled.

If yes, I think I should update the `AL.TryGetExtension`  XML too.

Perksey: OpenAL doesn't have a concept of enabling extensions like Vulkan does - with Vulkan if you try to get the address to a function in an extension which is not enabled, you will get a symbol loading exception whereas OpenAL allows it and automatically grants access to all extensions available on that implementation. There is `alEnable` which enables the extension's functionality in some cases, but that's different to enabling the extension itself. So I would say leave that out.

src/OpenAL/Silk.NET.OpenAL/AL/AL.cs
```diff
@@ -350,11 +349,33 @@ public static AL GetApi(bool soft = false)
...
+        /// </summary>
+        /// <param name="ext">
+        ///     The loaded extension.
+        /// </param>
+        /// <typeparam name="T">
+        ///     Type of <see cref="NativeExtension{T}" /> to load.
+        /// </typeparam>
+        /// <returns>
+        ///     <c>True</c> if the extension was loaded, otherwise <c>False</c>.
+        /// </returns>
```
Perksey, line 363: style:
- `summary` is always in "paragraph format" where the tags are on separate lines to the content
- `param`s and `typeparam`s are on the same line as their tags except where that would exceed 120 characters on that line, in which case "paragraph format" is used.
- "paragraph format" shouldn't have an indent (unlike regular XML)
- `c` and `code` elements should be valid C# code.

```suggestion
        /// <summary>
        /// Attempts to load a native OpenAL extension of type <typeparamref name="T" />.
        /// </summary>
        /// <param name="ext">The loaded extension.</param>
        /// <typeparam name="T">Type of <see cref="NativeExtension{T}" /> to load.</typeparam>
        /// <returns><c>true</c> if the extension was loaded, otherwise <c>false</c>.</returns>
```

src/OpenAL/Silk.NET.OpenAL/AL/AL.cs
```diff
@@ -350,11 +349,33 @@ public static AL GetApi(bool soft = false)
...
+        public bool TryGetExtension<T>(out T ext)
+            where T : NativeExtension<AL>
+        {
+            ext = IsExtensionPresent(ExtensionAttribute.GetExtensionAttribute(typeof(T)).Name)
+                ? (T) Activator.CreateInstance(typeof(T), Context)
+                : null;
+            return ext != null;
```
Perksey, line 370: Prefer identity comparisons
```suggestion
            return ext is not null;
```

src/OpenAL/Silk.NET.OpenAL/AL/AL.cs
```diff
@@ -350,11 +349,33 @@ public static AL GetApi(bool soft = false)
...
+        }
+
         /// <summary>
         /// Gets an instance of the API of an extension to the API.
         /// </summary>
         /// <typeparam name="TExtension">The extension type.</typeparam>
         /// <returns>The extension.</returns>
+        [Obsolete("This method has been deprecated and will be removed in Silk.NET 3.0. Please use TryGetExtension instead.")]
```
Perksey, line 378: I know I gave you this code but it exceeds 120 characters.
```suggestion
        [Obsolete
        (
            "This method has been deprecated and will be removed in Silk.NET 3.0. " +
            "Please use TryGetExtension instead."
        )]
```

src/OpenAL/Silk.NET.OpenAL/ALC/ALContext.cs
```diff
@@ -110,12 +111,26 @@ public static unsafe ALContext GetApi(bool soft = false)
...
+        /// </summary>
+        /// <typeparam name="T">The extension type.</typeparam>
+        /// <param name="device">The device the context is on.</param>
+        /// <param name="ext">The extension to check for.</param>
+        /// <returns>Whether the extension is available.</returns>
+        public unsafe bool TryGetExtension<T>
+            (Device* device, out T ext) where T : NativeExtension<ALContext> => 
+            !((ext = IsExtensionPresent(device, ExtensionAttribute.GetExtensionAttribute(typeof(T)).Name)
+                ? (T) Activator.CreateInstance(typeof(T), Context)
+                : null) is null);
```
Perksey, line 125: Style

```suggestion
        public unsafe bool TryGetExtension<T>(Device* device, out T ext) where T : NativeExtension<ALContext>
            => !((ext = IsExtensionPresent(device, ExtensionAttribute.GetExtensionAttribute(typeof(T)).Name)
                ? (T) Activator.CreateInstance(typeof(T), Context)
                : null) is null);
```

src/OpenAL/Silk.NET.OpenAL/ALC/ALContext.cs
```diff
@@ -110,12 +111,26 @@ public static unsafe ALContext GetApi(bool soft = false)
...
+
         /// <summary>
         /// Gets an instance of the API of an extension to the API.
         /// </summary>
         /// <typeparam name="TContextExtension">The extension type.</typeparam>
         /// <param name="device">The device the context is on.</param>
         /// <returns>The extension.</returns>
+        [Obsolete("This method has been deprecated and will be removed in Silk.NET 3.0. Please use TryGetExtension instead.")]
```
Perksey, line 133: Same again, apologies, my bad!

```suggestion
        [Obsolete
        (
            "This method has been deprecated and will be removed in Silk.NET 3.0. " +
            "Please use TryGetExtension instead."
        )]
```

src/OpenAL/Silk.NET.OpenAL/Extensions/ALExtensionLoader.cs
```diff
@@ -11,6 +11,7 @@ namespace Silk.NET.OpenAL.Extensions
     /// <summary>
     /// A loader for OpenAL extensions.
     /// </summary>
+    [Obsolete("This class is deprecated and will be removed in Silk.NET 3.0. Please use the TryGetExtension method on ALContext.")]
```
Perksey, line 14:
```suggestion
    [Obsolete
    (
        "This class is deprecated and will be removed in Silk.NET 3.0. "+
        "Please use the TryGetExtension method on ALContext."
    )]
```

Perksey (changes requested):
Thanks so much! Once again, your code is superb it's just missing a few stylistic choices we make in our codebase which could do with being implemented. All one click changes so no need to open the IDE :)

Perksey (approved):

PR #603, opened 2021-09-03
Promote ImGui lab application to an example

Perksey:
Does what it says on the tin.


Perksey:
Ready for review but not release critical.

HurricanKai (approved):
Assuming the original files were good and this works.

Issue #597, opened 2021-08-30
Create a .NET 6 code path for GlobalMemory


Perksey:
In .NET 6, new `NativeMemory` APIs were introduced to enable direct access to the underlying C standard library's implementation of `malloc` and `free`. Given we just need pinned memory, we should use this instead of `GC.AllocateUninitializedArray` - using that method can cause heap fragmentation given the pinned object heap is intended for long-lived objects, and our users quite often use SilkMarshal.Alloc and our GlobalMemory APIs for short to medium lived blocks of memory. In C `malloc` is used for just about everything, so we should follow suit.

The work is:
- [ ] Add a `net6.0` target in the [csproj](https://github.com/dotnet/Silk.NET/blob/main/src/Core/Silk.NET.Core/Silk.NET.Core.csproj#L4)
- [ ] Add a `NativeMemoryPtr` type in [GlobalMemory](https://github.com/dotnet/Silk.NET/blob/main/src/Core/Silk.NET.Core/Native/GlobalMemory.cs#L181) which mirrors `HGlobal`, except it uses [`NativeMemory`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.nativememory) to allocate instead
- [ ] In [`GlobalMemory.Free`](https://github.com/dotnet/Silk.NET/blob/main/src/Core/Silk.NET.Core/Native/GlobalMemory.cs#L121), add a new switch `case` for the new type to free the handle using `NativeMemory.Free`
- [ ] In [`GlobalMemory.Allocate`](https://github.com/dotnet/Silk.NET/blob/main/src/Core/Silk.NET.Core/Native/GlobalMemory.cs#L162), add a preprocessor branch for `NET6_0` to use `NativeMemoryPtr` as the memory object.

Community contributions welcome.

Perksey:
@HurricanKai before I mark as up-for-grabs, comments?

HurricanKai:
sounds good to me

Perksey:
Will leave this is a community issue, no rush to get this in for 2.8.

Perksey:
Closed by #609

PR #609, opened 2021-09-06
Create a .NET 6 code path for GlobalMemory

DmitryGolubenkov:
# Summary of the PR
- Added a net6.0 target to `Silk.NET.Core.csproj`
- Added a new unsafe struct `NativeMemoryPointer` inside a preprocessor `NET6_0_OR_GREATER`
- Modified `GlobalMemory.Free` and `GlobalMemory.Allocate` to use the new struct.

#597 

# Further Comments
I don't know if this PR require testing or contain breaking changes as I am not aware of any possible problems.

I had to add `unsafe` to `GlobalMemory.Free` to use `NativeMemory` because that is required. And this change will affect .NET 5 and .NET Standart versions even if this change is only for .NET 6.
Also: maybe it is better to use `NET6_0_OR_GREATER` directive in `GlobalMemory.Free`? Because `NET6_0` targets only one version, and I think next year someone will have to do this change anyway.

`NativeMemoryPtr` requires minimum .NET 6 too. Also I had to make some convertions line `(nint)` and `(nuint)` to be able to use `NativeMemory`.

I think that `GlobalMemory.Allocate` requires a little work, because the preprocesor directives now are.. Confusing.
Now it is:
`#if NET6_0`
`#elif !NET5_0`
`#else`

I think it is better to make 
`#if NET6_0`
`#elif NET5_0`
`#else`
and make according changes to the code. Only to make the code more readable.


HurricanKai:
Thank you for you contribution!
I agree that the preprocessors in Allocate are confusing, if possible do change that.
Also, there are a lot of places in Silk.NET.Core where NET5_0 is used, and those codepaths should be available on NET6_OR_HIGHER as well, if possible please change those as well.

DmitryGolubenkov:
Should I make another PR to make these changes to whole Silk.NET.Core? Or just commit everything to this?

HurricanKai:
Committing to this is fine

Perksey (approved):

Perksey:
Thanks again for the shear volume of contributions as of late! Really appreciate it :)

PR #612, opened 2021-09-14
Website

Perksey:
We should start looking to get this merged in, even if it doesn't have much content yet. This will also allow other contributors to add stuff to the site.

# Progress
## In scope for this PR
- [x] Basic theme
- [x] Generating HTML from content using Statiq
- [x] Table of contents system
- [x] Blogging
    - [x] RSS feeds
- [x] Home page

## Not in scope
- FAQ
- Web tutorials
- Other documentation
- API reference


.github/workflows/publish-site.yml
```diff
@@ -0,0 +1,27 @@
+name: Website Build
+on:
+  push:
+    branches:
+      - 'feature/heavy-wip-website'
```
HurricanKai, line 5: Should probably add main here?

.github/workflows/publish-site.yml
```diff
@@ -0,0 +1,27 @@
...
+  pull_request:
+    # none
+jobs:
+  Build:
+    runs-on: windows-latest
```
HurricanKai, line 10: Does this have to be windows? They are usually harder to aquire and given we end up in a queue always already, Ubuntu would be preferred I think.

.github/workflows/publish-site.yml
```diff
@@ -0,0 +1,27 @@
...
+        dotnet-version: 6.0.100-preview.7.21379.14
+    - name: Run Statiq
+      run: dotnet run -c Release --project src/Website/Silk.NET.Statiq/Silk.NET.Statiq.csproj
+    - name: Push to Website Branch
+      run: |
+        git config --local user.email "9011267+dotnet-bot@users.noreply.github.com"
+        git config --local user.name "The Silk.NET Automaton"
+        git add -f docs/
+        git commit -am "Push latest built in-development GitHub Pages site"
+        git push -f origin HEAD:ci/github-pages
```
HurricanKai, line 27: There has to be a way to not make this force pushed always. Can't we just merge this?

.gitmodules
```diff
@@ -3,4 +3,4 @@
 	url = https://github.com/libsdl-org/SDL
 [submodule "build/submodules/Vulkan-Headers"]
 	path = build/submodules/Vulkan-Headers
-	url = https://github.com/KhronosGroup/Vulkan-Headers
+	url = https://github.com/KhronosGroup/Vulkan-Headers
```
HurricanKai, line 6: What has changed here?

website/_Layout.cshtml
```diff
@@ -0,0 +1,56 @@
...
+  <link href="@InputUrl("theme/argon.css")" rel="stylesheet" />
+  <link rel="stylesheet" href="https://cdn.jsdelivr.net/docsearch.js/2/docsearch.min.css" />
+  <link rel="stylesheet" href="@InputUrl("theme/docs.css")" />
+  <link rel="stylesheet" href="@InputUrl("theme/silk.css")" />
+</head>
+
+<body>
+  @Html.Partial("_theme/src/10-nav.cshtml")
+  @Html.Partial("_theme/src/20-contentroot.cshtml", new ViewDataDictionary(ViewData) { { "body", RenderBody() }})
+  <script src="@InputUrl("theme/jquery.min.js")" type="text/javascript"></script>
```
HurricanKai, line 35: These are a lot of JavaScript files, do they have to all be there?

website/_theme/lib/application.js
```diff
@@ -0,0 +1,119 @@
+// NOTICE!! DO NOT USE ANY OF THIS JAVASCRIPT
```
HurricanKai, line 1: ?

src/Website/Silk.NET.Statiq/Utilities.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    public abstract class SilkPage : StatiqRazorPage<IDocument>
+    {
+        /// <summary>
+        /// Converts a path which is relative to the input path to a URL which is relative to the current page's
+        /// containing directory.
+        /// </summary>
+        /// <param name="path">The path.</param>
+        /// <returns>The URL.</returns>
+        public string InputUrl(string path)
```
HurricanKai, line 18: Why is this needed? Seems like it's only used for the JS/CSS files, which should be absolute to the root anyways, which is just an HTTP feature. Just start the path of a referenced file with a leading '/' and then it'll be an absolute path.

HurricanKai (changes requested):

src/Website/Silk.NET.Statiq/Utilities.cs
Perksey: This is used for all URLs. I don't want to have it use `/something`, I want to ensure all URLs are relative mainly to keep the site operable outside of a server environment as well as for future enhancements such as pull request publishing (i.e. we could have `https://dotnet.github.io/Silk.NET/pr/612` so we can view the rendered webpages, something I can see being useful when we have tutorial PRs)

.github/workflows/publish-site.yml
Perksey: Will do in a follow up PR immediately after this is merged.

.github/workflows/publish-site.yml
Perksey: Let's see...
```suggestion
    runs-on: ubuntu-latest
```

website/_theme/lib/application.js
Perksey: lol, deleted.

website/_Layout.cshtml
Perksey: Deleted some which were pulled in with the template used that didn't look to be doing much.

.github/workflows/publish-site.yml
Perksey: I'll align the histories after merge, but during development of the website this isn't really possible without the histories going wibbly wobbly timey wimey. Will do this in a follow-up MR immediately after merge.

.gitmodules
Perksey: probably line endings idk, innocent in any case

Perksey: oh i know now, fixed

```suggestion
	url = https://github.com/KhronosGroup/Vulkan-Headers

```

Perksey:
Happy with this, probably won't do any further work on this myself for a little bit. Ready for review as an early prototype to get merged in for further iterative work.

.github/workflows/publish-site.yml
WhiteBlackGoose: There's [Action](https://github.com/marketplace/actions/deploy-to-github-pages) for it

HurricanKai (approved):

Issue #641, opened 2021-10-16
Add a Samples window option


Perksey:
# Summary of feature
GLFW and SDL have support for OpenGL multisampling, so we should expose this in the windowing library.

- GLFW: `GLFW_SAMPLES` (WindowHintInt)
- SDL: `SDL_GL_MULTISAMPLESAMPLES` (GLattr)

As such, we should also have support for this. The work is:
- [ ] Add a get-only `Samples` property of type `int?` in `IViewProperties`
- [ ] Add a gettable and settable `Samples` property of type `int?` in `ViewOptions` and `WindowOptions`
    - [ ] Initialise these with a default value of `null` in all of the constructors
- [ ] Implement it in Silk.NET.Windowing.Glfw
    - [ ] Set a WindowHint [before](https://github.com/dotnet/Silk.NET/blob/main/src/Windowing/Silk.NET.Windowing.Glfw/GlfwWindow.cs#L342) creating the window only if the `Samples` property **is not** `null`
- [ ] Implement it in Silk.NET.Windowing.Sdl
    - [ ] When [creating the context](https://github.com/dotnet/Silk.NET/blob/main/src/Windowing/Silk.NET.Windowing.Sdl/SdlView.cs#L169), if the `Samples` property **is not** `null` then set the `GLattr`. You may need to refactor this code as the code doesn't leave a lot of wiggle room for that.

# Comments
SDL has support for selecting the number of multisampling buffers (`SDL_GL_MULTISAMPLESAMPLES`) but GLFW does not, so that will not be implemented here.

# Does this have a proposal?
No, scope is not large enough to warrant one.

PR #660, opened 2021-10-24
Add a Samples window option

ThomasMiz:
# Summary of the PR
Implemented multisample support for the window's framebuffer, as highlighted in issue #641.


ThomasMiz:
Latest commit provides a fix for SDL problems. Many attributes that are set on context creation seem to require to be actually set before window creation. Changing this fixes problems where SDL was not recognizing certain attributes being set.

This fixed multisampling not working on SDL, plus the SDL backend now recognizes the `PreferredDepthBufferBits` and `PreferredStencilBufferBits` options, which should fix #662.

Perksey (approved):

PR #661, opened 2021-10-24
Build improvements & dynamic Windows SDK resolution (backport from 3.0)

### LINKED ISSUE #630: Incorrect string marshaling when "ID3D12Object::SetName" was called
# Summary
`SetName("DepthStencil")`
But I got
`(0x000001CAA011C1C0:'Dphtni)`

# Steps to reproduce
- Platform:Desktop
- Framework Version: .NET 6
- API: DirectX
- API Version: DirectX 12

```C#
ID3D12Resource* depthStencil;
.......... Some Code Here ............
depthStencil->SetName($"DepthStencil");
```
Then I got this error:
```
Execution (From D3D12): ID3D12CommandQueue::ExecuteCommandLists: An ID3D12Resource object (0x000001CAA011C1C0:'**Dphtni**) referenced in a command list ............
```

# Comments
It looks like Silk is using `Marshal.StringToHGlobalAnsi()` rather than `StringToBSTR()`, which drops half of the letters due to the encoding.

image

image

### LINKED ISSUE #652: [Linux] GLFW/SDL native library not loaded
# Summary
On linux the simplest sample fails because windowing platform fails to load native library.
Ultz.Native does place .so file, but it is under `runtimes/linux-x64/native` however the default path resolvers does not check this path.

# Steps to reproduce
- Platform: Desktop/Linux
- Framework Version: .NET 5
- API: GLFW
1. Created a new .net5 project (I was using Rider for that)
2. Added Silk.NET reference
3. Copied sample code from https://github.com/dotnet/Silk.NET/blob/main/examples/CSharp/OpenGL%20Tutorials/Tutorial%201.1%20-%20Hello%20Window/Program.cs

The application fails to run.

# Comments

I get exception: `The currently bound window platform(s) only support views, instead of windows. Use the view APIs instead.`
After some debugging I narrowed it down to inability to load glfw library even though it is available at output `runtimes/linux-x64/native/libglfw.so.3`.

If I set `LD_LIBRARY_PATH=runtimes/linux-x64/native` then I can run the example.
However I think that the default loader could also check this directory for native libraries.

Perksey:
- Updates Vulkan and Vulkan Video
- Updates to Windows 11 SDK
- Adds dynamic Windows SDK resolution (partially backported from #559) so that the MSVC and Windows SDK paths aren't hardcoded and will work wherever you have the Windows SDK and Visual Studio installed.
- Adds a "native" feature-set in NUKE to make doing the gradle bits optional (cc @ThomasMiz)
- Uses `dotnet nuget push` instead of the NuGet .NET API in NUKE
- Fixes #630 
- Fixes #652


Perksey:
This is ready for review.

HurricanKai (approved):
I've not reviewed everything here, but it's largely autogenerated and they generally look good.

PR #664, opened 2021-11-02
Changes to the math simd proposal

WhiteBlackGoose:
# Summary of the PR
Some methods in the prop operating on two arguments for some reason had the second argument as just `T` whereas it should be Vector<T>, I think.


Perksey (approved):
Happy with this, and happy to forgo taking this back to the working group seeing as this was an oversight and not a functionality change. @HurricanKai fine with this?

PR #665, opened 2021-11-02
SIMD for Silk.NET.Math

WhiteBlackGoose:
## Summary of the PR
Starting implementing the [SIMD proposal](https://github.com/dotnet/Silk.NET/blob/main/documentation/proposals/Proposal%20-%20Vectorization%20-%20SIMD.md).

## Plan for this PR

All methods are for three bitnesses: 64, 128, 256.

- [x] `Create`
- [x] `Load`
- [x] `Add`
- [x] `Subtract`
- [x] `Multiply`
- [x] `Divide`
- [x] `Abs`
- [x] Constants
- [x] `And`
- [x] `Or`
- [x] `Xor`
- [x] `Not`
- [x] `AllBitsSet`
- [x] `IsHardwareAccelerated`

Other:
- [x] Docs


src/Maths/Silk.NET.Maths/Simd/Simd128.Abs.cs
```diff
@@ -0,0 +1,168 @@
...
+            static Vector128<T> Single(Vector128<T> vector)
+            {
+                if (typeof(T) == typeof(float))
+                {
+#if AdvSIMD
+                    return AdvSimd.And(vector.AsSingle(), Vector128.Create(0x7FFF_FFFFu).AsSingle()).As<float, T>();
+#endif
+#if SSE
+                    return Ssse3.And(vector.AsSingle(), Vector128.Create(0x7FFF_FFFFu).AsSingle()).As<float, T>();
+#endif
```
HurricanKai, line 129: I think just calling down to Simd128.And is preferable here?

src/Maths/Silk.NET.Maths/Simd/Simd128.Add.cs
```diff
@@ -0,0 +1,306 @@
...
+                    {
+                        return Sse2.Add(left.AsByte(), right.AsByte()).As<byte, T>();
+                    }
+#endif
+#if AVX
+                    if (Avx2.IsSupported)
+                    {
+                        return Avx2.Add(left.AsByte(), right.AsByte()).As<byte, T>();
+                    }
+#endif
```
HurricanKai, line 49: This doesn't make sense - if AVX2 is supported, SSE2 is as well, so the lower branch is never hit - but I'm unsure whether we should prefer SSE or AVX here
Edit: I think usually the AVX method for 128bit is just the same as the SSE one, as the Avx2 type literally inherits from SSE42

src/Maths/Silk.NET.Maths/Simd/Simd128.Divide.cs
```diff
@@ -0,0 +1,88 @@
...
+{
+    public static unsafe partial class Simd128
+    {
+        /// <summary>
+        /// Performs hardware-accelerated Divide on 128-bit vectors.
+        /// </summary>
+        [MethodImpl(Scalar.MaxOpt)]
+        public static Vector128<T> Divide<T>(Vector128<T> left, Vector128<T> right) where T : unmanaged
+        {
+
```
HurricanKai, line 25: Consider adding a TODO and raising an issue that non-floating point types need to be added here.

src/Maths/Silk.NET.Maths/Simd/Simd128.Multiply.cs
```diff
@@ -0,0 +1,190 @@
...
+            static Vector128<T> Byte(Vector128<T> left, Vector128<T> right)
+            {
+                if (typeof(T) == typeof(byte))
+                {
+#if AdvSIMD
+                    if (AdvSimd.IsSupported)
+                    {
+                        return AdvSimd.Multiply(left.AsByte(), right.AsByte()).As<byte, T>();
+                    }
+#endif
```
HurricanKai, line 37: Please add an SSE path (see https://stackoverflow.com/questions/8193601/sse-multiplication-16-x-uint8-t)

HurricanKai (changes requested):
Reviewed SSE for now - I think the review comments apply to the other ones as well so not reviewing the rest

src/Maths/Silk.NET.Maths/Simd/Simd128.Abs.cs
WhiteBlackGoose: But it doesn't exist 😢 

Hmm

HurricanKai: I'm fine mering it if it doesn't compile 😛

src/Maths/Silk.NET.Maths/Simd/Simd128.Add.cs
WhiteBlackGoose: > This doesn't make sense - if AVX2 is supported, SSE2 is as well, so the lower branch is never hit - but I'm unsure whether we should prefer SSE or AVX here

Hmmmm. So I remove avx then?

HurricanKai: Yes

src/Maths/Silk.NET.Maths/Simd/Simd128.Multiply.cs
WhiteBlackGoose: May I leave a todo here too? To keep PRs not too heavy

HurricanKai: Sure

src/Maths/Silk.NET.Maths/Simd/Simd128.Add.cs
WhiteBlackGoose: I better remove sse

HurricanKai: II don't think that's better, SSE is available on NETCore 3.1, AVX isn't, I think

HurricanKai: Nevermind, but the check would still prevent some machines that can actually run this from running it, so SSE is preferable

src/Maths/Silk.NET.Maths/Simd/Simd128.Load.cs
```diff
@@ -0,0 +1,318 @@
...
+                }
+                
+        
+                return Other(ptr);
+            }
+            
+            [MethodImpl(Scalar.MaxOpt)]
+            static Vector128<T> Other(T* ptr)
+            {
+                            return Unsafe.ReadUnaligned<Vector128<T>>(ptr);
```
Perksey, line 311: What if T is not 32-bit?

Perksey (commented):
Not codeowner, so these are just general things I noticed in passing: kai can decide whether to care.

src/Maths/Silk.NET.Maths/Simd/Simd128.Divide.cs
Perksey: Might be sensible to have a general SIMD tracking issue as there'll probably be lots of notes that we'll want to keep track of.

HurricanKai: Seems fair to me as well

src/Maths/Silk.NET.Maths/Simd/Simd128.Load.cs
WhiteBlackGoose: Why does it matter?

Perksey: Because you'd get a segfault. This method is meant to load a pointer containing 4 Ts, not a pointer containing a 128-bit vector.

src/Maths/Silk.NET.Maths/Simd/Simd.GetWithElement.cs
```diff
@@ -0,0 +1,91 @@
...
+    {
+        // these two should ONLY be used in the constant-sized unrolled loops.
+        // unfortunately there is an issue otherwise. See https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAJI8ovLrl5hcAbho1iAZiakGAYRoBvGg11MUDACoxcGABQBKHXu3U99hgDNoDM+IwNeDALwM6MzwYAHgYANRgwDGhyUgAOIMcAGwhsDAA+FjUIbgwA3gBqfKs7B1KAOQgAeQAHNws5Et0AX2tdVv0GCpq3Hk9i+1tS3V5HVzMOdwtAtN8AegAqccn52fDI6LiE5NSMrJyp2dnguDgGbkh8AV6wPAx2oaZyAE5LBvsW6iagA
+        [MethodImpl(Scalar.MaxOpt)]
+        private static Vector128<T> WithElement<T>(Vector128<T> vector, int index, T value) where T : unmanaged
+        {
+            Vector128<T> result = vector;
+            ref T e0 = ref Unsafe.As<Vector128<T>, T>(ref result);
+            Unsafe.Add(ref e0, index) = value;
+            return result;
```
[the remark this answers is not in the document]
WhiteBlackGoose: This API existed before me, so I assumed there's no such 😅 
Will fix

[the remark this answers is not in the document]
WhiteBlackGoose: Actually @HurricanKai was there a reason to have those methods, given that vector provides those methods?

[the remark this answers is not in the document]
HurricanKai: Well, I don't remember why I did this, but I think something something JIT is confused.
I mean change it and check with disasmo, or try searching #lowlevel for this code

src/Maths/Silk.NET.Maths/Simd/Simd128.Add.cs
WhiteBlackGoose: Oh, no, I will remove avx for 128 bit

src/Maths/Silk.NET.Maths/Simd/Simd.GetWithElement.cs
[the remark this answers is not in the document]
Perksey: These are aggressively inlined and do the same thing under-the-hood now, only it's validated and is `[Intrinsic]` (so the JIT may do special things if it feels like it!).

source: source.dot.net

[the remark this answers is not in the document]
HurricanKai: I know them, but I remember something about it being wrong. As I said, try it.

src/Maths/Silk.NET.Maths/Simd/Simd128.Abs.cs
WhiteBlackGoose: Uh. That would make your job of reviewer much more difficult

src/Maths/Silk.NET.Maths/Simd/Simd128.Load.cs
WhiteBlackGoose: Oh, right. It's more than that, it should convert `ptr` to `Vector128<T>*`, that's why

src/Maths/Silk.NET.Maths/Simd/Simd128.Abs.cs
WhiteBlackGoose: I can probably add `And` as part of this PR too

src/Maths/Silk.NET.Maths/Simd/Simd.GetWithElement.cs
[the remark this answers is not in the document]
WhiteBlackGoose: [Here](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABABgAJiBGAOgCUBXAOwwEt8YaBhCfAB1YAbGFADKIgG6swMXAG4AsACgylWoxbtOASRZRWTXNPnLV1eszYcaujPsPGaADQAcSRUtNUk5AGowwDGgAJgBWJAAeAwwAPnJxDFIMAAsAFQhSAH0AaWxWAAp/QJDwqJY4iQBKZXJa8gBeOIB1VhSAUWEOFnyJNHJSPtJKjy8fIqCoMMjouISktIzMgCEuABlCgImpstjyKpq6xr2aFvbOmG6B/uHlW6UAbQBZGBSIABNtfkF859ePr4A8nw2BBDDQAIIAc0hsFwRgkMF0ggMBkhlQAuqM/JsSpFUs1WskOjAuhgIviNsVJqV8XscVA+tFyAY3jAEH1UntsIIGDBKuQAO7JEQwcickDkZj4bBMbCQmBvZQAbwOtXGuPJcVhDEEGAadKpHjq5FgADMxeQYBR6iaYOaAKqGbCmzjg3ARdXUvExDkxfJm224HUYG5KY2O3DO11vN7+u2Wq4stn8m0Sbm8o11YgAdkDwY8AF8gA===) BCL's WithElement gives a better codegen than manually written. Also we probably should rely on BCL anyway?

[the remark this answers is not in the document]
WhiteBlackGoose: [128 bit](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABABgAJiBGAOgCUBXAOwwEt8YaBhCfAB1YAbGFADKIgG6swMXAG4AsACgylWoxbtOASRZRWTXNPnLV1eszYcaujPsPGaADQAcSRUtNUk5AGowwDGgqACYXAB4DDAA+cnEMUgwACwAVCFIAfQBpbFYACn9A4LDIlliJAEplchryAF5YgHVWZIBRYQ4WPIk0clJe0gqPLx9CoKhQiKjY+MTU9IyAIS4AGQKA8cnSmPJK6tqG3Zpmto6YLv6+oeUbpQBtAFkYZIgAE21+QTynl/fPgHk+GwIIYaABBADmENguCMEhgukEBgMEIqAF0Rn4NsUIikmi0ku0YJ0MOE8esihMSnjdtioL0ouQDK8YAheildthBAwYBVyAB3JIiGDkDkgcjMfDYJjYCEwV7KADe+xqYxxZNiMIYggw9VplI8tXIsAAZqLyDAKHVjTAzQBVQzYE2cMG4cJqqm46Ls6J5U023DajDXJRGh24J0u16vP22i2XZmsvnWiRcnmG2rEADsAaDHgAvkA=) too, for 64-bit it's [about the same](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABABgAJiBGAOgCUBXAOwwEt8YaBhCfAB1YAbGFADKIgG6swMXAG4AsACgylWoxbtOASRZRWTXNPnLV1eszYcaujPsPGaADQAcSRUtNUk5AGowwDGhUAB4DDAA+cnEMUgwACwAVCFIAfQBpbFYACn9A4JQwliiJAEplckryAF4ogHVWBIBRYQ4WbIk0clIu0lKPLx88oKhQ8KiYuKSU1IAhLgAZXICRseLyMoqq2o2aBubWmHae7v7lc6UAbQBZGASIABNtfkFs2/unl4B5PjYIQxoAEEAObA2C4IwSGC6QQGAzA0oAXUGfhWBRCiXqjXiLRgbQwGIiy3yo0KmI2aKgXXC5AMDxgCC6iQ22EEDBgpXIAHd4iIYORmSByMx8NgmNhgTAHsoAN5bSrDdHk8EMQQYGoUkkeKrkWAAMwF5BgFGqupgBoAqoZsHrOIDcCFFaTCUyifqzbhVRgzkodVbcDa7Q8Htl3cbqUx6QhOaaJKz2dqqsQAOwer0eAC+QA===).

src/Maths/Silk.NET.Maths/Simd/Simd128.Load.cs
WhiteBlackGoose: @Perksey we're reading a 128-bit vector, so the bitness of T doesn't matter

src/Maths/Silk.NET.Maths/Simd/Simd.GetWithElement.cs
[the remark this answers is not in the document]
HurricanKai: You need to look at the [loop version](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABABgAJiBGAOgCUBXAOwwEt8YaBhCfAB1YAbGFADKIgG6swMXAG4AsACgylWoxbtOASRZRWTXNPnLV1eszYcaujPsPGaADQAcSRUtNUk5AGowwDGgAJgBWJAAeAwwAPnJxDFIMAAsAFQhSAH0AaWxWAAp/QJDwqJY4iQBKZQBvZXIG8gAzaHJ86PJWcgBeclI5TvIIvwCgqDDI6JjuCEsB1gBqBcryCR7yAHVWFIBRYQ4WfIk0TpPSSo9GygB2VY8AX2UvHyKxibLY+JhElPSsgCEuAAZQqjEqTcqrapKOpKK4tKBtDpdXr9QbDV7gj7THhzTpLFZrXoSGhbXb7GCHVhnC71RrEW4SB5PJQAbQAst9khAACbafiCfKclK8/l8QQAeT4bAghhoAEEAOaK2C4IwSGC6QQGAyKyoAXWeI2K41KqTiZOSexgBwwEXNoJN73NqzBUBOyKYPJgCBOqVW2EEDBgKwA7skRDByP6QORmPhsExsIqYDzanSGpjTZEXaqGIIMOsNSbLo1YE1o+QYBReuXyABVQzYJqceW4CJZ50xP0xfJ1vMF2lwxqN3DN1s8nl9mAV6ser0+lbEwPB0sNBnkAcYB5AA)

It seems the check isn't elided in BCL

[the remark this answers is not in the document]
WhiteBlackGoose: Indeed. But it looks more like a reason to PR to JIT than to replace with something which will be better for a short time. Have you filed an issue on that thing?

src/Maths/Silk.NET.Maths/Simd/Simd128.Abs.cs
WhiteBlackGoose: Actually, as for consistency, wouldn't it make sense to leave it as it is? @HurricanKai . I can do `And` too, but that will require much more time as I will need to PR `Add` for Scalar before

src/Maths/Silk.NET.Maths/Simd/Simd.GetWithElement.cs
```diff
@@ -0,0 +1,91 @@
...
+    /// <summary>
+    /// Provides several hardware-accelerated methods
+    /// for 128-bit vectors
+    /// </summary>
+    public static unsafe partial class Simd128
+    {
+        // these two should ONLY be used in the constant-sized unrolled loops.
+        // unfortunately there is an issue otherwise. See https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAJI8ovLrl5hcAbho1iAZiakGAYRoBvGg11MUDACoxcGABQBKHXu3U99hgDNoDM+IwNeDALwM6MzwYAHgYANRgwDGhyUgAOIMcAGwhsDAA+FjUIbgwA3gBqfKs7B1KAOQgAeQAHNws5Et0AX2tdVv0GCpq3Hk9i+1tS3V5HVzMOdwtAtN8AegAqccn52fDI6LiE5NSMrJyp2dnguDgGbkh8AV6wPAx2oaZyAE5LBvsW6iagA
+        [MethodImpl(Scalar.MaxOpt)]
+        private static Vector128<T> WithElement<T>(Vector128<T> vector, int index, T value) where T : unmanaged
```
Perksey, line 25: If we keep the manually written versions of these methods can we validate `index` please? This method is actually unsafe otherwise, rather than just (ab)using unsafe.

WhiteBlackGoose: Then won't it destroy the purpose of the method? Kai's [comment](https://github.com/dotnet/Silk.NET/pull/665#discussion_r741953852). Actually, I personally would prefer to remove manually written methods, because it's not a good idea to have them in long-run perspective.

So can we decide one and forever here? @HurricanKai

src/Maths/Silk.NET.Maths/Simd/Simd128.Abs.cs
WhiteBlackGoose: (`And` is added)

Perksey (changes requested):
Will put my big X here until #668 is merged.

src/Maths/Silk.NET.Maths/Simd/Simd.GetWithElement.cs
HurricanKai: I'm fine with removing them. We should for sure raise an issue on dotnet/runtime then, cause while not a huge deal and only used in fallback paths it'd be nice to have that anyways

src/Maths/Silk.NET.Maths/Simd/Simd128.Load.cs
HurricanKai: @Perksey this method doesn't read 4 Ts it reads Vector128<T>.Count Ts. All methods operate on that count

src/Maths/Silk.NET.Maths/Simd/Simd128.Constants.cs
```diff
@@ -115,6 +118,15 @@ static Simd128()
             Pi = Simd128.Create(Scalar<T>.Pi);
             PiOver2 = Simd128.Create(Scalar<T>.PiOver2);
             Tau = Simd128.Create(Scalar<T>.Tau);
+#if NET5_0_OR_GREATER
+            AllBitsSet = Vector128<T>.AllBitsSet;
+#else
+            AllBitsSet = Vector128<T>.Zero;
+            for (int i = 0; i < Vector128<T>.Count; i++)
+            {
+                Simd128.WithElement(AllBitsSet, i, Scalar.Not(default(T)));
```
HurricanKai, line 127: nit: Scalar<T>.Zero  > default(T)

src/Maths/Silk.NET.Maths/Simd/Simd128.Load.cs
Perksey: @HurricanKai ok fair enough.

src/Maths/Silk.NET.Maths/Simd/Simd64.Constants.cs
```diff
@@ -0,0 +1,133 @@
...
+
+#if INTRINSICS
+using System;
+using System.Runtime.CompilerServices;
+using System.Runtime.Intrinsics;
+
+namespace Silk.NET.Maths
+{
+    public static class Simd64<T> where T : unmanaged
+    {
```
HurricanKai, line 12: Shouldn't there be an IsHardwareAccelerated somewhere in this file?

src/Maths/Silk.NET.Maths/Simd/Simd64.Not.cs
```diff
@@ -0,0 +1,139 @@
...
+
+namespace Silk.NET.Maths
+{
+    public static unsafe partial class Simd64
+    {
+        /// <summary>
+        /// Performs hardware-accelerated Not on 64-bit vectors.
+        /// </summary>
+        [MethodImpl(Scalar.MaxOpt)]
+        public static Vector64<T> Not<T>(Vector64<T> vector) where T : unmanaged
```
HurricanKai, line 23: Seems like we can make this whole method
```cs
if (Simd64.IsHardwareAccelerated) return Simd64.Xor(vector, Simd64<T>.AllBitsSet);
return Other(vec);
```

src/Maths/Silk.NET.Maths/Simd/Simd.GetWithElement.cs
Perksey: I personally am not keen on not validating user input (segfaults are ugly), but can let this slide if Kai wants.

src/Maths/Silk.NET.Maths/Simd/Simd128.Not.cs
```diff
@@ -0,0 +1,139 @@
...
+        [MethodImpl(Scalar.MaxOpt)]
+        public static Vector128<T> Not<T>(Vector128<T> vector) where T : unmanaged
+        {
+            return Byte(vector);            
+            [MethodImpl(Scalar.MaxOpt)]
+            static Vector128<T> Byte(Vector128<T> vector)
+            {
+                if (typeof(T) == typeof(byte))
+                {
+                    return Simd128.Xor(vector, Simd128<T>.AllBitsSet);
```
HurricanKai, line 31: Same as in Simd64.Not

src/Maths/Silk.NET.Maths/Simd/Simd256.Not.cs
```diff
@@ -0,0 +1,139 @@
...
+        [MethodImpl(Scalar.MaxOpt)]
+        public static Vector256<T> Not<T>(Vector256<T> vector) where T : unmanaged
+        {
+            return Byte(vector);            
+            [MethodImpl(Scalar.MaxOpt)]
+            static Vector256<T> Byte(Vector256<T> vector)
+            {
+                if (typeof(T) == typeof(byte))
+                {
+                    return Simd256.Xor(vector, Simd256<T>.AllBitsSet);
```
HurricanKai, line 31: Same as in Simd64.Not

src/Maths/Silk.NET.Maths/Simd/Simd.GetWithElement.cs
```diff
@@ -0,0 +1,91 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
HurricanKai, line 1: Just putting this here so I don't forget whether we went through with this change

HurricanKai (changes requested):

src/Maths/Silk.NET.Maths/Simd/Simd64.Constants.cs
WhiteBlackGoose: @HurricanKai that's what I wanted to ask about. I didn't get what this method is supposed to do

src/Maths/Silk.NET.Maths/Simd/Simd64.Not.cs
WhiteBlackGoose: ._.

src/Maths/Silk.NET.Maths/Simd/Simd64.Constants.cs
HurricanKai: Basically it's just a shorthand for "is one of the ISAs we use available" so usually `Sse(2).IsSupported || AdvSimd.IsSupported`

WhiteBlackGoose: Got it

HurricanKai (approved):
I approve and thank you for your huge contribution!
@Perksey please remove your X

Perksey (approved):
eggs benedict

PR #668, opened 2021-11-03
Minor proposal changes

WhiteBlackGoose:
# Summary of the PR
Updates to proposals

### DIFF
--- a/documentation/proposals/Proposal - Bitwise - Scalar.md
+++ b/documentation/proposals/Proposal - Bitwise - Scalar.md
@@ -9,7 +9,7 @@ Bitwise operations in `Scalar[<T>]`
 - [x] Proposed
 - [x] Discussed with API Review Board (ARB)
 - [x] Approved
-- [ ] Implemented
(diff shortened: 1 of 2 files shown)


Perksey:
Needs working group sign-off, added to list for next meeting.

WhiteBlackGoose:
🙄 when is it going to be? `Not` is not going to work without `AllBitsSet` property

Perksey:
There's no date at the moment, and there's still feedback from the last working group meeting we need to address. Will try to work on that this evening/this weekend to further it along.

WhiteBlackGoose:
._. sounds beurocratic to me.

Now that means that my whole PR which implements `AllBitsSets` will be frozen for a while. Maybe I can remove `Not` and `AllBitsSet` from it though...

Perksey (approved):
Actually on second glance this seems fine and is a continuation of the already-approved message of "like Scalar but with Vectors". The main thing I was concerned about in the details of the bitshifts, but it looks like there were addressed in the last meeting. Approving

PR #673, opened 2021-11-05
Remove problematic implicit in overloads

Perksey:
Does what it says on the tin.


HurricanKai (changes requested):
It seems the span overloader relies on the `in` overloader to then generate .... `in span.GetPinnableReference()` 
It should instead just use fixed and pass the pointer.

Perksey:
Ok but a Span is invalid here too?

HurricanKai:
Yeah, I agree. But some of these methods seem like a ref is reasonable?

Perksey:
Nope, none of them :D All of these refer to pointers in the vertex layout, not actual CPU memory pointers.

HurricanKai (approved):
I don't know enough about OpenGL to disagree so LGTM

PR #674, opened 2021-11-06
Silk.NET backed Skia

HurricanKai:
Just a basic Silk.NET backed Skia demo on OpenGL. We could consider adding this as a demo, instead of a Lab app.
@Perksey, thoughts?


Perksey:
Yeah making it an official demo seems fine, will review it accordingly though.

src/Lab/Experiments/SilkBackedSkiaRendering/Program.cs
```diff
@@ -0,0 +1,32 @@
+﻿using Silk.NET.Maths;
+using Silk.NET.Windowing;
+using Silk.NET.Windowing.Glfw;
+using SkiaSharp;
+
+var options = WindowOptions.Default;
```
Perksey, line 6: Should we start using `with` in our tutorials?

src/Lab/Experiments/SilkBackedSkiaRendering/Program.cs
```diff
@@ -0,0 +1,32 @@
...
+options.Size = new Vector2D<int>(800, 600);
+options.Title = "Silk.NET backed Skia rendering!";
+options.PreferredStencilBufferBits = 8;
+options.PreferredBitDepth = new Vector4D<int>(8, 8, 8, 8);
+GlfwWindowing.Use();
```
Perksey, line 11: Why?

src/Lab/Experiments/SilkBackedSkiaRendering/Program.cs
```diff
@@ -0,0 +1,32 @@
...
+var window = Window.Create(options);
```
Perksey, line 12: Could do with some whitespace before this line for readability

src/Lab/Experiments/SilkBackedSkiaRendering/Program.cs
```diff
@@ -0,0 +1,32 @@
...
+window.Initialize();
```
Perksey, line 13: Why do we need to use Initialize? We should encourage the use of Load callbacks, this function is only intended for niche use cases.

src/Lab/Experiments/SilkBackedSkiaRendering/Program.cs
```diff
@@ -0,0 +1,32 @@
...
+using var grGlInterface = GRGlInterface.Create((name => window.GLContext!.TryGetProcAddress(name, out var addr) ? addr : (IntPtr) 0));
```
Perksey, line 14: Can you add a comment on what this does.

Perksey, line 14: Why do we need the IntPtr cast?

src/Lab/Experiments/SilkBackedSkiaRendering/Program.cs
```diff
@@ -0,0 +1,32 @@
...
+grGlInterface.Validate();
+using var grContext = GRContext.CreateGl(grGlInterface);
+var renderTarget = new GRBackendRenderTarget(800, 600, 0, 8, new GRGlFramebufferInfo(0, 0x8058)); // 0x8058 = GL_RGBA8`
```
Perksey, line 17: What are these values? Can you add a comment.

Perksey, line 17: Can some of these numbers be grabbed from WindowOptions to prevent duplication (i.e. you only need to change the size in one place)

Perksey, line 17: What happens if you resize the window? These seem like size-dependent values, so do we need to recreate this in a FramebufferResize callback?

src/Lab/Experiments/SilkBackedSkiaRendering/Program.cs
```diff
@@ -0,0 +1,32 @@
...
+using var surface = SKSurface.Create(grContext, renderTarget, GRSurfaceOrigin.BottomLeft, SKColorType.Rgba8888);
+using var canvas = surface.Canvas;
+
+window.Render += d =>
```
Perksey, line 21: Use `delta` instead of `d` for explicitness.

src/Lab/Experiments/SilkBackedSkiaRendering/SilkBackedSkiaRendering.csproj
```diff
@@ -0,0 +1,20 @@
...
+        <OutputType>Exe</OutputType>
+        <TargetFramework>net6.0</TargetFramework>
+        <ImplicitUsings>enable</ImplicitUsings>
+        <Nullable>enable</Nullable>
+        <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <ProjectReference Include="..\..\..\Windowing\Silk.NET.Windowing.Common\Silk.NET.Windowing.Common.csproj" />
+      <ProjectReference Include="..\..\..\Windowing\Silk.NET.Windowing.Glfw\Silk.NET.Windowing.Glfw.csproj" />
```
Perksey, line 13: Why are you pulling in just Glfw? Why not the Windowing metapacakge? (would eliminate the .Common reference as well)

Perksey (changes requested):

Perksey:
If you don't want to graduate this from the Lab you can discard all of the above comments, just want to ensure we keep a high standard for our tutorials.

src/Lab/Experiments/SilkBackedSkiaRendering/SilkBackedSkiaRendering.csproj
HurricanKai: Because this only works with GLFW. Will see if I can also make it compatible with SDL

src/Lab/Experiments/SilkBackedSkiaRendering/Program.cs
HurricanKai: Is _ also fine?

src/Lab/Experiments/SilkBackedSkiaRendering/Program.cs
HurricanKai: Yes. Will do

src/Lab/Experiments/SilkBackedSkiaRendering/Program.cs
Perksey: I mean yes, but it might be unclear what we're discarding especially given this is a minimal example. So probably worth naming the variable anyway.

src/Lab/Experiments/SilkBackedSkiaRendering/SilkBackedSkiaRendering.csproj
Perksey: There's no reason that I know why this wouldn't work with either, so give it a try :) if it doesn't work with both we should probably keep this in the Lab.

Perksey (changes requested):
Not sure if you intended to request another review, but my previous comments still stand.

HurricanKai:
I have no idea why, but it automatically requested a review from the codeowners :thinking:

Perksey:
Oh just noticed you moved this into the lab. Happy to merge once conflicts are resolved.

Perksey (approved):
Fixed conflicts myself as the stale PR was making me sad, happy to merge into lab only. Would love to upgrade it to a demo at some point, though.

HurricanKai:
Just me or is GitHub refusing to merge this without a reason?

Perksey:
Nah, the conversations must be resolved. It's a new requirement they've added, and I enabled it as I figured it'd make us look at all the comments again to make sure they're actually resolved. I'll resolve them real quick as this is the lab and we don't care.

PR #670, opened 2021-11-05
SIMD for Silk.NET.Math #2

WhiteBlackGoose:
## Summary of the PR

Keep working on #666

## Tasks

- [x] `Equal`
- [x] `NotEqual`
- [x] `GreaterThan`
- [x] `GreaterThanOrEqual`
- [x] `LessThan`
- [x] `LessThanOrEqual`
- [x] `Min`
- [x] `Max`

## Benchmarks

For all bitnesses x for all types

- [x] `GreaterThan`, `LessThan`, `GreaterThanOrEqual`, `LessThanOrEqual`
- [x] 4 arithmetic operators
- [x] `Min`, `Max`
- [x] `Equal`, `NotEqual`

## Tests

For all bitnesses x for all types

- [x] 4 arithmetic operators
- [x] `Less`/`Greater` `Than` (`OrEqual`)
- [x] `Min`, `Max`
- [x] (`Not`)`Equal`
- [x] `Bitwise`
- [x] `Abs`

## Other

- [x] Use naive operations for 8-byte values on simd64

## What is NOT done
`long` and `ulong` are not supported by majority of methods (added to TODO of #666)

~~`*OrEqual` don't support most types (so I replaced it with `Not(...Than` for types that aren't supported)~~ done in the best way

~~128-bit fallbacks when 256-bit is not supported~~ done where appropriate

PR #678, opened 2021-11-08
[Proposal] AllBitsSet for Scalar

WhiteBlackGoose:
Similarly to .NET 5's Vector64<T>.AllBitsSet (and other bitnesses) I suggest adding `AllBitsSet` public readonly field to `Scalar<T>`.

It is already implemented as `internal` field in #670 (so I propose making it public)

### DIFF
--- a/documentation/proposals/Proposal - Bitwise - Scalar.md
+++ b/documentation/proposals/Proposal - Bitwise - Scalar.md
@@ -20,6 +20,8 @@ To make this simpler for both me, and anyone who reads this, I've provided the A
 This same API would also be amended to the Vectorization/SIMD proposal, if both are accepted.
 
 ```cs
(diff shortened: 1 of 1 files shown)


WhiteBlackGoose:
Um, idk if it's a good idea to add it to an existing proposal...

Perksey:
I think this is small enough to just go in a PR directly. In the past we've modified historical proposals, but we all agreed in the 1.0 Preview era that this isn't really the purpose of them and just resulted in too much maintenance for the little things.

> (you won't need to make a proposal for just one function)

From the (somewhat outdated) proposals README

WhiteBlackGoose:
I interpreted it as there's no need to create a new proposal document (so better to be added to old ones). 😄 

>  I think this is small enough to just go in a PR directly

So I can just make it public, no other actions needed?

PR #685, opened 2021-11-10
Typo fixed

WhiteBlackGoose:
🦆


WhiteBlackGoose:
Or is it not a typo? But vector is fixed-size, not length, so it won't work

HurricanKai:
It's intentional, you only want the sign in those, so aligning with BCL int makes sense.
Could consider having a copy sign method or changing it though

WhiteBlackGoose:
[Discussion in discord](https://discord.com/channels/521092042781229087/587346162802229298/907982387802173450)

HurricanKai (approved):

HurricanKai (changes requested):
This looks good to me, but I apparently can't make it wait for @Perksey s review.
Can we change this as is or do we need to take this back for review?

Perksey (approved):
Nothing functional or logical is changing, good for merge.

HurricanKai (approved):

PR #691, opened 2021-11-14
No need to run CI on source-unrelated files

WhiteBlackGoose:
I intentionally only wrote down just two: those are assumed to change often. We could add all files which don't affect the build to it, but imo it makes maintaining this thing harder.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,7 +5,7 @@ on:
       - 'main'
       - 'release/*'
     paths-ignore:
-      - 'proposals/**'
+      - 'documentation/**'
       - 'README.md'
   pull_request:
     # all branches
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,8 +4,14 @@ on:
     branches:
       - 'main'
       - 'release/*'
+    paths-ignore:
+      - 'proposals/**'
+      - 'README.md'
   pull_request:
-    # none
+    # all branches
+    paths-ignore:
+      - 'proposals/**'
+      - 'README.md'
 jobs:
   Build:
     runs-on: windows-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -4,8 +4,14 @@ on:
     branches:
       - 'main'
       - 'release/*'
+    paths-ignore:
+      - 'proposals/**'
+      - 'README.md'
```
Perksey, line 9: The README does in fact change the pack output. Probably best to run CI for this.

.github/workflows/build.yml
```diff
@@ -4,8 +4,14 @@ on:
     branches:
       - 'main'
       - 'release/*'
+    paths-ignore:
+      - 'proposals/**'
```
Perksey, line 8:
```suggestion
      - 'documentation/**'
```

.github/workflows/build.yml
```diff
@@ -4,8 +4,14 @@ on:
...
+      - 'README.md'
   pull_request:
-    # none
+    # all branches
+    paths-ignore:
+      - 'proposals/**'
```
Perksey, line 13:
```suggestion
      - 'documentation/**'
```

Perksey (changes requested):

.github/workflows/build.yml
WhiteBlackGoose: You will, of course, ask, *how* on Earth was I able to make this mistake

.github/workflows/build.yml
WhiteBlackGoose: Oh... um, does it really affect anything?

Perksey: Nothing we check for, but the README is bundled in every NuGet package and is fed into MSBuild (see [here](https://github.com/dotnet/Silk.NET/blob/main/build/props/common.props#L46))

Perksey (approved):

PR #696, opened 2021-11-17
Codegen tests (experimental)

WhiteBlackGoose:
Working on tests verifying the correct JIT codegen


HurricanKai:
Just a heads up, I might end up throwing the majority of this out with the new test suite.

WhiteBlackGoose:
Could you elaborate?

WhiteBlackGoose:
For now I keep integer types out of the tests, because they give bad codegen in a lot of cases, as well as all 64-bit ops. (out of all 750 tests, 239 failed).

I'm opening this as ready for review, but there's no need to rush with review, it's not blocking me.

Also if you don't like failing tests, I'll think what we can do about them

HurricanKai:
We should probably port something like [this](https://github.com/crummel/dotnet_arcade/blob/master/src/Microsoft.DotNet.XUnitExtensions/src/Attributes/ActiveIssueAttribute.cs) to our repo, because Failing tests shouldn't really be merged every, just skipped if the work is put of.
I'll think about how to do this, won't merge for a while

Perksey:
+1 I think the current Silk ATG activity has proven we need this!

WhiteBlackGoose:
Nice! The lib's almost released (I can release it any moment, just waiting if some new bug appears)

Perksey:
Closing draft PR due to inactivity. This is useful, but there's no work happening here right now. GitHub should preserve the PR patch should we pursue this again.

PR #703, opened 2021-11-25
[Experimental] CodegenAnalysis benchmarks

WhiteBlackGoose:
Report that they generate can be found [**here**](https://gist.github.com/WhiteBlackGoose/e452ea5351a73a698b21aedc44441593) (warning: *big* file).

(there's a markdown table (it's not well fitted, you might wanna scroll right), and then there's a list of codegens for each benched function (note that in benchmarks those are loops!))

@HurricanKai what do you think?

### DIFF
--- a/src/Maths/Silk.NET.Maths.Benchmarks/Program.cs
+++ b/src/Maths/Silk.NET.Maths.Benchmarks/Program.cs
@@ -3,6 +3,7 @@
 
 using System.Reflection;
 using BenchmarkDotNet.Running;
+using CodegenAnalysis.Benchmarks;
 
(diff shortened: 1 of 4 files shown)


Perksey:
Note: all dependencies added pass the license check.

WhiteBlackGoose:
What's that? 😅 
The lib's under MIT

Perksey:
Yes I know, that’s why I said it *passed* the license check. All other dependencies added are also MIT or MIT-compatible.

Perksey:
Closing draft PR due to inactivity.

PR #704, opened 2021-11-25
start of adding window class option

Pyrdacor:
# Summary of the PR

Adds the possibility to specify an optional window class for a window. Systems like X11 use the window class to store settings for specific applications. Glfw defaults the window class to the window title which is not always wanted. Especially with version numbers in the title this causes problems as a window class should stay the same for the same application.

### DIFF
--- a/src/Windowing/Silk.NET.Windowing.Common/Interfaces/IWindowProperties.cs
+++ b/src/Windowing/Silk.NET.Windowing.Common/Interfaces/IWindowProperties.cs
@@ -66,11 +66,12 @@ namespace Silk.NET.Windowing
         /// <summary>
         /// Window class name used in systems like X11.
         /// Also known as WM_CLASS.
-        /// 
+        /// </summary>
+        /// <remarks>
         /// If omitted the following default values are used:
         /// - The name of the main assembly without extension
         /// - "SilkNET_App" if the name of the main assembly could not be determined
-        /// </summary>
+        /// </remarks>
         string? WindowClass { get; }
     }
 }
--- a/src/Windowing/Silk.NET.Windowing.Common/Interfaces/IWindowProperties.cs
+++ b/src/Windowing/Silk.NET.Windowing.Common/Interfaces/IWindowProperties.cs
@@ -70,7 +70,7 @@ namespace Silk.NET.Windowing
         /// <remarks>
         /// If omitted the following default values are used:
         /// - The name of the main assembly without extension
-        /// - "SilkNET_App" if the name of the main assembly could not be determined
+        /// - "Silk.NET" if the name of the main assembly could not be determined
         /// </remarks>
         string? WindowClass { get; }
     }
--- a/src/Windowing/Silk.NET.Windowing.Common/Interfaces/IWindowProperties.cs
+++ b/src/Windowing/Silk.NET.Windowing.Common/Interfaces/IWindowProperties.cs
@@ -62,5 +62,15 @@ namespace Silk.NET.Windowing
         /// The context with which this window's context's resources are shared.
         /// </summary>
         IGLContext? SharedContext { get; }
+
+        /// <summary>
+        /// Window class name used in systems like X11.
+        /// Also known as WM_CLASS.
+        /// 
+        /// If omitted the following default values are used:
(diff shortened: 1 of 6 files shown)


Pyrdacor:
@Perksey Please check if this is ok this way.

Pyrdacor:
@Perksey I changed the default window classes. Hope it's ok this way. I have two fallbacks and I prefer current processes MainModule as it will also work with 3.1 self-contained exes where the entry assembly might be "dotnet.exe". At least it should.

src/Windowing/Silk.NET.Windowing.Common/Window.cs
```diff
@@ -19,12 +19,25 @@ public static class Window
...
+        internal static string DefaultWindowClass { get; }
+
+        static Window()
+        {
+            string? defaultWindowClassName = Process.GetCurrentProcess().MainModule?.ModuleName;
+
+            if (defaultWindowClassName != null)
+                DefaultWindowClass = Path.GetFileNameWithoutExtension(defaultWindowClassName);
+            else
+                DefaultWindowClass = Assembly.GetEntryAssembly()?.GetName()?.Name ?? FallbackWindowClass;
```
Perksey, line 38:
```suggestion
            var defaultWindowClassName = Process.GetCurrentProcess().MainModule?.ModuleName;
            if (defaultWindowClassName is not null)
            {
                DefaultWindowClass = Path.GetFileNameWithoutExtension(defaultWindowClassName);
                return;
            }

            DefaultWindowClass = Assembly.GetEntryAssembly()?.GetName()?.Name ?? FallbackWindowClass;
```

src/Windowing/Silk.NET.Windowing.Common/Window.cs
```diff
@@ -19,12 +19,25 @@ public static class Window
         private const string SdlBackendNamespace = "Silk.NET.Windowing.Sdl";
         private const string GlfwBackendName = "GlfwPlatform";
         private const string SdlBackendName = "SdlPlatform";
+        private const string FallbackWindowClass = "SilkNET_App";
```
Perksey, line 22: Why `_App`? Why can't it just be `Silk.NET` or `silkdotnet`? Is there some naming convention I'm missing?

src/Windowing/Silk.NET.Windowing.Common/Interfaces/IWindowProperties.cs
```diff
@@ -62,5 +62,15 @@ public interface IWindowProperties : IViewProperties
...
         IGLContext? SharedContext { get; }
+
+        /// <summary>
+        /// Window class name used in systems like X11.
+        /// Also known as WM_CLASS.
+        /// 
+        /// If omitted the following default values are used:
+        /// - The name of the main assembly without extension
+        /// - "SilkNET_App" if the name of the main assembly could not be determined
+        /// </summary>
```
Perksey, line 73:
```suggestion
        /// </summary>
        /// <remarks>
        /// If omitted the following default values are used:
        /// - The name of the main assembly without extension
        /// - "SilkNET_App" if the name of the main assembly could not be determined
        /// </remarks>
```

src/Windowing/Silk.NET.Windowing.Common/Window.cs
Pyrdacor: I haven't found much about naming conventions or restrictions. Only control characters are not allowed. SDL uses "SDL_App". But we can just use a name without it. I think the '.' is ok but I generally avoid all characters beside letters, digits and underscores when I deal with names that are processed by different tools and systems. This makes it more robust imo. Can you change the name to something you prefer please?

Perksey: Looking at the documentation it seems like a dot should be allowed, so let's go with that I think. If it proves problematic we can change later.

```suggestion
        private const string FallbackWindowClass = "Silk.NET";
```

Pyrdacor:
I fixed the doc comment about the default name from "SilkNET_App" to "Silk.NET" as well. Should be ok now.

Perksey (approved):

PR #708, opened 2021-12-03
Fix website build pipeline

Perksey:
Apparently the CI token can't push anymore so let's use the same one the rewrite merger pipeline uses...


HurricanKai (approved):
That is dirty, but sure

PR #711, opened 2021-12-04
Add support for cloning Vk objects to reuse their function tables independently

Perksey:
Popped up while looking at the Stride migration, which has a graphics abstraction that can create multiple devices but only ever has one instance: let's just reuse the instance functions we've already loaded!

Considering our (my) advice is to have one object for each instance-device combination, this is quite a nice-to-have but at the same time a niche one so if you're not comfortable with it that's fine.


HurricanKai (approved):
I don't mind having this

PR #735, opened 2021-12-27
Add Flags attribute to bitmasks on the clang backend

### LINKED ISSUE #729: Adding FlagsAttribute to Silk.NET.Direct3D11.BindFlag
We will use `Silk.NET.Direct3D11.BindFlag` as the flag.

```csharp
            var texture2DDesc = new D3D11.Texture2DDesc()
            {
                BindFlags = (uint)(D3D11.BindFlag.BindRenderTarget | D3D11.BindFlag.BindShaderResource),
            };
```

ReSharper will tell me "Bitwise Operator On Enum Without Flags".

Perksey:
Does what it says on the tin. There's no way to cleanly determine this, so we just say if 50% of the enums are a bit then it's a bitmask.

Closes #729


src/Core/Silk.NET.BuildTools/Cpp/Clang.cs
```diff
@@ -48,6 +48,12 @@ public static class Clang
             "LONG_PTR"
         };
 
+        public static bool IsProbablyABitmask(Enum @enum)
+            => @enum.Tokens.Count > 1 && // there is more than one token
```
HurricanKai, line 52: This condition doesn't need to exist?

HurricanKai (commented):
Can we have [Flags] instead of [Flags()]?

src/Core/Silk.NET.BuildTools/Cpp/Clang.cs
Perksey: Not technically, but logically it does need to exist. An enum containing a single token of value 1 would be identified as a bitmask, but logically a bit mask requires more than one bit. 

A single token having just a bit enough is not enough certainty that we have a bitmask on our hands.

Perksey:
Probably, will just need to modify the code elsewhere. Obviously being NYE won't do this right now.

Perksey:
Review comments rectified.

HurricanKai (approved):

PR #746, opened 2022-01-04
ANGLE, SwiftShader, and Vulkan Loader prebuilts

Perksey:


Perksey:
Ready to review. ANGLE on Windows was built locally because GitHub Actions doesn't have enough disk space.

This does not need to be in 2.12.

.nuke
```diff
@@ -1 +0,0 @@
-Silk.NET.sln
```
HurricanKai, line 1: I don't know what this does just putting this message to give you a heads up that this happened

.nuke/build.schema.json
```diff
@@ -0,0 +1,185 @@
...
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "Build Schema",
+  "$ref": "#/definitions/build",
+  "definitions": {
+    "build": {
+      "type": "object",
+      "properties": {
+        "Configuration": {
+          "type": "string",
+          "description": "Configuration to build - Default is 'Debug' (local) or 'Release' (server)"
```
HurricanKai, line 11: nit
```suggestion
          "description": "Configuration to build - Default is 'Debug'. Possible Values: 'Debug' (local) or 'Release' (server)"
```

.nuke/build.schema.json
```diff
@@ -0,0 +1,185 @@
...
+        },
+        "Help": {
+          "type": "boolean",
+          "description": "Shows the help text for this build assembly"
+        },
+        "Host": {
+          "type": "string",
+          "description": "Host for execution. Default is 'automatic'",
+          "enum": [
+            "AppVeyor",
```
HurricanKai, line 25:
```suggestion
            "automatic",
            "AppVeyor",
```

HurricanKai (commented):
Overall looks good!
Not remembering why we need to have a Vulkan native package though, IIRC that should never be needed anyways (double-checking that right now)

Perksey:
The Vulkan loader package is primarily for drivers which aren't integrated to the OS, i.e. MoltenVK and SwiftShader (the most pertinent! originally needed this because of Jupyter and the custom layers it needs for headless surfaces)

.nuke/build.schema.json
Perksey: This is autogenerated, recommend raising this upstream at NUKE.

.nuke/build.schema.json
HurricanKai: This originates [here](https://github.com/dotnet/Silk.NET/blob/073e45d090fc59f045d2e606fec6b9f18e8f48b4/build/nuke/Build.Support.cs#L76) but I can't be bothered to open up the solution just to add 3 words very few people will ever see  and even fewer people will misunderstand...

HurricanKai:
For future reference I've asked about the ins-and-outs of shipping the vulkan loader on the [VK discord](discord.gg/vulkan) and this seems reasonable ([link to convo](https://discord.com/channels/427551838099996672/427833316536746005/949786825306820628))

HurricanKai (approved):

HurricanKai:
Please have a look at the build error, but given it's nothing related to this PR this is good to go.

.nuke
Perksey: Yep, NUKE upgrade did this.

Perksey:
Action items post-merge and post-release:
- [ ] Install Windows SDK in Actions (will fix ANGLE Windows Build)
- [ ] Ignore cases where the binaries don't change (will fix Linux Builds)

PR #750, opened 2022-01-05
Fix a small divide by zero error in PointToFramebuffer

Perksey:
Noticed this when adding a Silk.NET backend to Trains.NET.


HurricanKai (approved):
I mean that's ugly but sure

Perksey:
XD

PR #762, opened 2022-01-06
Update input context in DoEvents instead of DoUpdate

Perksey:
As discussed in Discord

### DIFF
--- a/src/Input/Silk.NET.Input.Glfw/GlfwInputContext.cs
+++ b/src/Input/Silk.NET.Input.Glfw/GlfwInputContext.cs
@@ -11,20 +11,19 @@ using Silk.NET.Windowing.Internals;
 
 namespace Silk.NET.Input.Glfw
 {
-    internal class GlfwInputContext : IInputContext
+    internal class GlfwInputContext : InputContextImplementationBase
     {
         private readonly GlfwGamepad[] _gamepads = new GlfwGamepad[16];
         private readonly GlfwJoystick[] _joysticks = new GlfwJoystick[16];
         private readonly GlfwKeyboard[] _keyboards = new GlfwKeyboard[1];
         private readonly GlfwMouse[] _mice = new GlfwMouse[1];
         private readonly IGlfwSubscriber[] _subscribers = new IGlfwSubscriber[2];
-        private IView _window;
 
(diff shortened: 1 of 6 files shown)


src/Input/Silk.NET.Input.Glfw/GlfwInputContext.cs
```diff
@@ -50,30 +50,48 @@ public unsafe GlfwInputContext(IView window)
...
         public unsafe void Dispose()
         {
-            _window.Update -= _update;
+            if (_window is ViewImplementationBase view)
+            {
+                view.ProcessEvents += ProcessEvents;
+            }
+            else
+            {
+                _window.Update += Update;
```
vdvman1, line 92: Shouldn't these be removing the event handlers rather than adding them again?

Perksey: Yes, thanks!

```suggestion
                _window.Update -= Update;
```

src/Input/Silk.NET.Input.Glfw/GlfwInputContext.cs
```diff
@@ -50,30 +50,48 @@ public unsafe GlfwInputContext(IView window)
...
+                updatable.Update();
+            }
+        }
+
         public unsafe void Dispose()
         {
-            _window.Update -= _update;
+            if (_window is ViewImplementationBase view)
+            {
+                view.ProcessEvents += ProcessEvents;
```
Perksey, line 88:
```suggestion
                view.ProcessEvents -= ProcessEvents;
```

ThomasMiz:
Shouldn't we change SDL to also use the new `ProcessEvents` event? It seems to basically be a generalized version of [this](https://github.com/dotnet/Silk.NET/blob/3a5f4508c62eaf6eb3d2466fb971ebc2b359ce12/src/Windowing/Silk.NET.Windowing.Sdl/SdlView.cs#L96) so I don't see why not.

Would be extremely confusing if `SdlView` had both the inherited `ProcessEvents` and a separate `ProcessingEvents` event

Perksey:
You're right, I've changed this to make the same event common (and make other code common as well)

ThomasMiz (approved):

PR #803, opened 2022-02-05
This is the February 2022 Update (v2.13.0)

Perksey:


Perksey:
This is ready for review. This is the first "This is" PR in a while to _just_ contain a bindings regeneration and patch notes, so this should be an easy review :D

HurricanKai (approved):

PR #818, opened 2022-02-20
Make the 3.0 history merger run less frequently

Perksey:


Perksey:
I've just deleted this workflow now. While it'd make our lives easier at release time, it doesn't make our lives easier during development time. On second thought, it's fine to let these branches diverge a little bit so long as we merge whenever we need to.

HurricanKai (approved):

PR #836, opened 2022-03-12
This is the Silk.NET March 2022 Update (v2.14)

### LINKED ISSUE #796: Silk.NET.Assimp Quaternions are incorrect
# Summary
Upon parsing Quaternions from `NodeAnim*`, I was stumbling upon incorrect Quaternions. The `aiQuaterniont` struct puts W first, but System.Numerics.Quaternion is X first. So it appears Silk.NET is casting those aiQuaternionT* to System.Numerics.Quaternion or something, which results in W being at X which is wrong.

Here's my workaround for parsing them:
```cs
NodeAnim* chan = a->MChannels[j];

var positions = new (float, Vector3)[chan->MNumPositionKeys];
for (int k = 0; k < positions.Length; k++)
{
    VectorKey pos = chan->MPositionKeys[k];
    positions[k] = ((float)pos.MTime, pos.MValue);
}
var rotations = new (float, Quaternion)[chan->MNumRotationKeys];
for (int k = 0; k < rotations.Length; k++)
{
    QuatKey rot = chan->MRotationKeys[k];
    rotations[k] = ((float)rot.MTime, new Quaternion(rot.MValue.Y, rot.MValue.Z, rot.MValue.W, rot.MValue.X)); // HERE: Incorrectly casted by library, so order is shifted by one field
}
var scales = new (float, Vector3)[chan->MNumScalingKeys];
for (int k = 0; k < scales.Length; k++)
{
    VectorKey scl = chan->MScalingKeys[k];
    scales[k] = ((float)scl.MTime, scl.MValue);
}
```

# Comments
- Platform: Desktop
- Framework Version: .NET 6.0
- API: OpenGL 3.3 Core

I'm not sure if this is unique to `NodeAnim` or if it applies all over the library, so it needs some looking into

### LINKED ISSUE #811: Silk.NET.Direct3D11.ID3D10Blob Missing Methods
Hi,

The Silk.NET.Direct3D11.ID3D10Blob structure appears to be missing these two methods:

- [ID3D10Blob::GetBufferPointer](https://docs.microsoft.com/en-us/windows/win32/api/d3dcommon/nf-d3dcommon-id3d10blob-getbufferpointer)
- [ID3D10Blob::GetBufferSize](https://docs.microsoft.com/en-us/windows/win32/api/d3dcommon/nf-d3dcommon-id3d10blob-getbuffersize)

The struct itself contains nothing:
image

EDIT: The workaround for this seems to be to use IDxcBlob for now, [as it is an alias of ID3D10Blob](https://github.com/microsoft/DirectXShaderCompiler/blob/9f60094d90bb2cc8b5997ad8a9b8ff777f8d7a0b/include/dxc/dxcapi.h#L148). However, I don't think this is a safe workaround given that Microsoft may change IDxcBlob structure/functionality later down the line and diverge from ID3D10Blob.

Perksey:
Closes #796 
Closes #811 

Still working on this, desc will update soon.


Perksey:
This is ready for review. Haven't given it a once over myself, will do that in a bit, but I doubt much will change now.

HurricanKai (approved):
LGTM

PR #877, opened 2022-04-07
Update prerequisites in Readme

roeyskoe:
# Summary of the PR
`Android-aot` workload no longer exists, so it should be removed.
Added `maccatalyst`, since it is also included in `.github/workflows/deploy.yml` file. Not 100% sure if it is actually required.
Updated recommended Visual Studio 2019 to 2022, because 2019 does not officially support .NET6.


roeyskoe:
Would it actually be better if it mentioned VS and Rider? Neither are actually required, but are there really any other options?

Perksey (approved):

PR #879, opened 2022-04-10
Extra tests for Struct emission

HurricanKai:
This adds a few extra tests to cover all the desirable properties of struct emission


Perksey (approved):
Oh yeah that looks real good that's the stuff

PR #880, opened 2022-04-10
Fields

HurricanKai:
This adds (struct) fields in both the symbol and emission layers. This does *not* include logic to include fields in a struct, it purely represents fields as symbols and is able to emit them into C# as root symbols


HurricanKai:
I am a bit unhappy about the field types. They right now just accept a `TypeSymbol` which is kind of wrong, since the name given there should be able to both represent an inbuild type (`int`) and externally defined type (`System.String`) and a type which has been emitted previously (-> these could be treated as "externally defined")
Right now all these goals can still be met, but it's kind of unclean (`new StructSymbol(new IdentifierSymbol("int"))`, `new StructSymbol(new IdentifierSymbol("string"))`).
Note that the type given to fields can *not* just be a string / identifier because this should also carry whether the type is a reference or value type (and other type-related info. Maybe we should just have `ExternalTypeReference`?) (-> using these references we could even build the dependency tree?)

Perksey (approved):
my dogs on my lap so i'll just assume this is fine

PR #881, opened 2022-04-13
Impelement OpenCL enum generation

Alexx999:
This should add some basic enum generation support for OpenCL aside from the umbrella CLEnum type.

The initial motivation is that some of the enum types have to be ulong-based causing OpenCL API calls to fail in funny ways.


src/OpenCL/Silk.NET.OpenCL/Enums/CLEnum.gen.cs
```diff
@@ -849,11 +854,13 @@ public enum CLEnum : int
         [NativeName("Name", "CL_KERNEL_LOCAL_SIZE_FOR_SUB_GROUP_COUNT")]
         KernelLocalSizeForSubGroupCount = 0x11B8,
         [NativeName("Name", "CL_INVALID_SPEC_ID")]
-        InvalidSpecID = unchecked((int)0xFFFFFFFFFFFFFFB9),
+        InvalidSpecID = unchecked((int) 0xFFFFFFFFFFFFFFB9),
         [NativeName("Name", "CL_MAX_SIZE_RESTRICTION_EXCEEDED")]
-        MaxSizeRestrictionExceeded = unchecked((int)0xFFFFFFFFFFFFFFB8),
+        MaxSizeRestrictionExceeded = unchecked((int) 0xFFFFFFFFFFFFFFB8),
+        [System.Obsolete("Deprecated in version  3.0")]
```
Perksey, line 860: Double space?

src/OpenCL/Silk.NET.OpenCL/Enums/IntelAdvancedMotionEstimationIntraSearchPredictionModesBufferChromaBlock.gen.cs
```diff
@@ -0,0 +1,24 @@
...
+
+using System;
+using Silk.NET.Core.Attributes;
+
+#pragma warning disable 1591
+
+namespace Silk.NET.OpenCL
+{
+    [NativeName("Name", "cl_intel_advanced_motion_estimation_intra_search_prediction_modes_buffer_chroma_block")]
+    public enum IntelAdvancedMotionEstimationIntraSearchPredictionModesBufferChromaBlock : int
```
Perksey, line 13: lmao this is hilarious

src/OpenCL/Silk.NET.OpenCL/Enums/LoaderLayers.gen.cs
```diff
@@ -7,10 +7,12 @@
 
 #pragma warning disable 1591
 
-namespace Silk.NET.OpenCL.Extensions.APPLE
+namespace Silk.NET.OpenCL
 {
-    [NativeName("Name", "GLenum")]
-    public enum APPLE : int
```
Perksey, line 13: This is a breaking change. While we're happy to break if the upstream specification dictates it, we generally always have ungrouped enums available thus if it's possible I'd prefer to either:
- have the grouped enum in addition to the ungrouped vendor-wide enum
- stop this change entirely

If 1 is not too much trouble ;)

src/OpenCL/Silk.NET.OpenCL/Enums/SemaphoreTypeKhr.gen.cs
```diff
@@ -7,10 +7,12 @@
 
 #pragma warning disable 1591
 
-namespace Silk.NET.OpenCL.Extensions.POCL
+namespace Silk.NET.OpenCL
 {
-    [NativeName("Name", "GLenum")]
-    public enum POCL : int
+    [NativeName("Name", "cl_semaphore_type_khr")]
+    public enum SemaphoreTypeKhr : int
```
Perksey, line 13: Another one here too.

Perksey (changes requested):

Perksey:
Thank you so much for this contribution! I've already commended your efforts plenty in Discord, but we are genuinely delighted to see community efforts being directed here! We cannot thank you enough.

I suppose one final comment, in addition to the nitpicks above, is: is there any way to make the GroupOverloader light up so that these enums are properly integrated to the rest of the library?

src/Core/Silk.NET.BuildTools/Converters/Readers/OpenCLReader.cs
```diff
@@ -930,16 +949,122 @@ public IEnumerable<Enum> ReadEnums(object obj, BindTask task)
...
+                    .Elements("feature")
+                    .Select(x => (x.Attribute("api")?.Value, x.Attribute("number")?.Value))
+                    .Distinct())
+                {
+                    var ret = new Enum
+                    {
+                        Name = group.Value.translatedName,
+                        NativeName = group.Key,
+                        Attributes = new List<Attribute>(),
+                        ExtensionName = "Core (Grouped)",
```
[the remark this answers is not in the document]
Alexx999: Basically, "(Grouped)" are separate enums. as opposed to the large root CLEnum

src/Core/Silk.NET.BuildTools/Converters/Readers/OpenCLReader.cs
```diff
@@ -930,16 +949,122 @@ public IEnumerable<Enum> ReadEnums(object obj, BindTask task)
...
+
+        private static string EnumName(string @enum)
+        {
+            return @enum switch
+            {
+                // typo in spec
+                "cl_kernel_arg_type_qualifer" => "cl_kernel_arg_type_qualifier",
+                _ => @enum,
+            };
+        }
```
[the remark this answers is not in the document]
Alexx999: I've submitted a PR to Khronos, so this wouldn't have to stay for long anyway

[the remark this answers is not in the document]
Alexx999: OK, I see what you're saying now, switched to the JSON-based renamer

src/OpenCL/Silk.NET.OpenCL/Enums/LoaderLayers.gen.cs
Alexx999: Yeah, found the issue, those empty enums should be back in business now

src/OpenCL/Silk.NET.OpenCL/Enums/CLEnum.gen.cs
Alexx999: Good catch

Alexx999:
Actually I have 2 more PRs pretty much ready - these changes don't actually achieve my initial goal (namely, incorrect stack size for some calls due to incorrectly sized stuff), I just don't want to dump too much code at once
Next up, constants handling and finally a typemapper change that builds on top of these 2 to achieve correct codegen (or, at least, "more correct").

src/Core/Silk.NET.BuildTools/Converters/Readers/OpenCLReader.cs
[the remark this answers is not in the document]
Perksey: Gotcha, thanks.

src/Core/Silk.NET.BuildTools/Converters/Readers/OpenCLReader.cs
```diff
@@ -930,19 +945,117 @@ public IEnumerable<Enum> ReadEnums(object obj, BindTask task)
...
                 }
             }
+
+
+            var groups = new Dictionary<string, (List<Token> list, bool is64bit, string translatedName)>();
+            foreach (var @enum in allEnums)
+            {
+                var group = @enum.Value.Item2;
+
+                var is64bit = ulongEnums.Contains(group);
```
Perksey, line 955: Can we use this data to add a `[Flags]` attribute to the relevant enums? Otherwise you get compiler warnings when doing bitwise stuff.

Perksey (changes requested):

src/Core/Silk.NET.BuildTools/Converters/Readers/OpenCLReader.cs
Alexx999: Yeah, I was also thinking about this.
I wonder what the process should be for larger PRs like this: keep adding stuff until it's all good?

Perksey: Generally we do prefer smaller PRs, but in the world of BuildTools this is still only a small PR so your call!

src/Core/Silk.NET.BuildTools/Converters/Readers/OpenCLReader.cs
```diff
@@ -835,42 +755,181 @@ public IEnumerable<Enum> ReadEnums(object obj, BindTask task)
...
                 .Elements("enum")
-                .DistinctBy(x => x.Attribute("name")?.Value)
                 .Where
                 (
-                    x => x.Parent?.Attribute("name")?.Value != "Constants" &&
-                         x.Parent?.Attribute("name")?.Value != "MiscNumbers"
+                    x => !x.Parent.Attribute("name").Value.StartsWith("Constants") &&
+                         !x.Parent.Attribute("name").Value.StartsWith("enums") &&
+                         x.Parent.Attribute("name").Value != "MiscNumbers" &&
+                         x.Parent.Attribute("name").Value != "cl_device_info"
```
Perksey, line 808: Can you add a comment explaining why these can't make the first pass?

src/Core/Silk.NET.BuildTools/Converters/Readers/OpenCLReader.cs
```diff
@@ -835,42 +755,181 @@ public IEnumerable<Enum> ReadEnums(object obj, BindTask task)
...
+                        else
+                        {
+                            typeStr = name.Renamed switch
+                            {
+                                "CL_CONTEXT_MEMORY_INITIALIZE_KHR" => "cl_context_properties",
+                                "CL_DEVICE_PARTITION_BY_NAMES_INTEL" => "cl_device_partition_property",
+                                _ => typeStr
+                            };
+
+                            if (typeStr?.Contains("cl_device_info ") == true) typeStr = "cl_device_info";
```
Perksey, line 863: Can this not be hardcoded

src/Core/Silk.NET.BuildTools/Converters/Readers/OpenCLReader.cs
```diff
@@ -835,42 +755,181 @@ public IEnumerable<Enum> ReadEnums(object obj, BindTask task)
...
+                            type = pass1type;
+                        }
+                        else
+                        {
+                            typeStr = name.Renamed switch
+                            {
+                                "CL_CONTEXT_MEMORY_INITIALIZE_KHR" => "cl_context_properties",
+                                "CL_DEVICE_PARTITION_BY_NAMES_INTEL" => "cl_device_partition_property",
+                                _ => typeStr
+                            };
```
Perksey, line 861: Can these not be hardcoded

src/Core/Silk.NET.BuildTools/Converters/Readers/OpenCLReader.cs
```diff
@@ -835,42 +755,181 @@ public IEnumerable<Enum> ReadEnums(object obj, BindTask task)
...
+                            if (typeStr == "Error codes" || typeStr.StartsWith("Error type")) typeStr = "ErrorCodes";
+                            if (typeStr?.StartsWith("OpenCL ") == true && typeStr.Contains("deprecated"))
+                            {
+                                var typeNameStart = typeStr.IndexOf(' ', "OpenCL ".Length) + 1;
+                                var typeNameEnd = typeStr.IndexOf(' ', typeNameStart + 1);
+                                typeStr = typeStr.Substring(typeNameStart, typeNameEnd - typeNameStart);
+                            }
+                            if (typeStr?.StartsWith("Command type ") == true) typeStr = "cl_command_type";
+                            if (typeStr?.StartsWith("cl_uint ") == true) typeStr = typeStr.Substring("cl_uint ".Length);
+                            if (typeStr?.StartsWith("Additional ") == true) typeStr = typeStr.Substring("Additional ".Length);
```
Perksey, line 873: All in all I'm extremely uncomfortable using the "comment" attribute meaningfully. If there is some data in the comment attributes that we need, I'd recommend this be augmented & extracted to its own attribute in the upstream specs. Or at least this has been the approach with OpenGL.

src/Core/Silk.NET.BuildTools/Converters/Readers/OpenCLReader.cs
```diff
@@ -1025,14 +1322,49 @@ private static string FormatToken(string token)
             }
 
             var valueString = $"0x{value:X}";
-            var needsCasting = value > int.MaxValue || value < 0;
-            if (needsCasting)
+
+            return valueString;
+        }
+
+        class RenamedEntry
```
Perksey, line 1329: Can this be made a record struct? This is very verbose.

src/Core/Silk.NET.BuildTools/Converters/Readers/OpenCLReader.cs
```diff
@@ -883,65 +942,275 @@ public IEnumerable<Enum> ReadEnums(object obj, BindTask task)
...
+            var startingWithDigit = names.Where(n => n.Length > tgtPos && char.IsDigit(n[tgtPos]));
+            if (startingWithDigit.Any())
+            {
+                return FindCommonPrefix(names, containsTypeName, tgtPos - 1);
+            }
+
+            return commonPrefixFirstPass;
+        }
+
+        private string FindCommonPrefix(List<string> names, bool containsTypeName, int maxPos)
```
Perksey, line 1122: These two methods are genius! Can you move them into a common place (such as `Utilities`) so we can use them elsewhere in the generator? Preferably documented, too.

src/OpenCL/Extensions/Silk.NET.OpenCL.Extensions.INTEL/Enums/AdvancedMotionEstimationClMotionDetectDesc.gen.cs
```diff
@@ -0,0 +1,22 @@
...
+
+
+using System;
+using Silk.NET.Core.Attributes;
+
+#pragma warning disable 1591
+
+namespace Silk.NET.OpenCL.Extensions.INTEL
+{
+    [NativeName("Name", "cl_intel_advanced_motion_estimation.cl_motion_detect_desc_intel")]
```
Perksey, line 12: It looks like the type name is actually cl_motion_detect_desc_intel and this is just sort of "namespaced" with the extension name. did we want to account for this?

src/OpenCL/Extensions/Silk.NET.OpenCL.Extensions.INTEL/Enums/AdvancedMotionEstimationClMotionDetectDesc2.gen.cs
```diff
@@ -0,0 +1,26 @@
...
+
+
+using System;
+using Silk.NET.Core.Attributes;
+
+#pragma warning disable 1591
+
+namespace Silk.NET.OpenCL.Extensions.INTEL
+{
+    [NativeName("Name", "cl_intel_advanced_motion_estimation.cl_motion_detect_desc_intel.2")]
```
Perksey, line 12: Do we need to care about .2?

src/OpenCL/Extensions/Silk.NET.OpenCL.Extensions.KHR/Enums/CommandBufferProperties.gen.cs
```diff
@@ -0,0 +1,18 @@
...
+
+#pragma warning disable 1591
+
+namespace Silk.NET.OpenCL.Extensions.KHR
+{
+    [NativeName("Name", "cl_command_buffer_properties_khr")]
+    public enum CommandBufferProperties : int
+    {
+        [NativeName("Name", "CL_COMMAND_BUFFER_FLAGS_KHR")]
+        Flags = 0x1293,
```
Perksey, line 16: ngl I'm gushing over these changes, this looks so amazing and clean!

src/Core/Silk.NET.BuildTools/Converters/Readers/OpenCLReader.cs
Alexx999: Yeah, it's basically constants, plus unnamed, plus bug in XML with cl_device_info that makes it unusable

src/Core/Silk.NET.BuildTools/Converters/Readers/OpenCLReader.cs
Alexx999: Well, yeah, I've been thinking about better ways but...
Probably adding it to JSON spec same as renaming pass could be an option.

src/Core/Silk.NET.BuildTools/Converters/Readers/OpenCLReader.cs
Alexx999: Done

src/OpenCL/Extensions/Silk.NET.OpenCL.Extensions.INTEL/Enums/AdvancedMotionEstimationClMotionDetectDesc.gen.cs
Alexx999: Well, it's one of the parts where the spec is inconsistent, especially so extensions coming from Intel, those are the most "special". Basically, this is used only by Intel - can be used but only in a form of special "if vendor == Intel" handling.

src/OpenCL/Extensions/Silk.NET.OpenCL.Extensions.INTEL/Enums/AdvancedMotionEstimationClMotionDetectDesc2.gen.cs
Alexx999: Will check the extension spec

src/Core/Silk.NET.BuildTools/Converters/Readers/OpenCLReader.cs
Alexx999: I do agree that it's not perfect but it's actually the last resort thing for stuff that's otherwise dangling.
For what it's worth, all this works back to Nov 2020 spec file, and the situation seems to be (slowly) improving.

Perksey: I still think we should add extra attributes to the upstream spec rather than try and dissect a field with no defined format. This is what we've done in the past with OpenGL (we added the `group` attribute).

src/Core/Silk.NET.BuildTools/Converters/Readers/OpenCLReader.cs
```diff
@@ -835,42 +755,167 @@ public IEnumerable<Enum> ReadEnums(object obj, BindTask task)
...
+
+            var ulongEnums = new HashSet<string>(registry
+                .Elements("types")
+                .Elements("type")
+                .Where(e => e.Elements("type").SingleOrDefault()?.Value == "cl_bitfield")
+                .Select(e => Rename(e.Element("name").Value, task).Renamed));
+
+            var flagEnums = new HashSet<string>(registry
+                .Elements("enums")
+                .Where(e => e.Attribute("type")?.Value == "bitmask")
```
Perksey, line 774: Apologies for coming back with comments i didn't pick up on before... don't bitfields also constitute the flags attribute?

src/Core/Silk.NET.BuildTools/Converters/Readers/OpenCLReader.cs
```diff
@@ -835,42 +755,167 @@ public IEnumerable<Enum> ReadEnums(object obj, BindTask task)
...
+                        {
+                            if (typeStr == "Error codes") type = new RenamedEntry(type.Original, "ErrorCodes");
+                            else if (typeStr?.StartsWith("OpenCL ") == true && typeStr.Contains("deprecated"))
+                            {
+                                // Core spec deprecation notices
+                                var typeNameStart = typeStr.IndexOf(' ', "OpenCL ".Length) + 1;
+                                var typeNameEnd = typeStr.IndexOf(' ', typeNameStart + 1);
+                                type = new RenamedEntry(type.Original, typeStr.Substring(typeNameStart, typeNameEnd - typeNameStart));
+                            }
+                            else if (typeStr?.StartsWith("cl_uint ") == true) type = new RenamedEntry(type.Original, typeStr.Substring("cl_uint ".Length)); // QCOM extensions
```
Perksey, line 864:
```suggestion
                            else if (typeStr?.StartsWith("OpenCL ") && typeStr.Contains("deprecated"))
                            {
                                // Core spec deprecation notices
                                var typeNameStart = typeStr.IndexOf(' ', "OpenCL ".Length) + 1;
                                var typeNameEnd = typeStr.IndexOf(' ', typeNameStart + 1);
                                type = new RenamedEntry(type.Original, typeStr.Substring(typeNameStart, typeNameEnd - typeNameStart));
                            }
                            else if (typeStr?.StartsWith("cl_uint ")) type = new RenamedEntry(type.Original, typeStr.Substring("cl_uint ".Length)); // QCOM extensions
```

src/Core/Silk.NET.BuildTools/Converters/Readers/OpenCLReader.cs
```diff
@@ -835,42 +755,167 @@ public IEnumerable<Enum> ReadEnums(object obj, BindTask task)
...
-            
-            var removals = registry
+            Debug.Assert(enumValues != null, nameof(enumValues) + " != null");
+
+            var removals =
+                registry
                 .Elements("feature")
-                .Elements("remove")
+                .Elements("require")
+                .Where(x => x.Attribute("comment")?.Value.Contains("deprecated") == true)
```
Perksey, line 898:
```suggestion
                .Where(x => x.Attribute("comment")?.Value.Contains("deprecated"))
```

src/Core/Silk.NET.BuildTools/Converters/Readers/OpenCLReader.cs
```diff
@@ -883,65 +928,229 @@ public IEnumerable<Enum> ReadEnums(object obj, BindTask task)
...
+                        Attributes = GetTokenAttributes(name.Renamed),
+                        Doc = string.Empty,
+                        Name = Naming.Translate(TrimName(name.Renamed, task), task.FunctionPrefix),
+                        NativeName = name.Original,
+                        Value = value
+                    };
+                }).ToList();
+
+
+
```
Perksey, line 1046:
```suggestion

```

Perksey:
Happy to merge after the above.

Alexx999:
those 2 are breaking changes - sadly, C# doesn't handle "bool?" type that well

Perksey:
Ah, in that case our preferred style is “?? false” instead of “== true” for a nullable LHS.

Alexx999:
OK, done

Perksey (approved):
Thank

Perksey:
Looks like CI is failing…

Alexx999:
Simply re-running it fixed the issue, also, the previous build was OK (without the "== true" -> "?? false" change), so looks like a build system glitch to me

PR #884, opened 2022-04-16
Add empty enums as per spec

Alexx999:


src/OpenCL/Extensions/Silk.NET.OpenCL.Extensions.KHR/Structs/DeviceIntegerDotProductAccelerationPropertiesKhr.gen.cs
```diff
@@ -14,7 +14,7 @@
 
 #pragma warning disable 1591
 
-namespace Silk.NET.OpenCL
+namespace Silk.NET.OpenCL.Extensions.KHR
```
Perksey, line 17: This is an unjustifiable breaking change.

src/OpenCL/Extensions/Silk.NET.OpenCL.Extensions.ARM/Enums/SvmMemFlagsArm.gen.cs
```diff
@@ -11,7 +11,7 @@ namespace Silk.NET.OpenCL.Extensions.ARM
 {
     [System.Flags]
     [NativeName("Name", "cl_svm_mem_flags_arm")]
-    public enum SvmMemFlags : ulong
+    public enum SvmMemFlagsArm : ulong
```
Perksey, line 14: Why has the suffix appeared?

src/OpenCL/Silk.NET.OpenCL/Enums/MemProperties.gen.cs
```diff
@@ -7,13 +7,13 @@
 
 #pragma warning disable 1591
 
-namespace Silk.NET.OpenCL.Extensions.IMG
+namespace Silk.NET.OpenCL
```
Perksey, line 10: How come this has moved to the main package?

Perksey (changes requested):
Happy with the empty enums, not sure about the other changes. Once again thank you so much for helping to improve these areas, 2.16 is going to be the best update for OpenCL yet.

src/OpenCL/Silk.NET.OpenCL/Enums/MemProperties.gen.cs
Alexx999: well, it being not in the main package was a bug - it is defined as an empty enum in the base spec and extensions add stuff to it.

Perksey: ok

src/OpenCL/Extensions/Silk.NET.OpenCL.Extensions.ARM/Enums/SvmMemFlagsArm.gen.cs
Alexx999: The logic currently is "strip suffix is enum is unique" - helps with general naming quality but can lead to some glitches when adding more enums. In theory we could always strip the suffix as that is namespaced in C# but that doesn't really play that well with how typemaps currently work (being simple string-string)

Perksey: I guess this is a call upon you to decide which is more usable. I personally would prefer stripping the extension prefix from enums in extension packages, as this is how C# works, but this isn't a precedent set elsewhere in the library as everywhere else puts enums and structs in the core packages.

Alexx999: OK, interesting, so only enums are separated?
I guess this relates to a breaking change above.

Thing is, this entire story started with OpenCL enums being variably sized, so using "one size fits all" CLEnum won't fly (because it's not one size and it doesn't fit all).
So enum members must be correctly typed to get correct struct sizes, having those enums in separate assemblies doesn't help.
On top of that, having extension-specific structs in the core package serves little purpose as you'd want to reference the extension package to use them anyway.
Or am I missing something here?

And, last but not the least, OpenCL extensions are broken in Silk as of now... GetProcAddress cannot be used in OpenCL, and I'm really surprised there's not a single issue open on the topic.

Perksey: > OK, interesting, so only enums are separated?

Not sure what you mean by this, but in other APIs the features and extensions have no knowledge of groups thus we just bundle them all in the core assembly.

> Thing is, this entire story started with OpenCL enums being variably sized, so using "one size fits all" CLEnum won't fly (because it's not one size and it doesn't fit all).

Wasn't denying this. Not sure how this links back to my original prefix question though.

 > So enum members must be correctly typed to get correct struct sizes, having those enums in separate assemblies doesn't help.

Not sure how the assembly of residence and the enum size are linked, nor how this links back to my original prefix question.

> On top of that, having extension-specific structs in the core package serves little purpose as you'd want to reference the extension package to use them anyway.

Agree, but this is a breaking change we can't justify. Breaking changes we can justify are changes where the upstream specification has changed the way they represent the API or otherwise changed the spec, meaning that the breaking change occurs regardless of changes to BuildTools. Breaking changes we can't justify is where we change our interpretation of the specification, meaning that a breaking change occurs regardless of changes to the upstream specification. We plan to fix things like this in 3.0, as Vulkan has the same problem wherein extension structs are in the core packages but we aren't doing anything about it as we can't justify the breaking change outside of a major release.

Alexx999: Well, those are related if you start tracking types of enums and actually strongly type enum members.
Anyway, rolled that back.

Enum prefix removal is intended to mimic a more "human-like" naming style, so I guess it's better to have it enabled...

Perksey (approved):

PR #886, opened 2022-04-17
OpenCL Typemaps

Alexx999:
This is a showcase draft for the fully-typemapped OpenCL bindings.

Open issues:
It's _very_ breaking
Includes struct type move that we've discussed and concluded that it's unwanted (necessary due to strongly typed struct fields)


Perksey:
This all looks really good! Obviously we can't incorporate it as is for breaking change reasons, and there's stuff we've commented on before like the struct package moves, but I think if your type mapping changes could somehow be implemented in addition to the existing code we'd have a solid update on our hands!

Obviously that's now how the `TypeMapper`, works, but perhaps:
1. Adding some state (or a dictionary of state, idk) into `Profile` (I know, horrible place) containing data about OpenCL typemappings. Unless proven otherwise, this can be JSON ignored.
2. Creating simple overloaders that inspects the extra state and the native name
3. Overloading accordingly

Perksey:
I think the way I currently feel is I'll accept this PR as is and change anything I feel necessary at update time. Won't merge until update week though as I'll forget to change BuildTools otherwise.

Perksey (approved):
Had a change of heart on this. This PR updates OpenCL updates the binder to interpret the specification as-is. My original complaints about enums (e.g. removing the ungrouped enums) are unjustified given that they deviate from the specification. Will merge and release as is, thank you for this extraordinary contribution.

Perksey (approved):

PR #888, opened 2022-04-18
Fix 3.0 CI

HurricanKai:


HurricanKai:
Rest of these failures are expected / not CI-related issues. Will fix in another PR.

.github/workflows/publish-site.yml
```diff
@@ -1,35 +1,30 @@
...
+              run: npm install -g wrangler@beta
+            - name: Install Deps of CF worker
+              run: npm i
+              working-directory: ./website/cf-worker/
+            - name: Publish
+              run: wrangler publish
+              working-directory: ./website/cf-worker/
+              env:
+                  CLOUDFLARE_ACCOUNT_ID: ${{ secrets.CF_ACCOUNT_ID }}
+                  CLOUDFLARE_API_TOKEN: ${{ secrets.CF_API_TOKEN }}
```
Perksey, line 30: Generally we try not to tie any project resource to any one person e.g. all project resources belong to the .NET Foundation/they have access to give other people access. Fine for a temp solution but we won't want to go into prod with this.

.github/workflows/publish-site.yml
```diff
@@ -1,35 +1,30 @@
...
+              uses: actions/setup-dotnet@v1
+              with:
+                  dotnet-version: 6.0.201
+            - name: Run Statiq
+              run: dotnet run -c Release --project src/infrastructure/Silk.NET.Statiq/Silk.NET.Statiq.csproj -- -l debug --nocache
+            - name: Install wrangler
+              run: npm install -g wrangler@beta
+            - name: Install Deps of CF worker
+              run: npm i
+              working-directory: ./website/cf-worker/
```
Perksey, line 24: You've deleted all trace of the GitHub Pages based solution. I'd personally prefer the CloudFlare stuff to be _in addition to_ the GitHub stuff, rather than instead of. If you need ideas, I'd just copy the website into a folder matching the branch name so it's deployed to e.g. https://dotnet.github.io/Silk.NET/develop/3.0

If you need to force push, overwrite/detach history, and/or modify the main pipeline, you have my blessing.

website/cf-worker/index.js
```diff
@@ -0,0 +1,90 @@
+import {
```
Perksey, line 1: Can you move this elsewhere? as this stands this will be included in the Statiq output because it's in the input folder.

HurricanKai:
My apologies, I made you review this for no reason. Those are correct comments and in addition this temporary and bad solution doesn't even work. I'll come back to another solution (probably a version toggle or something, not sure) but will just close this for now.

PR #890, opened 2022-04-18
Update Symbols to be records

HurricanKai:


src/generators/Silk.NET.SilkTouch.Symbols/FieldSymbol.cs
```diff
@@ -7,26 +7,4 @@ namespace Silk.NET.SilkTouch.Symbols;
 /// A <see cref="FieldSymbol"/>. A field is simply a named location that can hold some type.
 /// </summary>
```
Perksey, line 8: Needs parameter documentation (everywhere)

Perksey (changes requested):

HurricanKai:
I can't really find any good docs on how to document records. I'm just doing what Rider auto-generates here.

Perksey (approved):
oh yeah that looks real good that's the stuff

PR #891, opened 2022-04-18
Better Identifier Emission & State Handling

HurricanKai:
Note that tests are indeed failing, but this is due to an oddity in roslyn that means Identifiers are created with trivia. Investigating but I'd like to merge this, ignoring the failing tests. I think it's worth tracking this oddity, but not worth blocking on.


HurricanKai:
Nevermind. I figured out why this was happening & the fix - I don't understand roslyn sometimes.

Perksey (approved):

PR #893, opened 2022-04-19
ImGui: Better disposal and multiple ImGuiController handling

Beyley:
# Summary of the PR
Before now we would lazily create an ImGui context and assume nothing else would tamper with ImGui state, along with never destroying our own context once Dispose was called

https://discord.com/channels/521092042781229087/587346162802229298/965745680503164968


HurricanKai (approved):
Thank you!

PR #898, opened 2022-04-20
Struct Layouts

HurricanKai:


src/generators/Silk.NET.SilkTouch.Symbols/StructSymbol.cs
```diff
@@ -9,11 +9,31 @@ namespace Silk.NET.SilkTouch.Symbols;
...
+    public static readonly StructLayout Empty = new StructLayout(ImmutableArray<LayoutEntry>.Empty);
+}
+
+/// <summary>
+/// Represents an entry in a <see cref="StructLayout"/>
+/// </summary>
+/// <param name="Member">The member symbol associated with this layout entry</param>
+/// <param name="ByteOffset">The offset of this entry in bytes</param>
+/// <seealso cref="StructLayout"/>
+public sealed record LayoutEntry(MemberSymbol Member, int ByteOffset);
```
Perksey, line 39: ByteOffset should probably be nullable in cases where we want the runtime to figure it out.

Perksey, line 39: Byte offsets can be different on 64-bit and 32-bit. e.g.:

```cs
public struct X
{
    [FieldOffset(0)]
    public nint Y;
    [FieldOffset(sizeof(nint))]
    public nint Z;
}
```

How would we represent this?

src/generators/Silk.NET.SilkTouch.Symbols/StructSymbol.cs
```diff
@@ -9,11 +9,31 @@ namespace Silk.NET.SilkTouch.Symbols;
...
 // /// Additionally it may contain one or multiple <see cref="MethodSymbol"/> that are called with an instance of this type as their first argument.
-public sealed record StructSymbol(IdentifierSymbol Identifier, ImmutableArray<MemberSymbol> Members) : TypeSymbol(Identifier);
+public sealed record StructSymbol(IdentifierSymbol Identifier, StructLayout Layout) : TypeSymbol(Identifier);
+
+/// <summary>
+/// A <see cref="StructSymbol"/> representing the layout of a <see cref="StructSymbol"/>
+/// </summary>
+/// <param name="Entries">The entries of this layout</param>
+public sealed record StructLayout(ImmutableArray<LayoutEntry> Entries)
+{
```
Perksey, line 26: Did we want to support auto layout in cases where we don't want to guarantee the memory layout to our users?

Perksey (commented):
Looks good, just a few exploratory questions.

src/generators/Silk.NET.SilkTouch.Symbols/StructSymbol.cs
HurricanKai: Do we ever want that? I've just glanced at Clangsharp, I thought that always tells us where exactly the field should go?

HurricanKai: Not sure, but I'm not sure what the expected emitted code would be either, sizeof(nint) is not constant.
https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABABgAJiBGAOgCUBXAOwwEt8YaBJFmKCAA4BlPgDdWYGLgDcAWABQAbSEYoDMBgAy2AJ4QGGABTa9BgNKsmAExoBRBAIA2E1hgCUAXXILiAZnK4quoY5AAaCgDeCuQx5IoAYqwwjlYA8gBm6bgwRqSe0bF+5EyWIQCacvKxcYnJaZnZRrisAF4wEOmGJSxu+VWF/t0hAFqVAL5AA==

Perksey: I'm not sure if the XML does all of the time, but ClangSharp always has that data available to it internally at least. In any case, ClangSharp by default uses implicit layout in the bindings it generates unless it's necessary to be explicit. In cases where it's explicit, it is always present in the XML.

HurricanKai: OK, then it's reasonable to include a sequential layout. I'll open up a separate PR for that though, if that's OK. Trying to get as good of a git history as possible

Perksey: fine

HurricanKai: (Work Item 69)

Perksey (approved):

src/generators/Silk.NET.SilkTouch.Symbols/StructSymbol.cs
Perksey: (nice)

PR #907, opened 2022-05-07
Initial XML generation from C++

HurricanKai:
# Summary of the PR
This only gets XML from a C++ header file. The configuration options are minimal, to allow incrementally adding the things we actually need.
Links go here.

# Further Comments
Note that the branch is inappropriately named, I've decided to develop the second part of this feature, visiting the XML and scraping the symbols from it, in a separate PR.
There are a lot of deletions of code in this PR. This is only to ensure having fully tested code, and I'll work to restore all of the code in one way or another down the road.
Docs aren't updated & out of date, I can fix this but I think due to the amount of changes that are currently happening there is no sense in writing docs while I still rework the code.


HurricanKai:
Good thing we squash PRs... For some reason my local git install prepends the current path if the commit message starts with `:`

Perksey (approved):
If you're happy I'm happy, I'll turn off the required review thing in a bit.

PR #908, opened 2022-05-07
Thin down CI

HurricanKai:
3.0 CI times are still quite long and our CI is clearly tailored to pull requests being big and heavy, with the target going straight to NuGet.
Just thinning down the CI a bit while 3.0 is being developed actively, when it's released we want to bring back much of this.


Perksey (approved):
3.0 blanket approval

PR #945, opened 2022-06-04
Add backwards compatibility with old CL enums, bindings improvements

Perksey:
#886 was merged adding significantly nicer enums to Silk.NET.OpenCL, with the caveat that all of the old-style enums were entirely removed/broken. The aim of this PR is to add logic to the generator to minimize the breaking changes as much as possible such that we can release this in the 2.X cycle.

There are also some unrelated changes which I'm unsure what they were triggered by (basically anything to do with calling conventions), but I looked into it and it seems fine don't worry about it.

Also contains the latest bindings seeing as you can't really avoid regenerating the bindings when working on the generator.


Perksey:
Ready for review

HurricanKai (approved):

PR #948, opened 2022-06-05
Shorter enum names everywhere

### LINKED ISSUE #458: Better naming for generated Vulkan enum and flag values
# Summary of feature
The generated enum values (regardless if bitflag or not) are often unnecessary long. For example in https://github.com/dotnet/Silk.NET/blob/main/src/Vulkan/Silk.NET.Vulkan/Enums/DebugUtilsMessageSeverityFlagsEXT.gen.cs
where the full name of the enum is a prefix of every value for example "DebugUtilsMessageSeverityVerboseBitExt".

This leads to extremly long code.

Some of the enums are handled in a more C# way, like https://github.com/dotnet/Silk.NET/blob/main/src/Vulkan/Silk.NET.Vulkan/Enums/StructureType.gen.cs where the prefix was removed.

Could all enums (and therefor flags) be handled like StructureType as well or are there reasons why not?

My suggestion would be:
* for all enum values to remove the common prefix (without the extension) and the extension at the end.
* for all flag values to remove the common prefix (without Flags and extension) and Bit + extension at the end.

# Comments
* I would like to help if needed, but got lost in the BuildTools project.
* A general naming scheme might need to be discussed upfront:
  1. Keep extension name (KHR, EXT, etc.) on value names or remove them?
  2. Always remove prefix or are there exceptions?

# Does this have a proposal?
Not as of now.

Perksey:
Closes #458 

Takes the common prefix logic from #881 and applies it library-wide.

This pull request depends on #945


HurricanKai (approved):
I'll need to debase this brach before merging. LGTM

Perksey:
Rebased and ready to go.

src/Vulkan/Silk.NET.Vulkan/Enums/VideoCodecOperationFlagsKHR.gen.cs
```diff
@@ -15,15 +15,30 @@ public enum VideoCodecOperationFlagsKHR : int
...
+        [NativeName("Name", "VK_VIDEO_CODEC_OPERATION_INVALID_BIT_KHR")]
+        InvalidBitKhr = 0,
+        [NativeName("Name", "VK_VIDEO_CODEC_OPERATION_ENCODE_H264_BIT_EXT")]
+        EncodeH264BitExt = 65536,
+        [NativeName("Name", "VK_VIDEO_CODEC_OPERATION_ENCODE_H265_BIT_EXT")]
+        EncodeH265BitExt = 131072,
+        [NativeName("Name", "VK_VIDEO_CODEC_OPERATION_DECODE_H264_BIT_EXT")]
+        DecodeH264BitExt = 1,
+        [NativeName("Name", "VK_VIDEO_CODEC_OPERATION_DECODE_H265_BIT_EXT")]
+        DecodeH265BitExt = 2,
```
HurricanKai, line 42: So sorry that I'm noticing this only now.
Shouldn't these be without the "BitKhr/Ext" suffixes?

HurricanKai:
In a future PR we should look to move the new members to the top of the declaration, this will result in `.ToString` and similar reflection-using functions to likely pick up the new names.

src/Vulkan/Silk.NET.Vulkan/Enums/VideoCodecOperationFlagsKHR.gen.cs
HurricanKai: Ok, if the bit is consistent and not ment to be excluded (and isn't in other places) I'm happy to merge this, just not too sure what the intended rules are here

Perksey: Can confidently say that this is as intended :D

PR #972, opened 2022-07-09
Update ImGui.NET to 1.87.3

Beyley:
# Summary of the PR
Updates all ImGui extensions to the latest revision of ImGui.NET

# Further Comments
Also puts all of them in sync, as before one of the extensions was a different version to the other 2


Perksey:
Recommend revert, as it stands ImGui has (unintentionally) removed Mac support which is a bit of a regression.

HurricanKai:
There'll be a separate PR to rollback all the versions to one common known good version

PR #988, opened 2022-08-03
Optimize Scalar.As

Perksey:


Beyley (approved):
looks good to me, although that isn't legal advice i'm not a doctor i think at least

HurricanKai (changes requested):
I'll check the disasm and merge. I'm assuming this is good, but blocking merge until we can be sure.

HurricanKai (approved):
Only checked discord now and saw you've already done everything 😅

Perksey:
I'm hoping one day this will light up, but failing that I need to add a "mono path" that stops mono from producing invalid assembly (basically just having a file with the old, unoptimized code in it. I think our blazor activity gives us reasonable evidence that it does try to specialize these generics even with the code unoptimized for CoreCLR so we may be good to do so anyway. sucks, but that's mono for you)

PR #990, opened 2022-08-04
Rework Type Refs into Generic -> External & Internal

HurricanKai:


Perksey:
Look at all those CI jobs go!

PR #998, opened 2022-08-04
Correctly Visit types

HurricanKai:
This visits types correctly.
Tests will fail, which is an unfortunate necessity to not make commits too large.


HurricanKai:
(Do not merge until CI has been fixed in #999)

HurricanKai:
Needs merging as #999 needs the interface wiring introduced in this PR

PR #1000, opened 2022-08-05
Add Direct2D to Silk

curin:
Adding bindings for Direct2D to silk.


src/Core/Silk.NET.Core/Native/Structs/FileTime.cs
```diff
@@ -0,0 +1,15 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Silk.NET.Core.Native
+{
+    public struct FileTime
```
Perksey, line 10: This struct already exists in Silk.NET.Core.Win32Extras

src/Core/Silk.NET.Core/Native/Structs/ISequentialStream.cs
```diff
@@ -0,0 +1,128 @@
...
+using Silk.NET.Core.Attributes;
+using Silk.NET.Core.Contexts;
+using Silk.NET.Core.Loader;
+
+#pragma warning disable 1591
+
+namespace Silk.NET.Core.Native
+{
+    [Guid("0c733a30-2a1c-11ce-ade5-00aa0044773d")]
+    public unsafe partial struct ISequentialStream
```
Perksey, line 18: Is there no way we can add objidl.h to the headers for Core? These look like they were generated but then manually copied to core (judging by the lack of .gen.cs)

src/Core/Silk.NET.Core/Native/Structs/FileTime.cs
curin: I missed that, it does mean that the IStream would need to move to that as they require this.

src/Core/Silk.NET.Core/Native/Structs/ISequentialStream.cs
curin: We could add them to core, but more likely would be Win32Extras as they rely on FileTime. These were also written by hand, just following the pattern of a generated file.

curin: So just looking into it, first objidlbase.h is what we would be adding, objidl.h isn't generating the structures. The only issue I'm seeing is that if we include all of that we need some types from WTypesBase.h(_COAUTHINFO, which relies on _COAUTHIDENTITY, for a struct) and rpcdcep.h(_RPC_MESSAGE for a bunch of global functions). 
I think we can these two without needing any more, just feels like this is a lot to add for IStream. What do you think?

src/Core/Silk.NET.Core/Native/Structs/FileTime.cs
curin: this is now fixed as objidlbase.h was added to win32extras

src/Core/Silk.NET.Core/Native/Structs/ISequentialStream.cs
curin: I did add objidlbase.h to win32Extras with its dependancies, WTypesBase.h and rpcdcep.h. the rpcdcep.h types seem to not be read correctly by clang and only the define consts seem to be read correctly, but I did bring in the required structs in manually.

Perksey (approved):
Sorry for the delay, thank you so much for this amazing contribution!

Perksey:
@Beyley we'll merge this once your changes are in, there will inevitably be conflicts that I'm happy to work to resolve

Perksey:
@Beyley I will work to merge this now, easier if you can rebase instead of waiting and prolonging a big rebase on my end.

PR #1002, opened 2022-08-05
Default includes

HurricanKai:
Enables OSX Support & Resolves default Includes

### DIFF
--- a/src/generators/Silk.NET.SilkTouch.Scraper/ClangScraper.cs
+++ b/src/generators/Silk.NET.SilkTouch.Scraper/ClangScraper.cs
@@ -102,10 +102,23 @@ public sealed class ClangScraper
                         {
                             osxSdkIncludePath = p;
                         }
+                        else
+                        {
+                            throw new Exception($"No Sub {output}");
+                        }
+                    }
+                    else
+                    {
+                        throw new Exception($"Does not exist {output}");
                     }
                 }
+                else
+                {
+                    throw new Exception
+                        ($"xcrun failed to start. {process.ExitCode} | \"{process.StandardOutput.ReadToEnd()}\"");
+                }
             }
-            catch { /* */}
+            catch { throw; }
             if (osxSdkIncludePath is not null)
             {
                 yield return osxSdkIncludePath;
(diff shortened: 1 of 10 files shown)


src/generators/Silk.NET.SilkTouch.Scraper/ClangScraper.cs
```diff
@@ -57,6 +59,51 @@ public IEnumerable<Symbol> ScrapeXML(XmlDocument document)
...
+            if (VisualStudioResolver.TryGetVisualStudioInfo(out var info))
+            {
+                if (info.UcrtIncludes.Length > 0)
+                {
+                    foreach (var include in info.UcrtIncludes)
+                    {
+                        yield return include;
+                    }
+                }
+                else
```
Perksey, line 96: Don't we want both?

HurricanKai: I think this is the ucrt vs msvc code? GH lost track of the snippet.
Not an expert, just quickly googled online, and there I got the impression ucrt is just a newer rewrite of C(++) standard headers, while msvc is the older stuff?

Perksey: I think the naming is misleading, the ucrt array is likely to only contain windows sdk headers and the MSVC tools contains the standard C headers.

HurricanKai: Ah, I see. Will change in that case :+1:

src/generators/Silk.NET.SilkTouch.Scraper/ClangScraper.cs
```diff
@@ -57,6 +63,51 @@ public IEnumerable<Symbol> ScrapeXML(XmlDocument document)
         return visitor.Visit(bindings);
     }
 
+    private string GetXCodeSdkPath()
+    {
+        var process = new Process();
+        process.StartInfo = new ProcessStartInfo("xcrun", $"--show-sdk-path --sdk {_options.Value.XcodeSdk}")
```
Perksey, line 69: This might be null

PR #1004, opened 2022-08-05
Use locked dependencies & cache in CI

HurricanKai:


HurricanKai:
Given there appears to be no way to restore for each RID and I don't want contributors to download gigabytes of stuff just for this little bit of hardening.

PR #1008, opened 2022-08-06
Type Resolution

HurricanKai:
The branch is named incorrectly, shouldn't matter though.

A followup PR will add this to integration, this PR only adds the type resolution itself.


HurricanKai:
I think this is done. Waiting for CI. Not too sure whether this is the right approach to references in immutable trees, but seems to work for now.

PR #1010, opened 2022-08-07
DirectX usability improvements

Perksey:
- [x] Adds a IComVtbl interface implemented by:
    - [x] ComPtr
    - [x] All COM interfaces
- [x] Adds extension methods to invoke COM methods through IComVtbl
- [x] Adds overloads that return ComPtrs in place of raw pointers and/or IID_PPV_ARGS
- [x] Implements HelloQuad using the newly-reborn Direct3D 11 bindings


Perksey:
Ready for review, but this was an absolute bear to get through the door so I'm probably only going to be doing small rework to it.

I welcome scrutiny on the example, but it is "best effort" only and there is still some improvements that can be done in the bindings to make it truly exemplar. If there are any significant comments for the example they will either have to be done by someone else or the example moved to the lab.

Really not interested in excessive rework for this one.

src/Core/Silk.NET.BuildTools/Bind/ClassWriter.cs
```diff
@@ -443,6 +443,7 @@ public static void WriteMixedModeClasses(this Project project, Profile profile,
                         sw.WriteLine("}");
                         sw.WriteLine();
                         sw.Flush();
+                        sw.Dispose();
```
Beyley, line 446: Why not use the `using` syntax here instead of manual disposal?

Perksey: Because I didn't have the time to evaluate whether it would be safe to do so from a scope POV. I essentially just did a Ctrl+F for `Flush` and whacked a `Dispose` after each call.

Beyley: ah, makes sense

Beyley (approved):
Im not seeing anything i dislike, this is a really good improvement, and seems to improve bindings across the board too, not just DX

PR #1012, opened 2022-08-08
Update Docs

HurricanKai:
This also updates a bunch of the other docs.


Beyley (approved):

Perksey:
General advice as these seem pretty light (at least for now), if you went in any of the projects, could you use these docs to describe what all of the files do and how they all link together?

HurricanKai:
Added a bit more info & list of symbols. Anything I forgot to mention / that's unclear?

Perksey:
- What is the Type ID? Where does it come from?
- Where does the type store come from in SymbolVisitor? Is it new?
- Could do with some example code, might be in the form of a link or an actual code snippet.
- How do the type references work? Again links back to what are the type IDs.
- Formatting documentation is awesome!
- Visitor documentation is awesome!
- Type resolution documentation doesn't actually indicate how the resolution logic works. I appreciate XML docs are used for individual type/function documentation, but the essence of the logic should still be documented I think.
- Scraper documentation is a bit lacking as well, not really describing the essence of the logic.
    - e.g. "we invoke ClangSharp to get XML outputs, then we read that XML output (via the XmlVisitor) and convert it into the symbols described by the symbol layer, and that gets passed to ???"
- Clang doesn't output XML, ClangSharp does though. This could be confusing as to where to look for further information.
- Testing documentation is awesome!

Future:

- Where do all of these symbols come from? Is there an example of them being created?
- Where is the SymbolVisitor used in the other components? Is there an example you can link to?

HurricanKai:
I've addressed most of those concerns, thanks a lot for the detailed response!
I'm a little confused about the scraper docs, not sure what else I can really add, can you elaborate?
The hope would be that the scraper is very straight forward and only a small bridge between ClangSharp XML output and the symbol layer.

HurricanKai:
Merging this to unblock follow up work, but I'll follow up in another PR. Feel free to comment here

documentation/for-contributors/generators/scraper.md
```diff
@@ -1,39 +1,13 @@
...
-
-TODO: Write words here about each XML mod once XML mods have been implemented
-
-## Transformation
-
-After the XML mods have been applied, the transformation stage is responsible for converting the bindings represented by the XML tree back into C# code that can be output via the SilkTouch Context (and added to the project compilation). Unlike ClangSharp, this stage will output bindings which are compatible with/invoke the Emitter and Overloader instead of using DllImport directly.
-
-TODO: Write more words here once this stage is implemented.
+This is what happens in `ClangScraper.cs`, which is configured using `ClangScraperConfiguration.cs`
+Check XML docs for details on what methods do.
```
Perksey, line 13: You should describe the resolution _logic_ here as well i.e. how we look for Windows SDKs, how look for XCode SDKs, as this is valuable information. Might be fit for user documentation, though.

Perksey:
> The hope would be that the scraper is very straight forward and only a small bridge between ClangSharp XML output and the symbol layer.

So say that! :D

PR #1025, opened 2022-08-11
Fix xinput GetApi()

ohtrobinson:
# Summary of the PR
Adds an implementation of `GetApi()` for XInput which currently throws a not implemented exception.

https://discord.com/channels/521092042781229087/587346162802229298/1007383458978467920

# Further Comments
None.

### DIFF
--- a/src/Microsoft/Silk.NET.XInput/XInput.cs
+++ b/src/Microsoft/Silk.NET.XInput/XInput.cs
@@ -12,7 +12,7 @@ namespace Silk.NET.XInput
     {
         public static XInput GetApi()
         {
-             throw new NotImplementedException();
+            return new XInput(CreateDefaultContext("XInput1_3"));
         }
 
         public bool TryGetExtension<T>(out T ext)


ohtrobinson:
Did some tests and after installing the DirectX 2010 sdk it is working fine.

src/Microsoft/Silk.NET.XInput/XInput.cs
```diff
@@ -12,7 +12,7 @@ public partial class XInput
     {
         public static XInput GetApi()
         {
-             throw new NotImplementedException();
+            return new XInput(CreateDefaultContext("XInput1_3"));
```
HurricanKai, line 15: We usually use a library name container instead of hardcoding the name for platform compat. Do we not want this for XInput for some reason?

ohtrobinson: I did this because the same thing is done in `XAudio`, and also I couldn't find a library container for XInput, not one rider could find anyway.

Beyley (approved):

src/Microsoft/Silk.NET.XInput/XInput.cs
mathew-odwyer: Does this mean I need to install XInput1_3 somewhere on my machine?

PR #1027, opened 2022-08-12
Dotnet tool

HurricanKai:


HurricanKai:
This also adds JSON serialization capabilities to TypeId, and a sample on how symbols may be serialized in DotnetTool. This logic will likely be reused for diffing support.

PR #1108, opened 2022-10-24
Remove NoWarn

Youssef1313:


Perksey (changes requested):
I don't think this should be removed, this is part of the NUKE template and also will produce a lot of warnings when removed: it is a NUKE style guide to ignore these warnings when writing code.

Youssef1313:
> will produce a lot of warnings when removed

Why not fix these warnings instead of ignoring them?

Perksey:
Again, it is a NUKE style guide to ignore these warnings. iirc one of those is missing accessibility attributes which are intentionally omitted when writing NUKE build scripts for example.

Youssef1313:
@Perksey Both warnings are related to redundant fields. One of them is fields that are never assigned, and the other is about fields that are never used.

I don't see how this could be a "style guide"

Example code that produces both warnings:

```csharp
public class C
{
    // warning CS0649: Field 'C._x' is never assigned to, and will always have its default value 0
    private int _x;

    // warning CS0169: The field 'C._y' is never used
    private int _y;

    public void M()
    {
        _ = _x;
    }
}
```

Perksey:
Ah in that case that has to do with how NUKE resolves parameters, though the unused field one is interesting. Nonetheless, unless you intend to fix the warnings please don't remove the NoWarn.

Perksey:
Closing this due to inactivity, feel free to reopen.

PR #1123, opened 2022-11-02
OpenGL Tutorials

ohtrobinson:
# Summary of the PR
This PR adds OpenGL tutorials to the Silk.NET website, using the code provided by the existing samples.

https://discord.com/channels/521092042781229087/587346162802229298/1037331531619905596

# Further Comments
I'd like to add a couple more tutorials before merging the PR, just so that there is enough to get started with.


ThomasMiz:
Thanks for the contribution! I like what you've written, but I will be leaving a few comments 👀

website/docs/opengl/1-hello-window.md
```diff
@@ -0,0 +1,263 @@
...
+            "DateTimeWritten": "02/11/2022 12:00",
+            "PreviewImage": ""
+        }
+    }
+}
+---
+
+# 1.1 - Hello Window
+> [!Note]
+> You can view the source code for this tutorial [here.](https://github.com/dotnet/Silk.NET/blob/main/examples/CSharp/OpenGL%20Tutorials/Tutorial%201.1%20-%20Hello%20Window/Program.cs)
```
ThomasMiz, line 17: The table looks a bit odd? Maybe if we use more human-readable text, as in instead of "TableOfContents" "Table Of Contents".

I'm not sure about the ```[!Note]``` notation, at least to me it's not displaying correctly.

image

website/docs/opengl/1-hello-window.md
```diff
@@ -0,0 +1,263 @@
...
+
+private static void OnUpdate(double dt) { }
+
+private static void OnRender(double dt) { }
+```
+
+These are the methods that will power our main application. For this tutorial, we'll only be using `OnLoad`, but it's helpful to have all three.
+
+> [!Note]
+> The `dt` parameter (which you will often see labelled as `obj`), is the **delta time**, which is the amount of time, in seconds, that has passed since the last frame. Using delta time is key for framerate-independent actions, such as movement.
```
ThomasMiz, line 151: Same comment about the ```[!Note]``` as before

website/docs/opengl/1-hello-window.md
Perksey: This is required for the website to encapsulate metadata and is never seen by the user, see Silk.NET.Statiq. 

As for [!Note], yeah I need to add a shortcode for this at some point to actually turn into a pretty note

website/docs/opengl/1-hello-window.md
```diff
@@ -0,0 +1,263 @@
...
+
+We won't be using the `OpenGL` package in this tutorial, however we'll be using it in every other tutorial, so it's a good idea to install it now.
+
+Using the dotnet cli, installing the packages is simple:
+
+```
+dotnet add package Silk.NET.Windowing
+dotnet add package Silk.NET.Input
+dotnet add package Silk.NET.OpenGL
+```
```
ThomasMiz, line 64: Or if you're using Visual Studio, you can use the NuGet Package Manager by right-clicking on your project in the Solution Explorer and selecting "Manage NuGet Packages..."

I think this would be a good addition to the tutorial given that a lot of people use Visual Studio for C#

website/docs/opengl/1-hello-window.md
ThomasMiz: Ah, I see, forget I said anything lol

website/docs/opengl/1-hello-window.md
ThomasMiz, line 151:
```suggestion
> The `dt` parameter is the **delta time**, which is the amount of time, in seconds, that has passed since the last frame. Using delta time is key for framerate-independent actions, such as movement.
```

I think it's better not to mention "obj", which is the name Visual Studio gives the parameter when it autocompletes the method, as we don't want people actually naming it "obj" lol

website/docs/opengl/1-hello-window.md
```diff
@@ -0,0 +1,263 @@
...
+_window.Load += OnLoad;
+_window.Update += OnUpdate;
+_window.Render += OnRender;
+```
+
+Launch your program again and you will see that... nothing has changed. Good! That means it's working correctly. Try adding some logs in the load, update, and render methods to see exactly when they are called.
+
+![Loading, Updating & Rendering](../../images/opengl/chapter1/loading-rendering.png)
+
+We're now ready to handle some input!
```
ThomasMiz, line 165: I would consider specifically mentioning/showing you added Console.WriteLine() (or similar) and say "Try temporarily adding [...] and render methods so we can see in the console when they are being called:"

Or in the "final result" code block at the end, add the logging too

ThomasMiz:
Nicely done!
- I like the friendly writing
- You explain what each line of code does instead of just it, letting readers better understand
- Nice to mention the "next tutorial", so readers know what to do next
- Nice to link the discord server!

website/docs/opengl/1-hello-window.md
ohtrobinson: Sounds good, I'll add that - I did mean to add it but I haven't used VS for nearly 2 years so couldn't remember..

website/docs/opengl/1-hello-window.md
ohtrobinson: The main reason I did this was so people knew that the obj parameter was the delta time in their editor, although I am happy to remove it.

website/docs/opengl/1-hello-window.md
```diff
@@ -0,0 +1,263 @@
...
+```cs
+WindowOptions options = WindowOptions.Default;
+```
+
+We'd still like to change a few of these options for our application, though, so go ahead and add the following underneath that line:
+
+```cs
+options.Size = new Vector2D<int>(800, 600);
+options.Title = "My first Silk.NET application!";
+```
```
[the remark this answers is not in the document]
ohtrobinson: I left it out for a couple reasons - first, I was trying to stick as much as possible to the already existing samples (so if I change this, perhaps it should be changed in the sample too?), but I also did it to make it a bit easier to read, so that a newbie can see "okay, we are using the default options and then changing a few bits", I doubt many of them have seen the `with` syntax.. Heck, I didn't know it existed until reading this

website/docs/opengl/1-hello-window.md
ohtrobinson: I've written that here: `Try adding some logs in the load, update, and render methods to see exactly when they are called.`, on line 161. Let me know if I should change it/make it clearer though. I've also added the logging to the final result now.

ohtrobinson:
I've added most of the suggested changes - just left a couple that I will wait on your response for before I add them.

I was originally going to change the `> [!Note]` stuff, but seems like you're planning to add that? It would be very handy (my tutorials will make heavy use of them), as will warnings too.

website/docs/opengl/1-hello-window.md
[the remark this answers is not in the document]
ThomasMiz: iirc when I was writing the OpenGL project templates we had the same discussion, and I was on the same spot as you are now of just finding out that syntax existed lol

Back then we decided to use the new syntax, so I think we should stick to that

website/docs/opengl/1-hello-window.md
ThomasMiz: (Just to be clear, I don't think the logging should stay in future tutorials, I just want to make sure newbies don't get confused and think that Silk is supposed to be printing that on its own, so it helps to be able to see that it was _you_ that added the code to do that)

website/docs/opengl/1-hello-window.md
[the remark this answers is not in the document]
ohtrobinson: alrighty, i'll update it with the new syntax.

ThomasMiz:
Looking great!

Perksey:
> I was originally going to change the > [!Note] stuff, but seems like you're planning to add that? It would be very handy (my tutorials will make heavy use of them), as will warnings too.

I mean, I did intend on adding it at some point. Don't know when I'll get a chance, but essentially it'll just end up being similar to the FancyImage shortcode (implemented in FancyImageShortCode.cs) to output something like
```html
<div class="alert alert-info">
Content...
</div>
```

Adding a shortcode is very easy, tbh the Silk.NET.Statiq project is super easy to navigate so if you wanted to give it a go by all means do! If not, just leave it as is for now and hopefully someone gets round to it.

Perksey:
Giving @ThomasMiz authority over this one given Thomas is probably the best maintainer for documentation/OpenGL user experience (and given I don't have anything to add :))

ohtrobinson:
> Adding a shortcode is very easy, tbh the Silk.NET.Statiq project is super easy to navigate so if you wanted to give it a go by all means do! If not, just leave it as is for now and hopefully someone gets round to it.

I'll take a look at adding it, it will be very handy to have for the tutorials.

website/docs/opengl/1-hello-window.md
```diff
@@ -0,0 +1,263 @@
...
+```cs
+private static void KeyDown(IKeyboard keyboard, Key key, int keyCode) { }
+```
+
+Now that we've done that, we can subscribe to it in our `Load` method. Add the following after you create the input context:
+
+```cs
+for (int i = 0; i < input.Keyboards.Count; i++)
+    input.Keyboards[i].KeyDown += KeyDown;
+```
```
BitzArt: Hello! Consider this:
```
foreach (var keyboard in input.Keyboards)
            keyboard.KeyDown += KeyDown;
```

website/docs/opengl/1-hello-window.md
```diff
@@ -0,0 +1,263 @@
...
+
+This will create our input context. To handle keyboard input, we need to subscribe to the `KeyDown` event.
+
+First, let's create the method that handles this event:
+
+```cs
+private static void KeyDown(IKeyboard keyboard, Key key, int keyCode) { }
+```
+
+Now that we've done that, we can subscribe to it in our `Load` method. Add the following after you create the input context:
```
BitzArt: You probably mean `OnLoad` and not `Load`?

website/docs/opengl/c1/2-hello-quad.md
```diff
@@ -0,0 +1,140 @@
...
+
+```cs
+using System.Drawing;
+```
+
+Then, in your `OnLoad` method, underneath `GetApi()`, add the following:
+
+```cs
+_gl.ClearColor(Color.CornflowerBlue);
+```
```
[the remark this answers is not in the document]
Perksey: Silk.NET is a 1:1 binding of OpenGL, and as a result the names are mapped 1:1. Please consult the OpenGL documentation for any clarifications on the bindings themselves.

(...and we're certainly not going to rename anything within 2.X as that'd be massively breaking!)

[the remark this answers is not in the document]
Redhacker1: Might be worth suggesting on 3.0 in one of the meetings though.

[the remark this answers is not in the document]
ohtrobinson: Can't imagine adding it into 3.0 is a good idea, as again that would go against the 1:1 bindings, and would certainly annoy plenty of users, myself included.

If some people wanted this type of functionality, perhaps a small library could be written over the GL functions to do this behaviour instead, and the developer could apply to have it as part of SilkCommunity (I'm not sure what the guidelines are there..). Just an idea, I personally won't be developing anything like that since `_gl.ClearColor = Color.CornflowerBlue;` is no real different to `_gl.ClearColor(Color.CornflowerBlue)`, so imo the change is a bit pointless.

ohtrobinson:
I will get back to the tutorial shortly, just juggling various projects and some irl issues. I'd like to complete the tutorial I'm currently working on and then I'll probably send a review request so it can potentially be merged! Ultimately I'd like to cover all existing samples, including OpenAL, before I branch out into more advanced topics and I'd also like to do D3D ones as well if someone doesn't start that before I do.

website/docs/opengl/1-hello-window.md
ohtrobinson: I will stick with the original as the code samples uses it, this can be revisited later.

ohtrobinson:
Okay, both tutorials should finally be complete. The site builds fine, and if all is well I am happy for this to be merged if you wish to.

I'll PR in the next set of tutorials later down the line.

website/docs/opengl/c1/1-hello-window.md
```diff
@@ -0,0 +1,271 @@
...
+
+Using the dotnet CLI, installing the packages is simple:
+
+```
+dotnet add package Silk.NET.Windowing
+dotnet add package Silk.NET.Input
+dotnet add package Silk.NET.OpenGL
+```
+
+Or, if you're using Visual Studio, you can use the NuGet Package Manager by right-CLIcking on your project in the solution explorer and selecting "Manage NuGet Packages...".
```
ThomasMiz, line 65:
```suggestion
Or, if you're using Visual Studio, you can use the NuGet Package Manager by right-clicking on your project in the solution explorer and selecting "Manage NuGet Packages...".
```

It looks like you ctrl-F-ed wrong hah

website/docs/opengl/c1/1-hello-window.md
```diff
@@ -266,3 +267,5 @@ public class Program
     }
 }
 ```
+
+You can also view this on its own [here.](../sources/1.1-final-result.html)
```
ThomasMiz, line 271:
```suggestion
You can also view this on its own [here](../sources/1.1-final-result.html).
```

website/docs/opengl/c1/1-hello-window.md
```diff
@@ -0,0 +1,271 @@
...
+            "AuthorGitHub": "ohtrobinson",
+            "DateTimeWritten": "02/11/2022 12:00",
+            "PreviewImage": ""
+        }
+    }
+}
+---
+
+# 1.1 - Hello Window
+<?# Info "You can view the source code for this tutorial [here.](../sources/1.1-final-result.html)" /?>
```
ThomasMiz, line 16:
```suggestion
<?# Info "You can view the source code for this tutorial [here](../sources/1.1-final-result.html)." /?>
```

website/docs/opengl/c1/2-hello-quad.md
```diff
@@ -0,0 +1,524 @@
...
+            "AuthorGitHub": "ohtrobinson",
+            "DateTimeWritten": "02/11/2022 12:00",
+            "PreviewImage": ""
+        }
+    }
+}
+---
+
+# 1.2 - Hello Quad
+<?# Info "You can view the source code for this tutorial [here.](../sources/1.2-final-result.html) This tutorial builds on the previous tutorial. If you haven't read it, you can do so [here.](1-hello-window.html)" /?>
```
ThomasMiz, line 16:
```suggestion
<?# Info "You can view the source code for this tutorial [here](../sources/1.2-final-result.html). This tutorial builds on the previous tutorial. If you haven't read it, you can do so [here](1-hello-window.html)." /?>
```

website/docs/opengl/c1/2-hello-quad.md
```diff
@@ -0,0 +1,524 @@
...
+
+## Clearing the window
+Before we start this, let's take a look at what makes up an OpenGL window. 
+
+A window consists of several **framebuffers**. A framebuffer is essentially a texture that can be drawn to by OpenGL. An OpenGL window will consist of the following framebuffers:
+
+* Color buffer
+* Depth stencil buffer
+
+On top of this, a window will often contain at least two of these buffers. This is known as **double-buffering** and is imperitive for VSync to work properly. One buffer is displayed, while another is rendered to.
```
ThomasMiz, line 64: I don't think double-buffering is specifically about VSync. If you're always rendering to the same framebuffer, then what will the screen display? The resulting frame will only exist for a brief period of time between you finishing your draw calls for that frame and you clearing the framebuffer to start the next. You can't display a framebuffer that's currently being worked on!

website/docs/opengl/c1/2-hello-quad.md
```diff
@@ -0,0 +1,524 @@
...
+private static GL _gl;
+```
+
+Finally, in your `OnLoad` method, add the following:
+
+```cs
+_gl = _window.CreateOpenGL();
+```
+
+What are we doing here? Silk.NET requires you to keep a **reference** to the OpenGL API. If you've used or seen OpenGL in C, you'll notice that this is different to the way that it is done there. This is done so that you can more easily keep track of multiple contexts. If you don't know what that is, don't worry about it for now, we won't be using it in these tutorials.
```
ThomasMiz, line 50: > This is done so that you can more easily keep track of multiple contexts.

@Perksey Just wanna check with you that this is correct

website/docs/opengl/c1/2-hello-quad.md
```diff
@@ -0,0 +1,524 @@
...
+Now, run your application again. If all is good, you should see no change. Awesome! Let's do our first steps in OpenGL: Clearing the window.
+
+## Clearing the window
+Before we start this, let's take a look at what makes up an OpenGL window. 
+
+A window consists of several **framebuffers**. A framebuffer is essentially a texture that can be drawn to by OpenGL. An OpenGL window will consist of the following framebuffers:
+
+* Color buffer
+* Depth stencil buffer
+
```
ThomasMiz, line 63:
```suggestion
A window consists of several default **framebuffers**. A framebuffer is a set of buffers that can be rendered to, essentially a texture that can be drawn to by OpenGL. An OpenGL window's default framebuffers will consist of the following buffers:

* Color buffer
* Depth stencil buffer
```

According to [the khronos docs](https://www.khronos.org/opengl/wiki/Framebuffer):
> A Framebuffer is a collection of buffers that can be used as the destination for rendering.

So a framebuffer _contains_ the color, depth and stencil buffers. Still, let's keep the "essentially a texture that can be drawn to" because that's much easier to understand for newbies :)

website/docs/opengl/c1/2-hello-quad.md
Perksey: This is probably the most obvious reason yeah.

website/docs/opengl/c1/2-hello-quad.md
ThomasMiz:
```suggestion
On top of this, a window will often contain at least two sets of these buffers. This is known as **double-buffering** and is imperitive for VSync to work properly. One buffer is displayed, while another is rendered to. 
```

website/docs/opengl/c1/2-hello-quad.md
```diff
@@ -0,0 +1,524 @@
...
+using System.Drawing;
+```
+
+Then, in your `OnLoad` method, add the following:
+
+```cs
+_gl.ClearColor(Color.CornflowerBlue);
+```
+
+You may notice that this function contains various **overloads**, including ones for a Vector4, and 4 floats. We'll be using the overload that takes `System.Drawing.Color`, because it is the easiest to use and understand.
```
ThomasMiz, line 82:
```suggestion
You may notice that this function contains various **overloads**, including ones for a `System.Numerics.Vector4`, and 4 floats. We'll be using the overload that takes `System.Drawing.Color`, because it is the easiest to use and understand.
```

website/docs/opengl/c1/2-hello-quad.md
```diff
@@ -0,0 +1,524 @@
...
+
+![Sky blue window](../../../images/opengl/chapter1/cornflower-window.png)
+
+Congrats! You've done your first thing in OpenGL! Didn't work? Check the [source code](../sources/1.2.2-clear-window.html) for this section here.
+
+But - how does OpenGL know to color the window blue? Sure, we've told it we want blue - but we only did that once... How does it know to keep using blue?
+
+Well, this is because OpenGL is a **state machine**.
+
+## OpenGL: The state machine
```
ThomasMiz, line 102: I like the smooth transition!

website/docs/opengl/c1/2-hello-quad.md
```diff
@@ -0,0 +1,524 @@
...
+At its core, a state machine holds... state. You set the state of something, and it retains that state until you change it.
+
+This is exactly how OpenGL works. Once you set something, it will remain set until you change it. This counts for everything in OpenGL. Clear color, binding objects, etc etc, everything goes through the state machine. You can *manipulate* the current state, however you have to be wary at all times of what part of the state you are changing. Change the wrong thing and suddenly you program might not work!
+
+So, this explains why clearing the window works. You set the clear color, and it remains *as* the clear color until you change it.
+
+Now that we've got that, let's move on!
+
+## Vertex Array Objects (VAOs)
+A unique feature that modern OpenGL has is what is known as a Vertex Array Object, or VAO. It stores the state of various parameters, such as which buffers you wish to use, and the shader inputs.
```
ThomasMiz, line 115:
```suggestion
A unique feature that modern OpenGL has is what is known as a Vertex Array Object, or VAO. It stores all of the state needed to supply vertex data, such as which buffers you wish to use and how the vertex data is read from said buffers.
```
I think this will be clearer, what do you think?

From [the khronos docs](https://www.khronos.org/opengl/wiki/Vertex_Specification):
> A Vertex Array Object (VAO) is an [OpenGL Object](https://www.khronos.org/opengl/wiki/OpenGL_Object) that stores all of the state needed to supply vertex data

website/docs/opengl/c1/2-hello-quad.md
```diff
@@ -0,0 +1,524 @@
...
+Then, in your `OnLoad` method, add:
+
+```cs
+_vao = _gl.GenVertexArray();
+_gl.BindVertexArray(_vao);
+```
+
+What's going on here? First, we generate the VAO. Then, before we can use, or update it, we need to **bind** it.
+
+Binding is a term you will hear a lot with OpenGL, and in these tutorials too. Binding essentially means updating the current OpenGL state with the given object. So, in this example, when we bind the VAO, we update OpenGL's state so that the VAO we are currently using is now the VAO we have just created. If you don't bind an object before updating/using it, you will use or update the previously bound object instead, which can cause problems and headaches later.
```
ThomasMiz, line 136:
```suggestion
Binding is a term you will hear a lot with OpenGL, and in these tutorials too. Binding essentially means updating the current OpenGL state with the given object. So, in this example, when we bind the VAO, we update OpenGL's state so that the VAO in the state machine is now the VAO we have just created. If you don't bind an object before updating/using it, you will use or update the previously bound object instead, which can cause problems and headaches later.
```

I think "We are currently using" can be confusing

website/docs/opengl/c1/2-hello-quad.md
```diff
@@ -0,0 +1,524 @@
...
+
+We're now ready to give the VAO some data. But first, we need a **vertex buffer**.
+
+## Vertex Buffer Objects (VBOs)
+Here is the part where we actually give the GPU some data it can work with. For 95% of graphics applications, a vertex buffer is **required** to display anything on screen.
```
ThomasMiz, line 141:
```suggestion
Here is the part where we actually give the GPU some data it can work with. For most graphics applications, a vertex buffer is **required** to display anything on screen.
```

website/docs/opengl/c1/2-hello-quad.md
```diff
@@ -0,0 +1,524 @@
...
+
+### What is a vertex buffer?
+Before we can continue, it's handy to know what a vertex buffer is, and what it does.
+
+Let's define both vertex, and buffer.
+
+* Vertex - A point where two lines, or edges meet.
```
ThomasMiz, line 148: (This is just a note, I'm not asking for this to be changed, I like it because it's simple to understand. I just want to comment this to make sure you are aware that the OpenGL concept of "vertex" is actually much more complex)

Note: A vertex in OpenGL isn't necessarily this. You can define the vertex attributes you want and use them however you wish to. For example, a vertex might have just a "texture coordinates" attribute, and then the vertex shader gets the position (to assign to `gl_Position`) from a texture. Another example is a small heightmap test project of mine, in which I had an empty VAO; No vertex attributes whatsoever! The vertex shader instead used `gl_VertexID` to organize the vertex into a flat XZ mesh pattern, and then the Y position came from reading the heightmap texture.

website/docs/opengl/c1/2-hello-quad.md
```diff
@@ -0,0 +1,524 @@
...
+{
+     0.5f,  0.5f, 0.0f,
+     0.5f, -0.5f, 0.0f,
+    -0.5f, -0.5f, 0.0f,
+    -0.5f,  0.5f, 0.0f
+};
+```
+
+### What makes up a quad?
+In modern graphics programming, you are expected to use triangles. While you may see options for quads, or lines, these are mostly deprecated.
```
ThomasMiz, line 169:
```suggestion
In modern graphics programming, you are expected to use triangles, lines, or points. While you may see options for quads, these are obsolete.
```

In [the knronos docs for primitives](https://www.khronos.org/opengl/wiki/Primitive), quads are deprecated as OpenGL 3.0, and have been removed from 3.1 and above. Lines and points are still part of the core specification.

website/docs/opengl/c1/2-hello-quad.md
```diff
@@ -0,0 +1,524 @@
...
+
+```
+_vbo = _gl.GenBuffer();
+_gl.BindBuffer(BufferTargetARB.ArrayBuffer, _vbo);
+```
+
+Much like the VAO, we need to first generate the buffer, and then bind it. Unlike vertex arrays, however, buffers need to be bound to a **target**. This allows you to bind buffers to different targets at the same time. Some of the more common targets include:
+
+* `ArrayBuffer` - The vertex buffer target (which is what we're using here).
+* `ElementArrayBuffer` - The element array buffer target (which you'll see later).
```
ThomasMiz, line 196: If you wish to make this more specific, here's [the doc on binding targets](https://www.khronos.org/opengl/wiki/Buffer_Object#General_use).

In summary:
- `ArrayBuffer`: Note that `glVertexAttribPointer()` doesn't have any way to specify which VBO to target! That's because it targets what's currently bound to `ArrayBuffer`.
- `ElementArrayBuffer`: Uhhhh this one's a bit of a pain. Unlike the other buffer targets, this one is not stored as a "general OpenGL state", but rather is part of the currently bound VAO. `ElementArrayBuffer` controls "which buffer to use as an index buffer for the currently bound VAO".

website/docs/opengl/c1/2-hello-quad.md
```diff
@@ -0,0 +1,524 @@
...
+* `UniformBuffer` - A uniform buffer, not used in this tutorial, but we'll get to it in a later tutorial.
+
+In this case, we're binding to `ArrayBuffer` since we're creating a vertex buffer.
+
+### Filling our buffer with data
+Let's fill our buffer with some data! Before we do that though, you need to be aware of `unsafe` in C#.
+
+#### Unsafe C#
+Silk.NET heavily uses `unsafe` code. Don't worry, this won't make your computer explode, however it does exit out of the "safe" managed environment of C#, and enters a realm where undefined behaviour, segmentation faults, and strange results are more likely to occur if you are not careful.
```
ThomasMiz, line 205: > Don't worry, this won't make your computer explode

probably****

website/docs/opengl/c1/2-hello-quad.md
```diff
@@ -0,0 +1,524 @@
...
+* `ArrayBuffer` - The vertex buffer target (which is what we're using here).
+* `ElementArrayBuffer` - The element array buffer target (which you'll see later).
+* `UniformBuffer` - A uniform buffer, not used in this tutorial, but we'll get to it in a later tutorial.
+
+In this case, we're binding to `ArrayBuffer` since we're creating a vertex buffer.
+
+### Filling our buffer with data
+Let's fill our buffer with some data! Before we do that though, you need to be aware of `unsafe` in C#.
+
+#### Unsafe C#
```
ThomasMiz, line 204: Good catch! Explaining what `unsafe` is and why it's necessary is important.

website/docs/opengl/c1/2-hello-quad.md
ThomasMiz, line 205:
```suggestion
Silk.NET heavily uses `unsafe` code. Don't worry, this won't make your computer explode, however it does exit out of the "memory safe" managed environment of C#, and enters a realm where undefined behavior, segmentation faults, and strange results are more likely to occur if you are not careful. Since we're working with low-level APIs, and OpenGL is defined in plain C, some unsafe code will be necessary in order for us to be able to communicate with it from C#.
```

website/docs/opengl/c1/2-hello-quad.md
```diff
@@ -0,0 +1,524 @@
...
+Silk.NET heavily uses `unsafe` code. Don't worry, this won't make your computer explode, however it does exit out of the "safe" managed environment of C#, and enters a realm where undefined behaviour, segmentation faults, and strange results are more likely to occur if you are not careful.
+
+<?# Info "If you wish to use `Span` instead, and remain in `safe` mode, Silk.NET does support these too. However, I will be using `unsafe` in this tutorial instead, as this is both what I personally use, as well as what the samples use." /?>
+
+Unsafe mode is not enabled by default, so we need to enable it. To enable it, add `<AllowUnsafeBlocks>true</AllowUnsafeBlocks>` to the `PropertyGroup` in your project's `.csproj` file. You also **MUST** add `unsafe` to your `OnLoad` and `OnRender` methods, like so:
```
ThomasMiz, line 209:
```suggestion
Unsafe mode is not enabled by default, so we need to enable it. To enable it:
- If you're on Visual Studio 2022, open your project properties and under Build --> General, make sure the box that says "Unsafe code" is checked.
- You can also edit your project's `.csproj` file manually, by adding `<AllowUnsafeBlocks>true</AllowUnsafeBlocks>` to the `PropertyGroup`.

You also **MUST** add `unsafe` to your `OnLoad` and `OnRender` methods, like so:
```

website/docs/opengl/c1/2-hello-quad.md
Beyley: are we sure we want american english here? there's [precedent](https://github.com/dotnet/Silk.NET/blob/93f987aa40f99c3b6e347d3863fe2e524c2a9398/examples/CSharp/OpenGL%20Tutorials/Tutorial%203.2%20-%20Diffuse%20Lighting/Program.cs#L122) for both british and american english

website/docs/opengl/c1/2-hello-quad.md
```diff
@@ -0,0 +1,524 @@
...
+
+```cs
+public static unsafe void OnLoad() { ... }
+```
+
+If you forget to do either of these, you'll get compile errors. While our `OnRender` method does not contain any unsafe code yet, it will when we draw to the screen later, so it's best just to do it now so you don't forget to do it later.
+
+Now let's fill the buffer!
+
+Add the following after you create the buffer:
```
ThomasMiz, line 219:
```suggestion
Add the following code to your `OnLoad` after you create the buffer:
```

website/docs/opengl/c1/2-hello-quad.md
```diff
@@ -0,0 +1,524 @@
...
+
+```cs
+fixed (float* buf = vertices)
+    _gl.BufferData(BufferTargetARB.ArrayBuffer, (nuint) (vertices.Length * sizeof(float)), buf, BufferUsageARB.StaticDraw);
+```
+
+Let's go over what's going on here. First, we `fix` the vertices data. This prevents the garbage collector from moving it around, so we can take a pointer to the data. We tell it the target we want, the `ArrayBuffer` in this case, give it the data length, the buffer pointer, and we choose `StaticDraw` for our usage.
```
ThomasMiz, line 226:
```suggestion
Let's go over what's going on here. First, we `fix` the vertex data. This prevents the garbage collector from moving it around, so we can take a pointer to the data. We then call `_gl.BufferData` and tell it the target we want, `ArrayBuffer` in this case, give it the data length (measured in bytes), the buffer pointer, and we choose `StaticDraw` for our usage hint.
```

website/docs/opengl/c1/2-hello-quad.md
ThomasMiz: I'm just gonna link [this commit](https://github.com/dotnet/Silk.NET/commit/4ae5a303e4238f3ed4fae199b6e64bb5ab2c0d0c)...

website/docs/opengl/c1/2-hello-quad.md
```diff
@@ -0,0 +1,524 @@
...
+
+Now, run the program again, and if you get the same blue window, you've set the buffer data successfully!
```
ThomasMiz, line 228: (Not necessarily true, the blue window will remain whether you did it right or wrong)

website/docs/opengl/c1/2-hello-quad.md
Beyley: ah, so is american english the standard in Silk? if so we should really do another British ->American english pass (maybe we should write something into the bakery to convert binded function names and enum names to american english :^)

website/docs/opengl/c1/2-hello-quad.md
```diff
@@ -0,0 +1,524 @@
...
+
+Now, let's create the **element buffer object**.
+
+## Element Buffer Objects (EBOs)
+Unlike a vertex buffer, an element buffer is not *strictly* required to display something on screen, however not using it will require making your vertex buffer a lot larger, with a lot more duplicate data.
```
ThomasMiz, line 233:
```suggestion
Unlike a vertex buffer, an element buffer is not *strictly* required to display something on screen, however not using it in some cases may require making your vertex buffer a lot larger, with a lot more duplicate data.
```

Just because I usually don't need to use EBOs lol

website/docs/opengl/c1/2-hello-quad.md
Beyley: > (Note: Not necessarily true, the blue window will remain whether you did it right or wrong)

The program may also segfault/throw an exception (along with the driver doing who knows what when a GL usage error occurs)

website/docs/opengl/c1/2-hello-quad.md
ThomasMiz: Wouldn't hurt to do so, but I don't think it's all that important either. We could always just ctrl-F "colour" through the entire project.

That commit still makes me chuckle lol

website/docs/opengl/c1/2-hello-quad.md
Beyley: Should we elaborate on what `StaticDraw` means? perhaps a list with small descriptions should be somewhere?

website/docs/opengl/c1/2-hello-quad.md
ThomasMiz: True that, but looking at the code it's waaaaay more likely that if you did something wrong it's gonna be something like forgetting to bind something, in which case it'll just silently fail.

I was just pointing it out though, I don't think it's wrong if that line remains as is

website/docs/opengl/c1/2-hello-quad.md
```diff
@@ -0,0 +1,524 @@
...
+
+Add the following to your `OnLoad` method:
+
+```cs
+uint[] indices =
+{
+    0u, 1u, 3u,
+    1u, 2u, 3u
+};
+```
```
ThomasMiz, line 249: Any particular reason for using `uint` for indices? I usually use `ushort` or if that's not enough `uint`, given that those are _always_ 16-bit and 32-bit respectively

Beyley: I think its best to suggest `ushort` by default, as its half the data

website/docs/opengl/c1/2-hello-quad.md
```diff
@@ -0,0 +1,524 @@
...
+* Fragment (or Pixel, as used by Direct3D)
+* Geometry
+* Compute
+* Tessellation
+
+These shaders all serve different purposes, however the most commonly used shaders are the **vertex** and **fragment** shaders, and we will be using those in this tutorial.
+
+A vertex shader is run (invoked) for every vertex in the vertex buffer. The vertex buffer is where you perform transformations, such as world and camera matrices (which we'll get into in a later tutorial).
+
+The fragment shader is invoked for every fragment of every vertex on screen. A fragment is essentially a pixel, and Direct3D even calls them pixel shaders. As you may expect, these shaders are a lot more intensive than vertex shaders, and can often be the cause of GPU slowdowns. Fragment shaders is where you perform the stuff that gets displayed on screen, such as texturing and many forms of lighting.
```
ThomasMiz, line 298:
```suggestion
A vertex shader is run (invoked) for every vertex in the vertex buffer. The vertex buffer is where you perform transformations such as translating, rotating or scaling and object. Once the vertex shaders are done running, our vertices are arranged into **primitives** (in our case, triangles), and the on-screen pixels these triangles occupy are filled by fragment shaders. This process is called **rasterization**.

The fragment shader is invoked for every fragment of every primitive on screen. A fragment is essentially a pixel, and Direct3D even calls them pixel shaders. As you may expect, these shaders are a lot more intensive than vertex shaders, and can often be the cause of GPU slowdowns, given that a single triangle may only invoke three vertex shaders, but can result in hundreds or thousands of fragment shader invocations!

Fragment shaders is where you perform the stuff that gets displayed on screen, such as texturing and many forms of lighting.
```

To keep the VS explanation simpler, and be more correct with the FS explanation

website/docs/opengl/c1/2-hello-quad.md
```diff
@@ -0,0 +1,524 @@
...
+
+For the moment, this is all you need to know about shaders. You will learn more about shaders as we progress further in this tutorial, as knowing how shaders work and operate is a vital thing to know when it comes to modern graphics programming. The sky is the limit! (Or in our case, an endless cornflower blue void...)
```
ThomasMiz, line 300: > The sky is the limit! (Or in our case, an endless cornflower blue void...)

lol

ThomasMiz: _"Sometimes, I feel like my entire life is just one big static void..."_

website/docs/opengl/c1/2-hello-quad.md
```diff
@@ -0,0 +1,524 @@
...
+{
+    gl_Position = vec4(aPosition, 1.0);
+}";
+```
+
+This is our vertex shader! Let's go through it line-by-line.
+
+First, we must tell it which GLSL version we wish to use. Since OpenGL 3.3 (the version we are using), the GLSL version corresponds to the OpenGL version. We tell it we want `core`, as we are using Core OpenGL 3.3, rather than compatibility.
+
+Next, we define our **shader attributes**. In our example, we only have one, which is the position we defined in the vertex buffer. Remember - the vertex buffer is executed *for each* vertex in the vertex buffer, which is why we only need one value at a time. We define it at a manual "location", 0. While this is not necessary, we will be using it in this tutorial, as certain drivers (such as intel) like to use random locations if you do not specify one, which can cause many problems if you are not careful.
```
ThomasMiz, line 325:
```suggestion
Next, we define our **shader attributes**. In our example, we only have one, which is the position we defined in the vertex buffer. Remember - the vertex shader is executed *for each* vertex in the vertex buffer, which is why we only need one value at a time. We define it at a manual "location", 0. While this is not necessary, we will be using it in this tutorial, as certain drivers (such as intel) like to use random locations if you do not specify one, which can cause many problems if you are not careful.
```

website/docs/opengl/c1/2-hello-quad.md
```diff
@@ -0,0 +1,524 @@
...
+
+Much like C, we define our entry point, the `main` function. In it, we set the value of `gl_Position` to the vertex position. As we are not performing any transformations, we simply set the input value directly to the output value. `gl_Position`, which is a built-in shader variable, only accepts a `vec4` value though, so we must convert `aPosition` to `vec4` first. You may notice we use `1.0` instead of `0.0` though. While in this case it does not matter, it's a habit you should get into, as this will become very important when we start using matrices and transformations.
```
ThomasMiz, line 327:
```suggestion
Much like C, we define our entry point, the `main` function. In it, we set the value of `gl_Position` to the vertex position. As we are not performing any transformations, we simply set the input value directly to the output value. `gl_Position`, which is a built-in shader variable, only accepts a `vec4` value though, so we must convert `aPosition` to `vec4` first. You may notice we use `1.0` for the W value instead of `0.0` though. While in this case it does not matter, it's a habit you should get into, as this will become very important when we start using matrices and transformations.
```

website/docs/opengl/c1/2-hello-quad.md
Perksey: Yeah, this is a liability risk!

Perksey: CAN WE PLEASE USE BRITISH ENGLISH PLEASE IT MAY NOT BE NORMAL IN OUR CODE BUT IT IS DEFINITELY THE PRECEDENT IN OUR LITERATURE SEE ALL OF THE BLOG POSTS USING BRITISH ENGLISH SURE THAT'S BECAUSE I WROTE THEM BUT THE POINT STILL STANDS BRITISH ENGLISH PLEASE THANK YOU THE ONLY REASON WE DON'T USE IT IN CODE IS BECAUSE WE'RE FOLLOWING MICROSOFT'S PRECEDENT THERE KTHX BYE

website/docs/opengl/c1/2-hello-quad.md
```diff
@@ -0,0 +1,524 @@
...
+const string fragmentCode = @"
+#version 330 core
+
+out vec4 out_color;
+
+void main()
+{
+    out_color = vec4(1.0, 0.5, 0.2, 1.0);
+}";
+```
```
ThomasMiz, line 344:
```suggestion
```cs
const string fragmentCode = @"
#version 330 core

out vec4 outColor;

void main()
{
    outColor = vec4(1.0, 0.5, 0.2, 1.0);
}";
```
```

To keep consistent naming with the variables in the vertex shader?

I usually name this one `FragColor`. Do as you please tho

File: website/docs/opengl/c1/2-hello-quad.md [hunk 37]
```diff
@@ -0,0 +1,524 @@
...
+```cs
+const string fragmentCode = @"
+#version 330 core
+
+out vec4 out_color;
+
+void main()
+{
+    out_color = vec4(1.0, 0.5, 0.2, 1.0);
+}";
```
ThomasMiz, line 343:
```suggestion
const string fragmentCode = @"
#version 330 core

out vec4 outColor;

void main()
{
    outColor = vec4(1.0, 0.5, 0.2, 1.0);
}";
```
To keep consistent naming with the variables in the vertex shader?

I usually name this one `FragColor`. Do as you please tho

website/docs/opengl/c1/2-hello-quad.md
```diff
@@ -0,0 +1,524 @@
...
+if (status != (int) GLEnum.True)
+    throw new Exception("Vertex shader failed to compile: " + _gl.GetShaderInfoLog(vertexShader));
+```
+
+First, we compile the shader. Then, we need to check to make sure it has compiled correctly, and if it hasn't, it needs to let us know what our error is. `glGetShader` can give us a lot of different parameters, some of which we'll be using in later tutorials. For now, we are just interested in the compile status.
+
+The compile status returns `1` if successful, and `0` if not. If it's not successful, we throw an exception. `glGetShaderInfoLog` gets the error string for the shader, and we output that in our exception. You don't *have* to throw an exception though, you can do whatever you wish with this result. Just don't try to continue further, as you will get a link error! (We'll get to this in a minute).
+
+We've now successfully created and compiled our vertex shader. To compile our fragment shader, simply copy and paste the code you just wrote. Just make sure to change the `ShaderType` to `FragmentShader`!
+
```
ThomasMiz, line 379:
```suggestion
We've now successfully created and compiled our vertex shader. To compile our fragment shader, simply copy and paste the code you just wrote. Just make sure to change the `ShaderType` to `FragmentShader`!

```cs
uint fragmentShader = _gl.CreateShader(ShaderType.FragmentShader);
_gl.ShaderSource(fragmentShader, fragmentCode);
_gl.CompileShader(fragmentShader);

_gl.GetShader(fragmentShader, ShaderParameterName.CompileStatus, out int status);
if (status != (int) GLEnum.True)
    throw new Exception("Fragment shader failed to compile: " + _gl.GetShaderInfoLog(fragmentShader));
```

```

File: website/docs/opengl/c1/2-hello-quad.md [hunk 39]
```diff
@@ -0,0 +1,524 @@
...
+Now we need to **compile** the shader.
+
+Add the following to your `OnLoad` method:
+
+```cs
+_gl.CompileShader(vertexShader);
+
+_gl.GetShader(vertexShader, ShaderParameterName.CompileStatus, out int status);
+if (status != (int) GLEnum.True)
```
ThomasMiz, line 370:
```suggestion
_gl.GetShader(vertexShader, ShaderParameterName.CompileStatus, out int vStatus);
if (vStatus != (int) GLEnum.True)
```

to match the resulting code from 1.2-final-result

website/docs/opengl/c1/2-hello-quad.md
```diff
@@ -0,0 +1,524 @@
...
+_gl.AttachShader(_program, fragmentShader);
+
+_gl.LinkProgram(_program);
+
+_gl.GetProgram(_program, ProgramPropertyARB.LinkStatus, out int lStatus);
+if (lStatus != (int) GLEnum.True)
+    throw new Exception("Program failed to link: " + _gl.GetProgramInfoLog(_program));
+```
+
+Much like before, we must check to make sure that the program has linked correctly. Link errors are rare, most often caused by attempting to link with a shader that has failed to compile, which, assuming you implemented the checks earlier, should basically never happen.
```
ThomasMiz, line 410:
```suggestion
Much like before, we must check to make sure that the program has linked correctly. Link errors are most often caused by things like a typo in an `in` or `out` variable name, or attempting to link with a shader that has failed to compile, which, assuming you implemented the checks earlier, should basically never happen.
```

At least from my experience lol

website/docs/opengl/c1/2-hello-quad.md
```diff
@@ -0,0 +1,524 @@
...
+<?# Warning "While fragment shaders (and other shaders) can have `in` attributes, the only ones you can directly set *outside* of a shader are the ones going into the vertex shader. The only way to set attributes in the fragment shader is to pass them through the vertex shader. Therefore, the attribute setup only affects the vertex shader and vertex buffer, not any other shader." /?>
+
+Add the following to your `OnLoad` method:
+
+```cs
+_gl.UseProgram(_program);
+
+const uint positionLoc = 0;
+_gl.EnableVertexAttribArray(positionLoc);
+_gl.VertexAttribPointer(positionLoc, 3, VertexAttribPointerType.Float, false, 3 * sizeof(float), (void*) 0);
```
ThomasMiz, line 437: This code is related to the VAO. Readers need to understand that this code is configuring the VAO, and not the shader program.

This code is usually written together with the creation of the VAO.

website/docs/opengl/c1/2-hello-quad.md
ThomasMiz: Okay yes, I'm gonna ask you to change it because afterwards you have to specify the element type, and you do so as `UnsignedInt`, therefore you need to make sure that your indices are actually 32-bit.

You can change it to `uint` if you wish, I'd still recommend `ushort`. You'd also have to update part of the "Drawing to the screen" section.

website/docs/opengl/c1/2-hello-quad.md
```diff
@@ -0,0 +1,524 @@
...
+
+## Drawing to the screen
+It's finally time to draw to the screen! That was a lot of setup work we just did there. Fortunately, that was the hard part. Drawing our result to the screen is now very easy.
+
+Add the following to your `OnRender` method:
+
+```cs
+_gl.BindVertexArray(_vao);
+_gl.UseProgram(_program);
+_gl.DrawElements(PrimitiveType.Triangles, 6, DrawElementsType.UnsignedInt, null);
```
ThomasMiz, line 481:
```suggestion
_gl.DrawElements(PrimitiveType.Triangles, 6, DrawElementsType.UnsignedInt, (void*) 0);
```

I think it's better if we avoid `null` on these numeric values. It is confusing enough that numeric values have to be passed as pointers, even though they don't point to anything! I hope for 3.0 we can turn all of these into `uint`s or something ;_;

website/docs/opengl/c1/2-hello-quad.md
```diff
@@ -0,0 +1,524 @@
...
+
+While this may have seen like a lot of set up for a boring result, this code can render pretty much anything you want to the screen. It remains pretty much the same, whether you're rendering a basic quad like this, or a complex 3D model. All you need to change are the vertices & indices going in, and some more complex shader code to handle the transformations.
+
+Once you get the basics understood and you can confidently use it, creating more complex stuff becomes a lot easier.
+
+If you would like to view the full source code for the application, you can view it [here](../sources/1.2-final-result.html).
+
+## Cleaning up
+When you're done with an OpenGL resource, you should delete it, to free GPU memory. For this though, it's not needed. The driver will automatically clean all created resources when your application closes. In fact, it's recommended that you *don't* manually delete resources when the application closes. The driver can free these objects a lot faster than manually removing them. If you have a lot of objects, manually deleting can cause the application to hang while the driver tries to free everything.
```
ThomasMiz, line 511: > The driver can free these objects a lot faster than manually removing them

Interesting, I did not know 🤔🤔

website/docs/opengl/c1/2-hello-quad.md
```diff
@@ -0,0 +1,524 @@
...
+
+You **MUST** remember to remove unused objects while your application is running, however, as forgetting to do so will use more and more of the GPU's memory, potentially causing it to run out of memory. OpenGL isn't garbage collected! It requires manual memory management, a lot like C.
+
+## Wrapping up
+That was a lot to digest! You've hopefully learned a lot along the way though, and if there's bits you still don't understand, just go back and re-read the sections again to hopefully improve your understanding. Don't worry about it too much though, the learning curve is steep, and once you do understand it, you'll be writing graphics programs in no time!
+
+Here's some stuff you can do now:
+
+* Move on to the [next tutorial](), where you'll learn how to create a GL context and display a quad on the screen.
```
ThomasMiz, line 520: Copy paste typo?

ThomasMiz:
Excellent work on the hello quad tutorial! I like it a lot. The tutorial is mostly straightforward, but mentions technical bits and pieces along the way as to not turn into a "just copy code, don't understand it". There is good pacing by interlacing code and theory.

I've come to appreciate how long these can take to write when I started writing some myself, so we're very grateful you've taken the time to make these.

website/docs/opengl/c1/2-hello-quad.md
ohtrobinson: It is clearer, my only concern is that it may confuse some people who don't understand what vertices are, since that hasn't been covered at this part of the tutorial - I'll include something along these lines but I'll try to make it easier to understand for newbies, perhaps something like `It stores all the necessary information required to draw an object to the screen, such as the vertex data (which we'll get into in a minute), and also information on how to process this data.`. How does that sound?

website/docs/opengl/c1/2-hello-quad.md
ohtrobinson: I agree, although personally I prefer `the VAO currently stored in the state machine` as it just makes it a bit clearer that the state machine holds this specific VAO we have just set. But it's a minor change and I think the suggested change is still perfectly understandable. I'll commit your change and can I always change it later.

website/docs/opengl/c1/2-hello-quad.md
ohtrobinson: Yeah, and this may be explored in later tutorials, I've been meaning to get into heightmaps and stuff myself too. For the purposes of this tutorial though, the vertex definition used here is helpful enough for the reader to understand.

website/docs/opengl/c1/2-hello-quad.md
ohtrobinson: I personally think this is a bit too specific to put in this tutorial, however I could add a note to the shader attributes section with the `glVertexAttribPointer` information, as it seems useful to know.

website/docs/opengl/c1/2-hello-quad.md
ohtrobinson: happy to change to british english if desired

website/docs/opengl/c1/1-hello-window.md
ohtrobinson: I did indeed!

website/docs/opengl/c1/2-hello-quad.md
ohtrobinson: I personally use uint for my indices which is why I used it here, but it makes little difference in practice. The main reason I use uint is to support more than 65k vertices, which some of the models I import have. Personally for the purposes of the tutorial I don't think it makes much difference but I'm happy to change it.

website/docs/opengl/c1/2-hello-quad.md
ohtrobinson: thanks for the VS suggestions, I have literally 0 idea how VS works so I just don't include it in the tutorials

website/docs/opengl/c1/2-hello-quad.md
ohtrobinson: Yeah you're right, hadn't really thought about it outside of vsync. I'll change this to make it clearer tomorrow.

website/docs/opengl/c1/2-hello-quad.md
ohtrobinson: I agree about changing it to a "set of buffers", that is an oversight on my part. However I dislike calling them "default framebuffers", since, while yes the window has a default framebuffer, I do mention (i think?) that you can't directly interact with the window's framebuffer. And for example in Direct3D you *can* interact with the window's framebuffer, I prefer to say that the window consists of several framebuffers instead of *default* framebuffer. (I could go on with a long rant about the fact that you can't directly interact with the window's framebuffer... just give me the damn framebuffer so I can mess with it...!)

website/docs/opengl/c1/2-hello-quad.md
ohtrobinson:
```suggestion
Add the following to your `OnLoad` method:
```

I "standardized" this in my writing a while ago, I just forgot to change it earlier on. I will commit this and double check it's all standardized when I read through it and make some adjustments tomorrow.

website/docs/opengl/c1/2-hello-quad.md
ohtrobinson: Yeah, I'll add a small description of what static draw means, along with dynamic draw and stream draw.

I don't remember the others and I've never used them so I'll probably skip their explanation for this tutorial.

The funny thing is that most GL drivers don't even care what usage hint you give them, it's pretty much all the same now (from what I've heard).

website/docs/opengl/c1/2-hello-quad.md
ohtrobinson: The amount of times I have had driver crashes, segfaults, etc from failing to call `BufferData` correctly leads me to believe that if you do it wrong the program will crash.

You are right though that if the reader forgot to bind the buffer then they won't know, but that's why I include the "compare your code with the result" sections.

website/docs/opengl/c1/2-hello-quad.md
ohtrobinson: yeah, this is much better.

website/docs/opengl/c1/2-hello-quad.md
ohtrobinson: Yeah I have a strict naming style I stuck with long ago:

* Attributes - `aAttributeName`
* Uniforms - `uUniformName`
* Out variables - `shadername_variableName` (for example `frag_texCoord`)
* Fragment shader out variables - `out_variableName`

Not sure if there's some "recommended" naming scheme but imo this makes it very easy to understand what each variable does and is doing.

website/docs/opengl/c1/2-hello-quad.md
ohtrobinson: I decided not to include the fragment shader code as it *should* be fairly obvious to the reader what needs to be copy-pasted and changed (as I even prompt them to do so and what to change), so I'd rather not include the duplicate code, unless you really think I should. If they get really stuck they could just look at the final result code.

website/docs/opengl/c1/2-hello-quad.md
ohtrobinson: Hrm. I virtually always see this code right at the bottom of the vao setup, including in the official silk example. I didn't know that it was configuring specifically the VAO though, since I remember having issues with this if I didn't have a shader program bound, which made me assume it was configuring both.

(As a side note, personally, I absolutely despise VAOs and GL's attribute setup, I don't even use them outside of a single vao for the entire program. Give me input layouts any day of the week...)

website/docs/opengl/c1/2-hello-quad.md
ohtrobinson: Yeah it would definitely be good to have it as a uint, or at least an overload for it. I often (including myself when I switched to silk) see people using `0` by itself and then being confused as to why it didn't work. While I personally use `null`, for the purposes of the tutorial I agree that `(void*) 0` is better. I'll leave this unresolved for now so I remember to update all the files containing this code (and not just the tutorial) when I make these changes tomorrow.

website/docs/opengl/c1/2-hello-quad.md
ohtrobinson: Whoops, yes, I didn't even notice it when I read it just now, lol.

ohtrobinson:
I appreciate the comments & suggestions! And yeah, it was a lot more effort than I was expecting too. Many tutorials just seem to be the "here copy this code and it works and you don't need to know why" sorta thing and I absolutely hate that, so I made sure to explain exactly what was going on.

I've committed most of the suggestions, I've left a couple open since I won't commit that *exact* suggestion but I'll write close to it, and a couple have been left open as I don't agree with a couple of changes, but if you wish I am happy to change them too.

I'll make the finishing touches tomorrow, and then if all is good it should be complete!

website/docs/opengl/c1/2-hello-quad.md
ThomasMiz: I think you should include it, mostly for those that are just copy-pasting, but also to show where the `fragmentShader` variable comes from.

website/docs/opengl/c1/2-hello-quad.md
ThomasMiz: Yup, `EnableVertexAttribArray` and `VertexAttribPointer` apply config changes to the current VAO. Nothing to do with the current shader program.

The only thing that connects the VAO to the shader program are the attribute locations. In this case, it's just a single attribute, at location 0, of type vec3.

(What?? You use a single VAO for your entire program?? If you have to go around enabling and disabling attributes and calling `VertexAttribPointer` whenever you need to change vertex type or source VBOs, that's very inefficient, and you should use multiple VAOs)

website/docs/opengl/c1/2-hello-quad.md
ThomasMiz: Ah, I usually name them like so:
- VS Input attributes: `vAttributeName`
- VS output or FS input attributes: `fAttributeName`
- FS output color attribute: `FragColor`
- Uniforms: just `UniformName`

I don't know any specific convention though. Do as you please.

website/docs/opengl/c1/2-hello-quad.md
ohtrobinson: Alrighty, I'll include it in the tutorial

website/docs/opengl/c1/2-hello-quad.md
ThomasMiz: Really? I don't remember ever being annoyed by segfaults or driver crashes from calling `BufferData`. My most common mistake is probably getting the size wrong, by forgetting to multiply by sizeof(whatever), which wouldn't cause any issues other than leaving part of the buffer with garbage. I do see why it could be a common problem tho

website/docs/opengl/c1/2-hello-quad.md
ThomasMiz: Ah sorry my bad; I got confused and somehow thought you were using native-sized integers (nuint). I'm a big dum dum. I'll mark this as resolved

website/docs/opengl/c1/2-hello-quad.md
ThomasMiz: Fair enough lol. _Damn you, window framebuffers!_

website/docs/opengl/c1/2-hello-quad.md
ThomasMiz: Not sure about the part on "how to process this data". All a VAO stores is the "how to read the vertex data". How to process is part of the shader program, which VAOs know nothing about.

website/docs/opengl/c1/2-hello-quad.md
ohtrobinson: I've removed the bits about vsync, and replaced it to say "imperative for rendering to work properly."

website/docs/opengl/c1/2-hello-quad.md
ohtrobinson: whoops, yes, this has now been fixed.

website/docs/opengl/c1/2-hello-quad.md
ohtrobinson: I do yeah, probably could use multiple VAOs though, I'll have to experiment with it.

I've changed the explanation to make it clear that its updating the VAO specifically (and also removed the bit that tells you to set the shader program)

> The last thing we need to do before we can begin drawing our quad to the screen is to set up the attributes. By default, OpenGL does not know how to send the data to our shader, so we must tell it.

> This is done with the **attribute setup**. This sets various parameters in the VAO, which tells OpenGL how to read the vertex data in the vertex buffer. As vertex buffers can contain more than just position data (it can contain almost anything you want), it is vital that OpenGL knows how to separate out the individual bits of data, so it can pass it to the shader correctly.

Hopefully this is okay?

ohtrobinson:
Seems like everything has now been resolved, so if all is good then this this PR should be ready.

I'll get to the next set of tutorials soon and submit a new PR for those. Trying to write at least two at a time, depending on their complexity, and more if possible.

ThomasMiz (approved):

Issue #1149, opened 2022-11-17
alBufferData has no overload for FloatBufferFormat


tilkinsc:
# Summary of feature
AL::BufferFormat should have an overload which accepts FloatBufferFormat (when a wav or ogg or etc has float data). 

# Comments
https://github.com/openalext/openalext/blob/master/AL_EXT_float32.txt
This is a feature request for Silk.NET.OpenAL.Extensions.EXT.

# Does this have a proposal?
No.

Beyley:
Since the OpenAL bindings are manually generated, you should add an extension method in [here](https://github.com/dotnet/Silk.NET/blob/main/src/OpenAL/Extensions/Silk.NET.OpenAL.Extensions.EXT/FloatFormat.cs), that just casts down to the base enum

tilkinsc:
So something like this?
```cs

namespace Silk.NET.OpenAL.Extensions.EXT
{
    /// <summary>
    /// Exposes the multi-channel buffers extension by Creative Labs.
    /// </summary>
    [Extension("AL_EXT_float32")]
    [NativeApi(Prefix = "al")]
    public partial class FloatFormat : FormatExtensionBase<FloatBufferFormat>
    {
        /// <inheritdoc cref="ExtensionBase" />
        public FloatFormat(INativeContext ctx)
            : base(ctx)
        {
        }
        
        public static implicit operator BufferFormat(FloatFormat format) => (BufferFormat) format;
}
```

Beyley:
thats not exactly what i meant, i meant an extension method over the methods that can take FloatFormat, not an operator override from FloatFormat -> BufferFormat

ohtrobinson:
This is part of the Silk.NET.OpenAL.Extensions.EXT nuget. You can just cast `FloatBufferFormat.XYZ` to `BufferFormat` when calling `BufferData()` and it should work fine.

tilkinsc:
> This is part of the Silk.NET.OpenAL.Extensions.EXT nuget. You can just cast `FloatBufferFormat.XYZ` to `BufferFormat` when calling `BufferData()` and it should work fine.

alBufferData will accept the enums, therefore it shall without explicit casts.

> thats not exactly what i meant, i meant an extension method over the methods that can take FloatFormat, not an operator override from FloatFormat -> BufferFormat

```cs
namespace Silk.NET.OpenAL.Extensions.EXT
{
    /// <summary>
    /// Exposes the multi-channel buffers extension by Creative Labs.
[...]

PR #1154, opened 2022-11-17
Implement extensions functions to make use of AL_EXT_float32 enums

tilkinsc:
# Summary of the PR
Currently you have to explicitly cast FloatBufferFormat to BufferFormat to use the AL::BufferFormat function. alBufferFormat takes in two new enums provided by AL_EXT_float32  to provide the ability to load floating point data into openal buffer objects.

Issue Ref: #1149
[Spec Ref](https://github.com/openalext/openalext/blob/master/AL_EXT_float32.txt)

# Further Comments
There may be a possibility that the Capture feature of OpenAL supports floating point audio. I have no mind to include this because it adds more research that isn't easily presented in google. Silk.NET doesn't compile too well either. So, to keep it simple I won't add anything to do with anything other than what the spec AL_EXT_float32 says.

### DIFF
--- a/src/OpenAL/Extensions/Silk.NET.OpenAL.Extensions.EXT/FloatFormat.cs
+++ b/src/OpenAL/Extensions/Silk.NET.OpenAL.Extensions.EXT/FloatFormat.cs
@@ -26,12 +26,12 @@ namespace Silk.NET.OpenAL.Extensions.EXT
     /// </summary>
     public static class FloatFormatALExtensions
     {
-        public static unsafe partial void BufferData(this AL al, uint buffer, FloatBufferFormat format, void* data, int size, int frequency)
+        public static unsafe void BufferData(this AL al, uint buffer, FloatBufferFormat format, void* data, int size, int frequency)
         {
             al.BufferData(buffer, (BufferFormat) format, data, size, frequency);
         }
 
-        public void BufferData<TElement>(this AL al, uint buffer, FloatBufferFormat format, TElement[] data, int frequency)
+        public static void BufferData<TElement>(this AL al, uint buffer, FloatBufferFormat format, TElement[] data, int frequency)
             where TElement : unmanaged
         {
             al.BufferData<TElement>(buffer, (BufferFormat) format, data, frequency);
--- a/src/OpenAL/Extensions/Silk.NET.OpenAL.Extensions.EXT/FloatFormat.cs
+++ b/src/OpenAL/Extensions/Silk.NET.OpenAL.Extensions.EXT/FloatFormat.cs
@@ -20,4 +20,21 @@ namespace Silk.NET.OpenAL.Extensions.EXT
         {
         }
     }
+
+    /// <summary>
+    /// Extends the AL functions abilities to make use of the float enums provided by AL_EXT_float32.
+    /// </summary>
+    public static class FloatFormatALExtensions
+    {
+        public static unsafe partial void BufferData(this AL al, uint buffer, FloatBufferFormat format, void* data, int size, int frequency)
+        {
+            al.BufferData(buffer, (BufferFormat) format, data, size, frequency);
+        }
+
+        public void BufferData<TElement>(this AL al, uint buffer, FloatBufferFormat format, TElement[] data, int frequency)
+            where TElement : unmanaged
+        {
+            al.BufferData<TElement>(buffer, (BufferFormat) format, data, frequency);
+        }
+    }
 }


Beyley (approved):

Beyley:
Want to get another maintainer's opinion before merging, as i dont think theres precedent for this kind of overload on extensions, and it only really applies to the manually generated bindings (like OpenAL)

HurricanKai (approved):

tilkinsc:
I want to add a forethought, it appears that OpenAL-soft doesn't provide functionality for AL_EXT_vorbis and AL_EXT_mp3.

https://github.com/kcat/openal-soft/blob/master/al/buffer.cpp
https://github.com/kcat/openal-soft/blob/master/include/AL/alext.h

So by all means only AL_EXT_float32, AL_EXT_IMA4, AL_EXT_double, AL_EXT_MULAW, AL_EXT_ALAW would ideally need toyed with in the future for alBufferData

THAT IS,

if there was some friggin documentation on what to pass INTO alBufferData. Do I give it the whole mp3 file?! We will never know.

Beyley:
> I want to add a forethought, it appears that OpenAL-soft doesn't provide functionality for AL_EXT_vorbis and AL_EXT_mp3.
> 
> https://github.com/kcat/openal-soft/blob/master/al/buffer.cpp https://github.com/kcat/openal-soft/blob/master/include/AL/alext.h
> 
> So by all means only AL_EXT_float32, AL_EXT_IMA4, AL_EXT_double, AL_EXT_MULAW, AL_EXT_ALAW would ideally need toyed with in the future for alBufferData
> 
> THAT IS,
> 
> if there was some friggin documentation on what to pass INTO alBufferData. Do I give it the whole [insert file format here] file?! We will never know.

@ohtrobinson you may be able to provide explination here on what `alBufferData` wants, i believe i know, but you may be able to explain better than i can, given you have more familiarity than i do

ohtrobinson:
I'm not too sure myself, I've not experimented with it before. I can't find where I heard it, but I believe that (taking vorbis as an example) it specifically wants vorbis *sound* data, and you need to parse the headers yourself. All it can do is process the given data into a PCM stream, and doesn't do any actual loading.

There really isn't that much about these extensions anywhere online, and as such I would recommend avoiding them, instead opting to use StbVorbis and an MP3 loader. At least then you can guarantee that it will work, as well as working on machines that don't support the extension.

I'm surprised OpenAL soft doesn't appear to support vorbis, but I sorta understand, it's a royal pain to work with, especially in the way OpenAL works.

tilkinsc:
SoLoud is implemented with loading as well as playing audio. I would trial it to be a useful addition to sit alongside OpenAL, being that it has the zlib (basically MIT) license. They sure as hell did a good job keeping it free software in that regard too.

SoLoud: zlib/libpng
OGG: public domain
MP3: Unlicense
Wav: Unlicense
Speech Synth: Public domain (including author changes)
Vizsn (speech synth): Public domain
FFT: public domain
Sfxr: zlib/libpng
RPGMaker Wrapper Generator: BSDL (never claimed to be part of Ruby) but actually public domain
TED/SID soundchip emulation: public domain
Vic soundchip emulation: zlib/libpng
Openmpt (completely optional): BSD-3 https://opensource.org/licenses/BSD-3-Clause
FreeVerb: public domain

http://solhsa.com/soloud/legal.html

Would this be an interesting proposal?

Perksey:
We generally try to stick to well-known, industry-standard APIs specified by authoritative bodies for silk-provided bindings, with the exception of the HLU which binds to whatever library necessary to deliver its managed abstractions.

The audio space is very fragmented, heck to be honest it’s questionable why we even have OpenAL bindings given it’s not nearly as recognised or operated in the same way as their graphics counterparts. As such, I do not believe Silk.NET would have SoLoud bindings given this seems like we’re “picking” a preferred technology, and making our selection of bindings more opinionated.

However, the SilkCommunity has no such rules. If someone made SoLoud bindings/wrappers/utilities using our tech (be that our generator tech and/or Silk.NET.OpenAL itself to “tie” those libraries together) I’m sure the SilkCommunity would love to see it.

https://GitHub.com/SilkCommunity/SilkCommunity

Beyley:
> We generally try to stick to well-known, industry-standard APIs specified by authoritative bodies for silk-provided bindings, with the exception of the HLU which binds to whatever library necessary to deliver its managed abstractions.
> 
> The audio space is very fragmented, heck to be honest it’s questionable why we even have OpenAL bindings given it’s not nearly as recognised or operated in the same way as their graphics counterparts. As such, I do not believe Silk.NET would have SoLoud bindings given this seems like we’re “picking” a preferred technology, and making our selection of bindings more opinionated.
> 
> However, the SilkCommunity has no such rules. If someone made SoLoud bindings/wrappers/utilities using our tech (be that our generator tech and/or Silk.NET.OpenAL itself to “tie” those libraries together) I’m sure the SilkCommunity would love to see it.
> 
> https://GitHub.com/SilkCommunity/SilkCommunity

To add to this, the generator is 95% of the way there to be able to generate SoLoud bindings (i believe there were a few random problems, but it mostly worked), so if someone wants to make them, @ me on discord, and ill help out best i can

PR #1157, opened 2022-11-23
Enable typemap-native for more clang bindings

Beyley:
# Summary of the PR
Enables the typemap-native control descriptor for D3D11, Win32Extras, DXGI, D3DCommon, DXVA, D2D, D3D12, D3D9, DXC, D3DCompilers, DirectStorage, XAudio, XInput, D2D1

# Further Comments
This mostly effects DWORD (which now gets generated as a `uint` instead of an `int`, as per the microsoft docs claim it is) and `BOOL`, which now gets mapped to our Bool32 type


Perksey (approved):
Given a quick skim of some of the changes and I'm not seeing anything I disagree with

PR #1184, opened 2022-12-26
WPF: Adding WGL bindings

Perksey:
Adds WGL bindings, needed for DX interop for WPF integration.


Perksey:
The WGL bit is done, but (as per Discord) @Beyley I could do with your hand in getting WGL working seeing as I don't have access to my dev PC until the new year.

Beyley (approved):
i see

PR #1185, opened 2022-12-28
Draft: Bring WASM support into Silk.NET

Beyley:
# Summary of the PR
Adds all the support logic and hacks to get it working, with a functional (albeit inadequate and in need of a split) example


Perksey:
I think we're all in agreement that this is a no-go until 3.0, despite the [successes](https://perksey.com/blog/blazor-silkdotnet.html) of #1746

PR #1193, opened 2022-12-30
Add d3d11on12 bindings

Beyley:
# Summary of the PR
Adds bindings for the D3D11on12 header

# Further Comments
D3D11 drivers on volterra are hot garbage, this may fix that


Perksey (approved):

Perksey:
NB: Your GPG signature is unverified.

Beyley:
> NB: Your GPG signature is unverified.

i commit this from my VM that i still havent setup a GPG key on because when i tried it made git segfault

Beyley:
Hmmm it seems this would make D3D11 have a hard dependency on D3D12, how is a way to go about this while still keeping the API good, if i put it into an extension would the VTable extensions still work?

Beyley:
Builds locally on my machine now as an extension

Perksey:
Where does the extension come from?

Perksey:
Oh nvm I see it, yeah this seems sensible.

Beyley:
CI hit an OOM, restarting

generator.json
```diff
@@ -1328,7 +1329,8 @@
                     "d3d11_4.h": "[Core]D3D11",
                     "d3d11shader.h": "[Core]D3D11",
                     "d3d11shadertracing.h": "[Core]D3D11",
-                    "d3d11sdklayers.h": "[Core]D3D11"
+                    "d3d11sdklayers.h": "[Core]D3D11",
+                    "d3d11on12.h": "[Core]D3D11"
```
Perksey, line 1333: You haven't staged generator.json

src/Microsoft/Extensions/Silk.NET.Direct3D11.Extensions.D3D11on12/Direct3D11on12.gen.cs
```diff
@@ -0,0 +1,829 @@
...
+using Silk.NET.Core.Loader;
+using Silk.NET.Direct3D11;
+using Extension = Silk.NET.Core.Attributes.ExtensionAttribute;
+
+#pragma warning disable 1591
+
+namespace Silk.NET.Direct3D11.Extensions.D3D11on12
+{
+    [Extension("d3d11on12.h")]
+    public unsafe partial class Direct3D11on12 : NativeExtension<D3D11>
```
Perksey, line 20: This should probably be `D3D11On12` to keep consistent with the main class being `D3D11` as well as to keep consistent with pascal case

src/Microsoft/Extensions/Silk.NET.Direct3D11.Extensions.D3D11on12/Direct3D11on12.gen.cs
```diff
@@ -0,0 +1,829 @@
...
+using Silk.NET.Core.Native;
+using Silk.NET.Core.Attributes;
+using Silk.NET.Core.Contexts;
+using Silk.NET.Core.Loader;
+using Silk.NET.Direct3D11;
+using Extension = Silk.NET.Core.Attributes.ExtensionAttribute;
+
+#pragma warning disable 1591
+
+namespace Silk.NET.Direct3D11.Extensions.D3D11on12
```
Perksey, line 17: Namespace should be D3D11On12 (uppercase O)

Perksey (changes requested):

generator.json
Beyley: a my bad

src/Microsoft/Extensions/Silk.NET.Direct3D11.Extensions.D3D11on12/Direct3D11on12.gen.cs
Beyley: u rite, just checked the repo https://github.com/microsoft/D3D11On12
shall fix

generator.json
Beyley: to be more accurate i actually have a separate json file i paste a single job into so i dont wait multiple mins per iteration on the generation, i just forgot to copy back to generator.json

Beyley:
> A project has not been included in the solution and will not be shipped! "D:\a\Silk.NET\Silk.NET\src\Microsoft\Extensions\Silk.NET.Direct3D11.Extensions.D3D11On12\Silk.NET.Direct3D11.Extensions.D3D11On12.csproj" if this is acceptable please add the project name (excluding the path and extension) to the AllowedExclusions array in the NUKE Build.CI.AutoReview.cs file.

aw damnit i forget to do that every time

Perksey (approved):

PR #1197, opened 2023-01-02
Fix to ImGuiVulkan Experiment to run on MacOSX

viniciusmiguel:
# Summary of the PR
The current implementation only works on Windows, based on my tests it was never validated over MoltenVK,
I successfully managed to run on the Mac after troubleshooting and fixing the Vulkan initialisation config.
My proposal is to bring the fixes to the Experiment and hopefully some enhancements to make this an official example.

### DIFF
--- a/src/Lab/Experiments/ImGuiVulkan/ImGuiVulkanApplication.cs
+++ b/src/Lab/Experiments/ImGuiVulkan/ImGuiVulkanApplication.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
@@ -75,8 +76,8 @@ namespace ImGuiVulkan
 		private KhrSwapchain _vkSwapchain;
 		private ExtDebugUtils _debugUtils;
 		private string[] _validationLayers = { "VK_LAYER_KHRONOS_validation" };
-		private string[] _instanceExtensions = { ExtDebugUtils.ExtensionName };
-		private string[] _deviceExtensions = { KhrSwapchain.ExtensionName };
+        private List<string> _instanceExtensions = new () { ExtDebugUtils.ExtensionName };
+		private List<string> _deviceExtensions = new () { KhrSwapchain.ExtensionName };
 
 		private void InitWindow()
 		{
@@ -320,12 +321,20 @@ namespace ImGuiVulkan
 		{
 			_vk = Vk.GetApi();
 
+            var isMacOs = Environment.OSVersion.Platform is PlatformID.Unix or PlatformID.MacOSX; 
+            
+            if (isMacOs)
+            {
+                _instanceExtensions.Add("VK_KHR_portability_enumeration");
+                _deviceExtensions.Add("VK_KHR_portability_subset");
+            }
+            
 			if (EnableValidationLayers && !CheckValidationLayerSupport())
 			{
 				throw new NotSupportedException("Validation layers requested, but not available!");
 			}
 
-			var appInfo = new ApplicationInfo
+            var appInfo = new ApplicationInfo
 			{
 				SType = StructureType.ApplicationInfo,
 				PApplicationName = (byte*)Marshal.StringToHGlobalAnsi("Hello Triangle"),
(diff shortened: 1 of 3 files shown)


viniciusmiguel:
The project works out of the box over Rosetta with the changes proposed.
On my test project, I also compiled native arm64 but for that to work, it was necessary to download ImGui.NET and build it locally.

src/Vulkan/Silk.NET.Vulkan/VulkanLibraryNameContainer.cs
```diff
@@ -14,7 +14,7 @@ internal class VulkanLibraryNameContainer : SearchPathContainer
         public override string Linux => "libvulkan.so.1";
 
         /// <inheritdoc />
-        public override string MacOS => "libMoltenVK.dylib";
```
Beyley, line 17: this is a possible breaking change for downstream projects, is there a reason for this?

viniciusmiguel: This lib definition is incorrect.

The MoltenVK library (dylib) is considered as an ICD (Installable Client Driver) by the Vulkan loader. The Vulkan loader dynamically loads ICDs at execution time. When using the Vulkan loader, ICDs are not linked directly to the application. Similarly, the Vulkan loader dynamically loads the layers at execution time. Layers are not linked directly to the application.
The MoltenVK library is a little unusual because it is built in such a way that an application can link to it directly and run MoltenVK (Vulkan subset) applications without using the loader and layers. But if you want to use the loader and layers, you don't link MoltenVK and the layers to your application. You link only the loader. The loader then dynamically loads the MoltenVK lib as an ICD and dynamically loads the layers if requested.

viniciusmiguel: in resume it is impossible to create a vkInstance on MacOs binding to MoltenVK directly. Did anyone test this before?

Perksey: I'm happy with this change, iirc Ryujinx had to do this anyway to get it working on macOS.

src/Lab/Experiments/ImGuiVulkan/ImGuiVulkanApplication.cs
```diff
@@ -320,12 +321,20 @@ private unsafe void CreateInstance()
 		{
 			_vk = Vk.GetApi();
 
+            var isMacOs = Environment.OSVersion.Platform is PlatformID.Unix or PlatformID.MacOSX;
```
Perksey, line 324: This seems wrong.

```suggestion
            var isMacOs = Environment.OSVersion.Platform is PlatformID.MacOSX; 
```

surely?

viniciusmiguel: Indeed, the problem is that MacOSX is showing like Unix in my system, I don't know why, it would need some testing on Linux to see how it behaves or may be a bug in the runtime.

Perksey: Can we use OperatingSystem.IsMacOS() instead given you've retargeted to .NET 6? These are the latest APIs that should work.

viniciusmiguel: I would suggest solving this right now to remove the Unix part to avoid issues on the Linux side and place a comment over the assignment warning the developers of the issue.
I could follow up with the runtime team to find out what is going on.

Perksey: I think the old Environment/RuntimeEnvironment variables are deprecated (and not really guaranteed to work) anyway, so the runtime team would probably just point you at `OperatingSystem`.

viniciusmiguel: Fixed now with the solution proposed here:
https://github.com/dotnet/runtime/issues/21660#issuecomment-633731507

Perksey (approved):
Looks fine but again RuntimeInformation is deprecated, please start using OperatingSystem APIs in the future (they're also a lot more obvious to use!).

src/Vulkan/Silk.NET.Vulkan/VulkanLibraryNameContainer.cs
marysaka: This is wrong and break Ryujinx see https://github.com/Ryujinx/Ryujinx/issues/4689.

Beyley: > This is wrong and break Ryujinx see [Ryujinx/Ryujinx#4689](https://github.com/Ryujinx/Ryujinx/issues/4689).

Our loader can now try multiple names, feel free to PR in something to try both libnames

viniciusmiguel: > This is wrong and break Ryujinx see [Ryujinx/Ryujinx#4689](https://github.com/Ryujinx/Ryujinx/issues/4689).

I will re-check the documentation and re-test.

marysaka: Will do if I find the time for that.

I should emphasis that this MoltenVK is built as `libMoltenVK.dylib` and never `libvulkan.dylib".

I know that a vulkan loader here could have done the job, but this is something **that should have been reported in the changelog as it's a breaking change.**

Beyley: > Will do if I find the time for that.
> 
> I should emphasis that this MoltenVK is built as `libMoltenVK.dylib` and never `libvulkan.dylib`.
> 
> I know that a vulkan loader here could have done the job, but this is something **that should have been reported in the changelog as it's a breaking change.**

Not listing it on the changelog is completely our fault (there were a *lot* of changes in 2.17, we surely were to have missed some), we will try to not miss breaking changes like this in the future!

Beyley: > Will do if I find the time for that.
> 
> I should emphasis that this MoltenVK is built as `libMoltenVK.dylib` and never `libvulkan.dylib`.
> 
> I know that a vulkan loader here could have done the job, but this is something **that should have been reported in the changelog as it's a breaking change.**

Found the time today to knock this down, PR opened https://github.com/dotnet/Silk.NET/pull/1419

PR #1198, opened 2023-01-04
Fix wide string marshalling

Perksey:


Beyley (approved):
bro WHAT A⹳坒祂整摁牤獥䉳晵敦獲㜮

HurricanKai:
Random that I saw this, but this doesn't seem like it'd work?
Won't this just write a 0 in the middle of the array?
Iirc it should null terminate (which is why the +1 existed) so it should be allocating input.Length * 2 + 1 and write to input.Length * 2?

Beyley:
> Random that I saw this, but this doesn't seem like it'd work? Won't this just write a 0 in the middle of the array? Iirc it should null terminate (which is why the +1 existed) so it should be allocating input.Length * 2 + 1 and write to input.Length * 2?

theres a second line of code writing the null terminator
image

Beyley:
oh shit

Perksey:
@Beyley Can you please be careful with language on GitHub? Appreciate the team is quite colloquial on Discord, but GitHub tends to be more formal :)

PR #1222, opened 2023-01-14
Use a self-hosted runner

Perksey:
This allows for more rapid iteration times and assurance against timeouts in release pipelines (which has been a big problem!)
- use windows-quick (self-hosted only) for release tags on this repo
- use windows-2022 (self-hosted or GitHub-hosted depending on what's available) for development in this repo
- use windows-latest (GitHub-hosted only) in all other cases (i.e. community contributions)

Note: the reason we use windows-2022 instead of windows-latest is so that both the self-hosted runner
(which has the windows-2022 label but NOT the windows-latest label) and the GitHub hosted runner are
treated as candidates. We will never have a windows-latest self-hosted runner even if it is the latest. 

There is currently only one self-hosted runner (`useaz1-githubrunner1v` aka US East Azure 1 GitHub Runner 1 Virtual Machine)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,7 @@ on:
       - 'documentation/**'
 jobs:
   Build:
-    runs-on: self-hosted
+    runs-on: ${{ github.repository == 'dotnet/Silk.NET' && 'windows-quick' || 'windows-latest' }}
     steps:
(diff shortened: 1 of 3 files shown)


Perksey:
This is now ready for review

Beyley (approved):
I dont see anything wrong

PR #1292, opened 2023-02-12
Implement AL_SOFT_callback_buffer extension

Beyley:
# Summary of the PR
Implements the openal-soft extension AL_SOFT_callback_buffer

# Further Comments
It seems checking whether OpenAL extensions are available is broken, but that looks like an unrelated problem to solve at a later time

### DIFF
--- a/src/OpenAL/Extensions/Silk.NET.OpenAL.Extensions.Soft/SoftCallbackBuffer.cs
+++ b/src/OpenAL/Extensions/Silk.NET.OpenAL.Extensions.Soft/SoftCallbackBuffer.cs
@@ -1,0 +1,40 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Silk.NET.Core.Attributes;
+using Silk.NET.Core.Contexts;
+using Silk.NET.Core.Native;
+using Silk.NET.OpenAL;
+
+namespace Silk.NET.OpenAL.Extensions.Soft
+{
+    /// <summary>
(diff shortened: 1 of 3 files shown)


src/OpenAL/Extensions/Silk.NET.OpenAL.Extensions.Soft/SoftCallbackBuffer.cs
```diff
@@ -0,0 +1,40 @@
...
+        [NativeApi(EntryPoint = "GetBufferPtrSOFT")]
+        public partial void GetBufferPtr(uint buffer, SoftBufferCallback param, void** ptr);
+        
+        /// <inheritdoc />
+        [NativeApi(EntryPoint = "GetBuffer3PtrSOFT")]
+        public partial void GetBuffer3Ptr(uint buffer, SoftBufferCallback param, void** ptr0, void** ptr1, void** ptr2);
+        
+        /// <inheritdoc />
+        [NativeApi(EntryPoint = "GetBufferPtrvSOFT")]
+        public partial void GetBufferPtrv(uint buffer, SoftBufferCallback param, void** ptr);
```
Perksey, line 38: We tend to omit the v suffixes

Perksey (changes requested):

Perksey (approved):

src/OpenAL/Extensions/Silk.NET.OpenAL.Extensions.Soft/SoftCallbackBuffer.cs
Perksey: Nevermind, reverting this, because the signature of the two GetBufferPtrs are exactly the same!

PR #1318, opened 2023-03-03
Update WebGPU, Dawn, and WGPU

Beyley:
# Summary of the PR
ABI keeps changing :^)

https://discord.com/channels/521092042781229087/607634593201520651/1081343813794140170


Perksey:
@Beyley did you want this automated? Right now the WIP auto regenerate bindings workflow doesn't update submodules, but if it did then you wouldn't need to do this. Only question would be around native binary versioning.

Beyley:
> @Beyley did you want this automated? Right now the WIP auto regenerate bindings workflow doesn't update submodules, but if it did then you wouldn't need to do this. Only question would be around native binary versioning.

having this be automated would be very helpful, as the native specs change rapidly, and native libs shouldnt be a problem, since when submodules are updated, when we actually do have native lib building, ittl just rebuild the libs to that version too

Perksey (approved):

PR #1251, opened 2023-01-22
Silly fixes

Perksey:
- [x] #1249 
- [x] #1155
- [x] #1095 
- [x] #1094 
- [x] #1093 
- [x] #1091
- [x] #1070 
- [x] #1054 
- [x] #975 
- [x] #960 
- [x] #950 
- [x] #821 
- [x] Fix Vulkan BuildTools
- [x] Create BuildTools CI job in NUKE

PR #1324, opened 2023-03-10
Regenerate bindings as of 10/03/2023

silkdotnet:


HurricanKai (changes requested):
Looks like we dropped a lot of [Flags] is that intended?

Perksey:
Not intended, I'd expect a few `None`s to be dropped but not `[Flags]`. Already have a fix on #1251, will rerun.

Perksey (changes requested):

Issue #1356, opened 2023-03-26
GLFW3.dll missing in Net461 projects


paralaxsd:
# Summary
Hi,
I stumbled upon an issue where the native `glfw3.dll` would not be copied to a console app's output directory if targeting the .Net 4.6.1 framework. I stripped the issue down to the basics and [published a repository that demonstrates the issue](https://github.com/paralaxsd/ultz.native.glw-framework-test).

I also looked at the [Silk.NET.GLFW.Native](https://github.com/dotnet/Silk.NET/blob/main/src/Native/Silk.NET.GLFW.Native/Silk.NET.GLFW.Native.csproj) project and compared the approach to other .net libraries with native dependencies [[1](https://github.com/mellinoe/ImGui.NET/blob/master/src/ImGui.NET/build/net40/ImGui.NET.targets)], [[2](https://github.com/mono/SkiaSharp/blob/main/binding/SkiaSharp/nuget/build/net462/SkiaSharp.targets)]. These projects are using a `.targets` rather than a `.props` file for copying their native dependencies on .Net 4.

I have locally tested with an adapted nuget package that's using a .targets file instead and it fixes this issue. [I am providing a hotfix branch](https://github.com/paralaxsd/Silk.NET/commit/d81fda7166e00fefb37ac048b870d48090ebd605) that could be converted into a pull request.

# Steps to reproduce
* Clone the repository from https://github.com/paralaxsd/ultz.native.glw-framework-test
* Open the Visual Studio solution and compile both projects
* Observe (as mentioned in the readme) that the `net461` project is missing `glfw3.dll` while the file should be copied into its output directory as per [Ultz.Native.GLFW.props](https://github.com/dotnet/Silk.NET/blob/main/src/Native/Silk.NET.GLFW.Native/build/net461/Ultz.Native.GLFW.props). The `.net7` project is using a different mechanism for pulling in the libraries and will not be affected.

It stands to reason that the same issue should also befall `Ultz.Native.SDL` but I have not tested this.

Perksey:
Sounds like a good fix we’d be happy to take as a PR, but just update to a modern .NET version haha

paralaxsd:
> Sounds like a good fix we’d be happy to take as a PR, but just update to a modern .NET version haha

It's a strategic thing really. My original project supports .net 4 - 7 but .net7 is def more fun :)

But in any case, thanks - let me prepare the PR then...

Perksey:
Thanks for the fix! :)

[...]

PR #1357, opened 2023-03-26
GLFW3.dll missing in Net461 projects

paralaxsd:
This fixes an issue for a consuming .net4 applications as demonstrated at https://github.com/paralaxsd/ultz.native.glw-framework-test and discussed in https://github.com/dotnet/Silk.NET/issues/1356

# Summary of the PR
When using .net 4.6.1, referencing the `Ultz.Native.GLFW` package will not copy the `glfw3.dll` library to the output directory.


paralaxsd:
Two things:
* currently, the nuget package version is set to `3.3.10-alpha`. Should I bump this up to `3.3.10`?
* It's probably sensible to also adapt the `Silk.NET.SDL.Native` project since it works exactly the same. What do you think?

HurricanKai:
I believe adapting SDL also makes sense, but we only really use SDL on mobile, which I can't imagine works on net461 anyways. I'd be happy to merge this regardless, but if you make the same changes for SDL that'd be good.

Versioning I'd defer to @Beyley or @Perksey, not sure if there's any real versioning strategy for the native packages :stuck_out_tongue:

Perksey:
We use this pattern in all the native libs, makes sense to change it everywhere at once.

We tend to use the 4th version number as the silk-specific version number. So just whack a .1 at the end of the versions if there isn't already a 4th number

paralaxsd:
Ok since this was originally `3.3.9`, I'd go with `3.3.9.1` then.

paralaxsd:
Made changes to all native projects. There were however three projects that took their package version from `..\..\..\build\props\common.props`. I didn't change those for the time being as I'm not sure if I should: `Silk.NET.OpenGLES.ANGLE.Native`, `Silk.NET.Vulkan.Loader.Native`, `Silk.NET.Vulkan.SwiftShader.Native`

Perksey:
No worries :) those version numbers are tied to the git submodule SHAs so they'll get swept up next time we update those submodules and regenerate their bindings.

Perksey (approved):

Issue #1307, opened 2023-02-23
Android OpenGL Sample did not start


rodrigovaras:
# Summary
Tried to build and run the Android demo sample and I get:
I did some research and I notice somebody open a similar bug that was closed.
It seems the situation has not changed since I get this (problems with bluetooth permissions):

Java.Lang.RuntimeException: 'Unable to start activity ComponentInfo{com.companyname.AndroidDemo/crc640cc520f458eee55e.MainActivity}: java.lang.SecurityException: Need android.permission.BLUETOOTH_CONNECT permission for android.content.AttributionSource@a7e74ba0: AdapterService getBondedDevices'

Beyley:
this is an SDL thing to my knowledge, you need to add the BLUETOOTH_CONNECT permission to the manifest

rodrigovaras:
I did several attempts none of them worked.
I added this to Android manifest without any luck:
<uses-permission android:name="android.permission.BLUETOOTH_CONNECT" />

HurricanKai:
You didn't attach anything.
But as another hint, on older versions the permission might just be called BLUETOOH, maybe that applies? See [the official docs](https://developer.android.com/reference/android/bluetooth/BluetoothAdapter)

rodrigovaras:
here is the log:
I'm just running the sample from the app which definitely does not work. Adding all the bluetooth permission still does not make it work:

Loaded assembly: /data/data/com.companyname.AndroidDemo/files/.__override__/System.Runtime.InteropServices.RuntimeInformation.dll [External]
[SDL] Device: sargo
[SDL] Model: Pixel 3a
[SDL] onCreate()
[monodroid-assembly] open_from_bundles: failed to load assembly System.Collections.Concurrent.dll
Loaded assembly: /data/data/com.companyname.AndroidDemo/files/.__override__/System.Collections.Concurrent.dll [External]
[SDL] nativeSetupJNI()
[SDL] AUDIO nativeSetupJNI()
[SDL] CONTROLLER nativeSetupJNI()
[hidapi] Initializing Bluetooth
[AndroidRuntime] Shutting down VM
[AndroidRuntime] FATAL EXCEPTION: SDLActivity
[AndroidRuntime] Process: com.companyname.AndroidDemo, PID: 14031
[...]

Issue #1264, opened 2023-01-31
System.ArgumentException when using CreateProgram with string[] as source parameter


CamelCaseName:
# Summary
calling 
```cs
cl.CreateProgramWithSource(nint context, uint count, string[] stringsSa, nuint* lengths, out int errcode_ret);
``` 
with a valid context and valid kernel as a string in Kernels.LayoutKernel throws a "System.ArgumentException" from within the string array to pointer conversion code. The Kernel is about 4500 characters long, can be found [here](https://github.com/CamelCaseName/HousePartyTranslator/blob/2.0_web_rework/TranslatorAdmin/Explorer/OpenCL/Layout.cl). providing a valid length didn't help, and is not needed per se if the string is null terminated according to the OpenCL docs.

# Stacktrace
```
System.ArgumentException
  HResult=0x80070057
  Message=The output byte buffer is too small to contain the encoded data, encoding codepage '65001' and fallback 'System.Text.EncoderReplacementFallback'. Arg_ParamName_Name
  Source=System.Private.CoreLib
  StackTrace:
   at System.Text.Encoding.ThrowBytesOverflow() in /_/src/libraries/System.Private.CoreLib/src/System/Text/Encoding.cs:line 1090
   at System.Text.Encoding.ThrowBytesOverflow(EncoderNLS encoder, Boolean nothingEncoded) in /_/src/libraries/System.Private.CoreLib/src/System/Text/Encoding.cs:line 1101
   at System.Text.Encoding.GetBytesWithFallback(ReadOnlySpan`1 chars, Int32 originalCharsLength, Span`1 bytes, Int32 originalBytesLength, EncoderNLS encoder) in /_/src/libraries/System.Private.CoreLib/src/System/Text/Encoding.Internal.cs:line 687
   at System.Text.Encoding.GetBytesWithFallback(Char* pOriginalChars, Int32 originalCharCount, Byte* pOriginalBytes, Int32 originalByteCount, Int32 charsConsumedSoFar, Int32 bytesWrittenSoFar) in /_/src/libraries/System.Private.CoreLib/src/System/Text/Encoding.Internal.cs:line 497
   at System.Text.UTF8Encoding.GetBytes(Char* chars, Int32 charCount, Byte* bytes, Int32 byteCount) in /_/src/libraries/System.Private.CoreLib/src/System/Text/UTF8Encoding.cs:line 358
[...]

Issue #922, opened 2022-05-18
Android Crashes when screen rotates


Hanprogramer:
# Summary
Basically the app just crashes when the screen rotate, even without any rendering code. The problem is on SDL, when your app rotates it called SDL_Quit or something (not sure) but basically the screen surface is destroyed

# Steps to reproduce
- Platform: Android 
- Framework Version: .NET 6 android
- API: SDL
1. I created a new android project with `dotnet new`
2. Then I copied some code from the example repo
3. It works, but when I rotate it, it crashes

# Comments
Errors that were found:
```
[DecorView[]] getWindowModeFromSystem  windowmode is 1
[libEGL] validate_display:87 error 3008 (EGL_BAD_DISPLAY)
[SDL] surfaceDestroyed()
[SDL] nativePause()
[SDL] surfaceCreated()
[SDL] surfaceChanged()
[SDL] pixel format RGBA_8888
[SDL] Window size: 2177x855
[SDL] Device size: 1080x2400
[SDL] nativeResume()
[Looper] PerfMonitor longMsg : seq=58 plan=15:32:27.306 late=8ms wall=1928ms running=28ms runnable=80ms binder=90ms h=org.libsdl.app.SDLActivity$SDLCommandHandler w=0 procState=18
[SDL] surfaceDestroyed()
[SDL] nativePause()
[SDL] onStop()
[SDL] onDestroy()
```
```
[Looper] PerfMonitor doFrame : time=64ms vsyncFrame=1218431 latency=675ms procState=18 historyMsgCount=7 (msgIndex=2 wall=645ms seq=233 running=22ms runnable=1ms binder=21ms late=102ms h=org.libsdl.app.SDLActivity$SDLCommandHandler w=0)
[OpenGLRenderer] Davey! duration=749ms; Flags=1, IntendedVsync=25496197045159, Vsync=25496872045132, OldestInputEvent=9223372036854775807, NewestInputEvent=0, HandleInputStart=25496873638449, AnimationStart=25496873642980, PerformTraversalsStart=25496875238293, DrawStart=25496928107564, SyncQueued=25496933524751, SyncStart=25496934146262, IssueDrawCommandsStart=25496935063345, SwapBuffers=25496943527772, FrameCompleted=25496947080637, DequeueBufferDuration=1200885, QueueBufferDuration=2394167, GpuCompleted=9223372036854775807, 
[SDL] surfaceDestroyed()
[SDL] nativePause()
[...]

PR #1372, opened 2023-03-29
New code signing & more silly fixes

Perksey:
- [x] #922 
- [x] #1264 
- [x] #1307 
- [x] #1350 
- [x] Implicit count not inverting calculation in count
    - [x] Add expression recognition
    - [x] Invert expression
- [x] Integrate new code signing tool
- [x] Use new code signing tool in CI/CD
- [x] Update patch notes


Perksey:
Ready for review.

src/Core/Silk.NET.Core/Native/SilkMarshal.cs
```diff
@@ -142,9 +142,8 @@ public static int GetMaxSizeOf(string? input, NativeStringEncoding encoding = Na
             => encoding switch
             {
                 NativeStringEncoding.BStr => -1,
-                NativeStringEncoding.LPStr => ((input?.Length ?? 0) + 1) * Marshal.SystemMaxDBCSCharSize + 1,
-                NativeStringEncoding.LPTStr => (input is null ? 0 : Encoding.UTF8.GetMaxByteCount(input.Length)) + 1,
-                NativeStringEncoding.LPUTF8Str => (input is null ? 0 : Encoding.UTF8.GetMaxByteCount(input.Length)) + 1,
+                NativeStringEncoding.LPStr or NativeStringEncoding.LPTStr or NativeStringEncoding.LPUTF8Str
+                    => (input is null ? 0 : Encoding.UTF8.GetMaxByteCount(input.Length)) + 1,
```
Beyley, line 146: isnt LPTStr different if `UNICODE` is defined? what does silk prefer for windows functions?
(eg. LPTStr is a pointer to a 16bit number when `UNICODE` is defined but a pointer to an 8bit number if not)

Beyley (approved):
seems to lgtm aside from one small question

src/Core/Silk.NET.Core/Native/SilkMarshal.cs
Perksey: You're right however we've the LPStr/LPTStr/LPUTF8Str enums are just for compatibility and aren't really intended to be used (we're using them here out of habit :/). The Ansi/Auto/Uni/UTF8 ones are designed to be used.

Silk.NET has never observed this LPTStr behaviour and may be confusing to users if we changed this. Admittedly this was an oversight, but a 3 year old one that noone seems to have kicked up a fuss about

PR #1403, opened 2023-04-13
Add SPIRV-Reflect bindings

### LINKED ISSUE #979: Spirv-Reflect Bindings
# Summary of feature
Add bindings for [SPIRV-Reflect](https://github.com/KhronosGroup/SPIRV-Reflect)

# Comments
Related to #482 but different as the latter does not require bundling a native library - as far as I can tell, spirv-reflect does not have an extant dynamic library, as it is usually included in raw source form in native code projects

# Existing projects along the same lines
https://github.com/hyblocker/spirv-reflect-sharp is the only example I could find of a similar extant project

Beyley:
# Summary of the PR
Adds SPIRV-Reflect bindings

Closes #979 

# Further Comments
Uses Zig to cross compile a shared library, same as the linux HLSL compiler


Khitiara:
Looks like the functions are keeping the Reflect prefix, not sure if thats intended or not

Beyley:
> Looks like the functions are keeping the Reflect prefix, not sure if thats intended or not

good point, ill change the prefix and see if that helps

Beyley:
> > Looks like the functions are keeping the Reflect prefix, not sure if thats intended or not
> 
> good point, ill change the prefix and see if that helps

That breaks enums and structs, which now eroneously have a `spv` prefix, so theres no really an easy way to fix this unless you fix all the function names or fix all the enums and structs

Khitiara:
> 

not a big issue, just figured id mention it

Beyley:
Moving back to draft, i need to separate `spirv.h` into its own project, [SPIRV-Cross](https://github.com/KhronosGroup/SPIRV-Cross) also depends on `spirv.h`, so tomorrow ill split it off into a Silk.NET.SPIRV project, so that SPIRV-Cross and SPIRV-Reflect can both reference it

src/SPIRV/Silk.NET.SPIRV/Enums/FunctionParameterAttribute.gen.cs
```diff
@@ -0,0 +1,36 @@
...
+
+namespace Silk.NET.SPIRV
+{
+    [NativeName("Name", "SpvFunctionParameterAttribute_")]
+    public enum FunctionParameterAttribute : int
+    {
+        [NativeName("Name", "SpvFunctionParameterAttributeZext")]
+        Zext = 0x0,
+        [NativeName("Name", "SpvFunctionParameterAttributeSext")]
+        Sext = 0x1,
```
Perksey, line 18: lol

src/SPIRV/Silk.NET.SPIRV/Enums/ImageChannelOrder.gen.cs
```diff
@@ -0,0 +1,58 @@
...
+        Rx = 0xA,
+        [NativeName("Name", "SpvImageChannelOrderRGx")]
+        RGx = 0xB,
+        [NativeName("Name", "SpvImageChannelOrderRGBx")]
+        RGBx = 0xC,
+        [NativeName("Name", "SpvImageChannelOrderDepth")]
+        Depth = 0xD,
+        [NativeName("Name", "SpvImageChannelOrderDepthStencil")]
+        DepthStencil = 0xE,
+        [NativeName("Name", "SpvImageChannelOrdersRGB")]
```
Perksey, line 45: This is incorrectly trimmed to be Rgb (instead of Srgb) so 2 things:
1. Can we figure out why the current trimming logic isn't stopping at Order vs Orders (as our current code should just think they are different prefixes)
2. Can we bring back prefix overrides to account for this

src/SPIRV/Silk.NET.SPIRVReflect/Enums/AnonymousEnumSpirvReflectL261C1.gen.cs
```diff
@@ -0,0 +1,23 @@
...
+using Silk.NET.Core.Attributes;
+
+#pragma warning disable 1591
+
+namespace Silk.NET.SPIRVReflect
+{
+    [Flags]
+    [NativeName("Name", "__AnonymousEnum_spirv_reflect_L261_C1")]
+    public enum AnonymousEnumSpirvReflectL261C1 : int
+    {
```
Perksey, line 15: Rename anonymous if possible?

Perksey (changes requested):
I have reviewed every single file which is a first for a bindings PR

src/SPIRV/Silk.NET.SPIRV/Enums/ImageChannelOrder.gen.cs
Beyley: I'm not sure why its not stopping, and yeah prefix overrides would work here, although we'd need another special case for "if its in the prefix overrides, then we use the native name as the trimming name as before instead of lenient underscore"

generator.json
```diff
@@ -2393,6 +2393,172 @@
...
+                    "-Wno-nonportable-include-path",
+                    "-Wno-pragma-pack",
+                    "-I$windowsSdkIncludes",
+                    "-Ibuild/submodules/SPIRV-Reflect"
+                ],
+                "traverse": [
+                    "build/submodules/SPIRV-Headers/include/spirv/unified1/spirv.h"
+                ],
+                "classes": {
+                    "spirv.h": "[Core]SPIRV"
```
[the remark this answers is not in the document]
Beyley: I personally like Spv more than SpirV aesthetically, so i think i'll go with that

src/SPIRV/Silk.NET.SPIRVReflect/Enums/AnonymousEnumSpirvReflectL261C1.gen.cs
Beyley: Yeah the current name is *bad*, but im not sure what would be a better name for this specific enum

src/SPIRV/Silk.NET.SPIRV/Enums/ImageChannelOrder.gen.cs
Perksey: Yeah that's fine, if it's a prefix override we take that as gospel including the casing, rather than going the underscore paths, converting back to pascal case after through the naming translator

src/SPIRV/Silk.NET.SPIRVReflect/Enums/AnonymousEnumSpirvReflectL261C1.gen.cs
Perksey: I wonder if we rename both of these anonymous enums to "Constants" the bakery will kick in?

generator.json
```diff
@@ -2393,6 +2393,172 @@
...
+                    "spirv_reflect"
+                ]
+            },
+            "output": {
+                "mode": "Default",
+                "path": "src/SPIRV/Silk.NET.SPIRVReflect",
+                "licenseFile": "build/LICENSE_HEADER.txt",
+                "props": "build/props/bindings.props"
+            },
+            "namespace": "Silk.NET.SPIRVReflect",
```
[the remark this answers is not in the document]
Beyley: that would make sense here, i'll change it to that

Perksey (approved):
Approved pending tests

src/SPIRV/Silk.NET.SPIRV.Reflect.Tests/ShaderModuleTests.cs
```diff
@@ -0,0 +1,47 @@
+using Silk.NET.Core.Native;
```
Perksey, line 1: copyright header you silly goose

src/OpenGL/Silk.NET.OpenGL.Legacy/Enums/FogParameter.gen.cs
```diff
@@ -12,22 +12,22 @@ namespace Silk.NET.OpenGL.Legacy
...
-        [Obsolete("Deprecated in favour of \"End\"")]
+        [System.Obsolete]
         [NativeName("Name", "GL_FOG_END")]
         FogEnd = 0xB64,
-        [Obsolete("Deprecated in favour of \"Mode\"")]
+        [System.Obsolete]
         [NativeName("Name", "GL_FOG_MODE")]
         FogMode = 0xB65,
-        [Obsolete("Deprecated in favour of \"Color\"")]
+        [System.Obsolete]
```
Beyley: It seems your code doesnt put a helpful obsolete message here, returning the obsolescence to what it was before, i feel like my patch's behaviour of always overwriting obsolete messages is something we should try to carry over

Perksey: This was intentional. If a token is obsoleted at the convert/construct level, all variants of that token should be obsoleted with the same message - you should not use that value!

This particular obsolete attribute is coming from OpenGLReader

Beyley: > This was intentional. If a token is obsoleted at the convert/construct level, all variants of that token should be obsoleted with the same message - you should not use that value!
> 
> This particular obsolete attribute is coming from OpenGLReader

I mean i do understand its coming from OpenGLReader and intentional, but if we are able to detect in the bakery the new name and create a more helpful message to the user, imo we should update the message, unless theres a case where the bakery produces a wrong message and doesnt have the proper information on the enum here?

Perksey: I wouldn't object to changing OpenGLReader to add a message.

Beyley: I attempted to do this, but i dont understand ReadEnums in OpenGLReader enough and could not figure out how to get the new name into there

Perksey: Shall we leave this comment then

Beyley: > Shall we leave this comment then

If you dont feel like working on it (i tried again last night and couldnt figure it out again), then its not a blocker, and this discussion can be closed, with maybe a `TODO` or a note in the code marking something to eventually be done, if someone cares

Perksey: Done. Donezo. Donion ring.

Perksey (approved):
I PRESENT UNTO THEE MY LARGE GREEN TICK

PR #1419, opened 2023-04-19
Vulkan: Add MoltenVK and portability to MacOS library names

Beyley:
https://github.com/dotnet/Silk.NET/pull/1197#discussion_r1168295230

# Further Comments
MoltenVK is (obviously) above portability, and while gfx-rs/portability is dead development wise, there are maybe people who want to package it with their app and use it for whatever reason, or end users who want to use it with some app, so lets shove it at the end of the libnames to check, since we are able to, and it shouldnt hurt anything, given the loader and Molten VK are prioritized


Perksey:
Can we not add libportability, I’d rather just only have MoltenVK and have it be for legacy reasons. Really we should be using the loader on all platforms but we (I) backed ourselves into a corner by Not Doing That.

Perksey (approved):

PR #1420, opened 2023-04-19
Further chain improvements

Khitiara:
# Summary of the PR
Adds HeadRef and ItemNRef to the managed Chain types in Silk.NET.Vulkan to support easier modification of properties on chain elements, as well as pulling the setters up to the interfaces.


Perksey:
Can't we just change the original properties? Silk.NET only has a policy against source breaking, which this won't violate.

Khitiara:
> Can't we just change the original properties? Silk.NET only has a policy against source breaking, which this won't violate.

wasnt aware that changing to a ref property wouldnt break source, will do that change instead then

Perksey (approved):

PR #1425, opened 2023-04-21
Update SDL deps

### LINKED ISSUE #1424: Update SDL2.dll on Windows
# Summary
The version of SDL that `Silk.NET.SDL` targets is newer than the DLL that `Ultz.Native.SDL` outputs, meaning that some newer methods (specifically `SDL_GetWindowSizeInPixels`) are not present in the existing DLL.

# Steps to reproduce
- Platform: Windows
Just use any SDL function that is present in >SDL 2.14.0, in my example SDL_GetWindowSizeInPixels which was added in SDL 2.26.

# Comments
I am going to look to open a PR simply replacing the SDL DLLs with the versions available on the SDL releases page, but I thought I'd open an issue in case there is some CI runner that does it automatically etc.

skyebird189:
# Summary of the PR
Fixes #1424.

# Further Comments
I am unable to update the ARM or Mac libraries as I do not have the appropriate machines to build SDL for them.


Perksey:
I don't think we want to entertain any updates to SDL at this time that we don't need to do for Windowing, same for GLFW. SDL specifically are bindings that the Silk.NET team do not want to maintain and are as good as obsolete already.

However, if we do go ahead with this PR, I'd want to move SDL to build in CI like the other native binaries. We have no guarantee that you haven't loaded these up with malware etc.

skyebird189:
Right ok, I understand.

Perksey:
Closing this for now, will be superseded by Beyley's zig build ports.

PR #1559, opened 2023-07-07
Build zig libs against glibc 2.16

Beyley:
# Summary of the PR
We build against such an old version mostly because "we can", theres not much of a downside in the case of these libs


Perksey (approved):
The glibc in use by silk rn fr: https://youtu.be/ur57IunS9To?t=52s

PR #1563, opened 2023-07-08
Vulkan: Explicit layer extensions

### LINKED ISSUE #1298: IsInstanceExtensionPresent/TryGetInstanceExtension and layers
The IsInstanceExtensionPresent/TryGetInstanceExtension functions don't take into account a layer, and report only core extensions (or extensions from implicitly enabled layers), but there are some extensions, that may not provided by core implementation, but are provided in layer. For example, VK_EXT_debug_utils may not be present in core, but is present in VK_LAYER_KHRONOS_validation

### LINKED ISSUE #1568: Vulkan IsDeviceExtensionPresent doesn't discriminate caches by device
# Summary
When discovering the available device extension, the device handle we're obtaining the extension cache for is not stored in the hash set (i.e. it should be a `HashSet<(Device, string)>` or `Dictionary<Device, HashSet<string>>`). This could lead to an extension being incorrectly reported as present for one logical device because the cache was previously populated using another logical device.

# Comments
As discussed in #1563

Beyley:
# Summary of the PR
Before we could not properly get layer extensions, because we always passed `null` for the `pLayerName` parameter

Closes #1298 

# Further Comments
Relies on #1559 because im just a lil goofy silly lil guy and forgot to switch branches properly

Fixes #1568


src/Vulkan/Silk.NET.Vulkan/Vk.cs
```diff
@@ -154,24 +156,32 @@ protected override void PostInit()
...
                 using var mem = GlobalMemory.Allocate((int) instanceExtPropertiesCount * sizeof(ExtensionProperties));
                 var props = (ExtensionProperties*) Unsafe.AsPointer(ref mem.GetPinnableReference());
 
                 // Get properties
-                EnumerateInstanceExtensionProperties((byte*) 0, &instanceExtPropertiesCount, props);
+                EnumerateInstanceExtensionProperties((byte*) layerPtr, &instanceExtPropertiesCount, props);
+
+                if(layerPtr != 0)
+                {
+                    SilkMarshal.Free(layerPtr);
```
HurricanKai, line 183: There's a memory leak here if an exception occurs in between.

HurricanKai (approved):
I'd also prefer using GlobalMemory instead of raw pointers.
This is fine to merge regardless, I don't think anything is important here.

src/Vulkan/Silk.NET.Vulkan/Vk.cs
```diff
@@ -167,31 +168,34 @@ public unsafe bool IsInstanceExtensionPresent(string extension, string layer = n
...
+                try {
+                    // Get count of properties
+                    var instanceExtPropertiesCount = 0u;
+                    EnumerateInstanceExtensionProperties((byte*) layerPtr, &instanceExtPropertiesCount, null);
 
-                // Initialise return structure
-                using var mem = GlobalMemory.Allocate((int) instanceExtPropertiesCount * sizeof(ExtensionProperties));
-                var props = (ExtensionProperties*) Unsafe.AsPointer(ref mem.GetPinnableReference());
+                    // Initialise return structure
+                    var props = stackalloc ExtensionProperties[(int)instanceExtPropertiesCount];
```
HurricanKai, line 177: This may be slightly nitpicky, but the extension property count is easily controlled externally and even for spec-compliant implementations there's not a low enough limit.
`ExtensionProperties` is additionally of significant size (~260 bytes).
I have seen implementations expose > 100 extensions, which likely isn't enough to outright crash the process, but depending on the setup still isn't ideal.
A misbehaving, or (highly unlikely) attacking implicit layer could easily bump this to numbers far to great for a stack allocation.
The benefit of this also seems very small so please elaborate or revert this change.

src/Vulkan/Silk.NET.Vulkan/Vk.cs
```diff
@@ -258,8 +262,7 @@ public unsafe bool IsDeviceExtensionPresent(PhysicalDevice device, string extens
                     EnumerateDeviceExtensionProperties(device, (byte*) layerPtr, &deviceExtPropertiesCount, null);
 
                     // Initialise return structure
-                    mem = GlobalMemory.Allocate((int) deviceExtPropertiesCount * sizeof(ExtensionProperties));
-                    var props = (ExtensionProperties*) Unsafe.AsPointer(ref mem.GetPinnableReference());
+                    var props = stackalloc ExtensionProperties[(int)deviceExtPropertiesCount];
```
HurricanKai, line 265: Same logic applies

HurricanKai (changes requested):

src/Vulkan/Silk.NET.Vulkan/Vk.cs
Beyley: i had misread what perksey had said [here](https://discord.com/channels/521092042781229087/587346162802229298/1127352316857688095)

src/Vulkan/Silk.NET.Vulkan/Vk.cs
```diff
@@ -154,34 +157,47 @@ protected override void PostInit()
...
         {
             // Note we use optimistic code to avoid locks, if this is called on multiple threads
             // then multiple initialisations can happen, only one thread will succeed in initialising the cache though.
             var cachedInstanceExtensions = _cachedInstanceExtensions;
             if (cachedInstanceExtensions is null)
             {
-                // Get count of properties
-                var instanceExtPropertiesCount = 0u;
-                EnumerateInstanceExtensionProperties((byte*) 0, &instanceExtPropertiesCount, null);
+                var layerPtr = layer == null ? 0 : SilkMarshal.StringToPtr(layer);
```
Perksey, line 169: Formalising my comment from Discord: this can be simplified by doing `fixed (byte* layerPtr = (layer is null ? Array.Empty<byte>() : Encoding.UTF8.GetBytes(layer)))`

src/Vulkan/Silk.NET.Vulkan/Vk.cs
```diff
@@ -154,34 +157,47 @@ protected override void PostInit()
...
-                // Get properties
-                EnumerateInstanceExtensionProperties((byte*) 0, &instanceExtPropertiesCount, props);
+                    fixed(ExtensionProperties* props = propsArr) {
+                        // Get properties
+                        EnumerateInstanceExtensionProperties((byte*) layerPtr, &instanceExtPropertiesCount, props);
+
+                        cachedInstanceExtensions = new HashSet<string>();
+                        for (var p = 0; p < instanceExtPropertiesCount; p++)
+                        {
+                            cachedInstanceExtensions.Add(Marshal.PtrToStringAnsi((nint) props[p].ExtensionName));
```
Perksey, line 186: It is not completely out of the realm of possibility that the same extension may have multiple hits. For example, consider the official Khronos Extension Layer which provides polyfill implementations where the driver does not natively support an extension, if this is being used alongside a driver that actually does support it (e.g. one polyfill is being used from the layer that isn't implemented, but the layer also provides other extensions which are implemented) this extension name would be returned for both `null` and `"VK_LAYER_KHRONOS_shader_object"` for example.

src/Vulkan/Silk.NET.Vulkan/Vk.cs
```diff
@@ -232,34 +250,46 @@ public unsafe bool IsDeviceExtensionPresent(PhysicalDevice device, string extens
                 // The lack of the device handle indicates we've not been previously initialised.  We now need a write lock.
                 _cachedDeviceExtensionsLock.EnterWriteLock();
                 GlobalMemory mem = null;
+                nint layerPtr = 0;
                 try
                 {
+                    if(layer != null)
+                    {
+                        layerPtr = SilkMarshal.StringToPtr(layer);
```
Perksey, line 258: Same comment here about Encoding.UTF8

src/Vulkan/Silk.NET.Vulkan/Vk.cs
```diff
@@ -232,34 +250,46 @@ public unsafe bool IsDeviceExtensionPresent(PhysicalDevice device, string extens
...
+                        // Get properties
+                        EnumerateDeviceExtensionProperties(device, (byte*) layerPtr, &deviceExtPropertiesCount, props);
+                        for (int j = 0; j < deviceExtPropertiesCount; j++)
                         {
-                            // We found the extension (no need to do another lookup as we're scanning anyway)
-                            // As such this has taken 2 lookups + initialisation scan.
-                            result = true;
+                            // Prefix the extension name
+                            var newKey = prefixSep + Marshal.PtrToStringAnsi((nint) props[j].ExtensionName);
+                            _cachedDeviceExtensions.Add(newKey);
```
Perksey, line 276: Wow, do we really not disambiguate for each device? No need to resolve in this PR as this is an existing issue, but really this should be a `Dictionary<Device, HashSet<string>>`

Perksey (changes requested):
Excellent, this looks much better thanks. Just a few more final commetns

Perksey (approved):

Beyley:
auto merge seems borked, not sure why
image
merging by hand

PR #1577, opened 2023-07-14
New experimental feed

Perksey:
The experimental builds are now hosted on GitLab herein. This is because it has a 250GB limit instead of the "2GB" one we have with Azure Artifacts (though we were using ~7.5GB somehow?)

The feed URL is https://dotnet.github.io/Silk.NET/nuget/experimental/index.json - this is https://gitlab.com/api/v4/projects/47661606/packages/nuget/index.json but mirrored to our site so we can give a URL to our users that we have control over, meaning future switches between experimental feed hosts are seamless.

This also adds back the documentation page on the experimental feed that was never ported over from docs.ultz.co.uk

### DIFF
--- a/website/docs/silk.net/experimental-feed.md
+++ b/website/docs/silk.net/experimental-feed.md
@@ -37,7 +37,7 @@ In order to use the experimental feed, you must change this project file slightl
 ```xml
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net5.0</TargetFramework>
+    <TargetFramework>net7.0</TargetFramework>
     <!-- Add the experimental feed as a restore source -->
     <RestoreSources>$(RestoreSources);https://dotnet.github.io/Silk.NET/nuget/experimental/index.json</RestoreSources>
   </PropertyGroup>
--- a/website/docs/silk.net/experimental-feed.md
+++ b/website/docs/silk.net/experimental-feed.md
@@ -22,7 +22,7 @@ Right now, your project should look something like this:
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net5.0</TargetFramework>
+    <TargetFramework>net7.0</TargetFramework>
   </PropertyGroup>
 
   <ItemGroup>
--- a/website/docs/silk.net/experimental-feed.md
+++ b/website/docs/silk.net/experimental-feed.md
@@ -1,0 +1,66 @@
+---
+{
+    "TableOfContents": {
+        "Name": "Experimental Feed",
+        "Url": "experimental-feed.html"
+    }
+}
+---
+
+# Experimental Feed
+
+<?# Warning "The experimental feed is not recommended for use for anything beyond playing around with the new features. We don't officially support these builds as they may be unstable, and should not be used in production." /?>
+
+<?# Note "Changes in the experimental feed happen rapidly. We recommend joining the [Silk.NET Discord server](https://discord.gg/DTHHXRt) so that you can keep up with development." /?>
+
+## Configure your project
+
+Right now, your project should look something like this:
+
+```xml
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net5.0</TargetFramework>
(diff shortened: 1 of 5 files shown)


Perksey (commented):
Ready for review

website/docs/silk.net/experimental-feed.md
```diff
@@ -0,0 +1,66 @@
...
+    <!-- Add the experimental feed as a restore source -->
+    <RestoreSources>$(RestoreSources);https://dotnet.github.io/Silk.NET/nuget/experimental/index.json</RestoreSources>
+  </PropertyGroup>
+```
+
+## Install an experimental version
+
+Now .NET has access to the experimental feed, it's time to install an experimental package. Check for the latest version on our [GitLab Package Registry](https://gitlab.com/silkdotnet/Silk.NET/-/packages).
+
+At the time of writing, the latest version is `2.0.0-build97.0`. Now to install this package, you can use your IDE's built-in NuGet client, `dotnet package add`, or just modify the project file again like so:
```
Beyley, line 50: These version numbers should probably be updated, this seems to be copied from very old docs (pre 2.0 release), and should not point to a out of support TFM

website/docs/silk.net/experimental-feed.md
```diff
@@ -0,0 +1,66 @@
...
+
+<?# Warning "Experimental Feed builds may be deleted without warning at the Silk.NET team's discretion." /?>
+
+Now you have access to bleeding-edge experimental builds of Silk.NET. Have fun!
+
+## GitHub Packages
+
+`https://dotnet.github.io/Silk.NET/nuget/experimental/index.json` shall always track the most up-to-date anonymous-access experimental feed. However, as an alternative, GitHub Packages can be used which requires authentication.
+
+To add this to your NuGet client, use URL `https://nuget.pkg.github.com/dotnet/index.json` in the configuration described in [the GitHub documentation](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-nuget-registry). The latest packages for this feed can be viewed [here](https://github.com/orgs/dotnet/packages?repo_name=Silk.NET).
```
Beyley, line 66: github packages was also broken, was this fixed?

Beyley (changes requested):

website/docs/silk.net/experimental-feed.md
Perksey: I think it’s best to leave the version numbers as is, or rather, it makes no difference as whatever build number we put here the new experimental feed won’t have. Will fix the TFM for sake of argument.

website/docs/silk.net/experimental-feed.md
Perksey: I’ve made an attempt!

website/docs/silk.net/experimental-feed.md
```diff
@@ -0,0 +1,66 @@
...
+## Configure your project
+
+Right now, your project should look something like this:
+
+```xml
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net5.0</TargetFramework>
```
Perksey, line 25:
```suggestion
    <TargetFramework>net7.0</TargetFramework>
```

website/docs/silk.net/experimental-feed.md
Perksey: Fixed the TFM, left version number as is, which thinking about it is probably better because it'll stick out like a sore thumb that the reader does not want this, whereas if we use a current build number they're more likely to be confused as to why this isn't on the experimental feed.

website/docs/silk.net/experimental-feed.md
```diff
@@ -0,0 +1,66 @@
...
+
+</Project>
+```
+
+In order to use the experimental feed, you must change this project file slightly. Add the following line to your project:
+
+```xml
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net5.0</TargetFramework>
```
Perksey, line 40:
```suggestion
    <TargetFramework>net7.0</TargetFramework>
```

Beyley (approved):
lgtm

Perksey:
image

PR #1578, opened 2023-07-14
DO NOT MERGE THIS UNDER ANY CIRCUMSTANCES

Perksey:
Hey now, don't try it at home
Do do do do do
Hey now, don't try it
Don't you dare try it
You might die if you do this at home
Do do do do
Oh, here's the bridge, woah
Don't try it
Don't do the thing that we're about to do
Do do do do
Oh, don't try this
At all, or you'll die if
You try this thing at home
Do do do do do
OH! IT'S A KEY CHANGE!
OH! IT'S A KEY CHANGE!
OH! IT'S THE DISCLAIMER SOOOOO-OOONG!
Don't try this at home
If you do, you might die
This is our disclaimer to you
Hey, don't be that guy
Just watch us do the thing you want to
Doooooooooo
Be safe
For me
But be safe
For you
DO DO DO DO
BAA-


Perksey:
I believe I have got what I wanted out of this now

PR #1584, opened 2023-07-21
Fix missing "this" keyword in ToSystem and ToGeneric for 4d vector

Wafer-EX:
I noticed that Vector4D<T> doesn't has the ToSystem function and Vector4 doesn't has the ToGeneric function, these functions in SystemNumericsExtensions class doesn't has the `this` keyword, so... I'll try to fix it. It's my first contributing experience so I'm sorry if I will make mistakes


Wafer-EX:
@dotnet-policy-service agree

Perksey:
@HurricanKai will let you take this one as it's maths (and should be fairly trivial :))

HurricanKai (approved):
LGTM

PR #1587, opened 2023-07-23
Equality operators for Bool32

### LINKED ISSUE #1586: Bool32 Equality operators
I haven't looked too much into the source code for [Bool32](https://github.com/dotnet/Silk.NET/blob/aaac557a5ed608798e02f63eb1a4282dc13eaf0b/src/Core/Silk.NET.Core/Miscellaneous/Bool32.cs) so I can't tell if it is an auto-generated type, but could `==` and `!=` operators be added to it?

Currently, doing an equality check with Bool32 is `if(a.Value == b.Value)`, which is fine, but it could be simplified to just `if(a == b)` and so on, if the operators were present.

Such operators would make sense for a bool type? 😄

Syncaidius:
I noticed that Bool32 doesn't have any `==` or `!=` operators and thought it would be nice to add them. @Perksey hinted at PRs for this in #1586, so here it is!

This adds equality operators for comparing `Bool32` with `Bool32`, `bool` or `uint`

This is my first Silk.NET PR, so feel free to tell me if I've missed something.

Closes #1586

### DIFF
--- a/src/Core/Silk.NET.Core/Miscellaneous/Bool32.cs
+++ b/src/Core/Silk.NET.Core/Miscellaneous/Bool32.cs
@@ -1,4 +1,4 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
+// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
 namespace Silk.NET.Core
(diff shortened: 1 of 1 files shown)


Syncaidius:
@dotnet-policy-service agree

Perksey (approved):
No objections, thank you for the PR :D

Khitiara:
maybe also implement IEquatable? though that could go for a different PR probably

Perksey:
Good point, I missed that.

Syncaidius:
If this PR goes through, I'll start a new one for the IEquatable changes. I missed that also. :)

PR #1588, opened 2023-07-23
Implement IEquatable in Bool32

Syncaidius:
This is a follow-up to a suggestion by @Khitiara in PR #1587, to implement `IEquatable<Bool32>`, `IEquatable<bool>` and `IEquatable<uint>` in `Bool32`.

As always, let me know if I missed anything. :)

### DIFF
--- a/src/Core/Silk.NET.Core/Miscellaneous/Bool32.cs
+++ b/src/Core/Silk.NET.Core/Miscellaneous/Bool32.cs
@@ -16,6 +16,15 @@ namespace Silk.NET.Core
         public uint Value { get; }
 
         /// <summary>
+        /// Returns the hash code for this instance.
+        /// </summary>
+        /// <returns> A 32-bit signed integer hash code.</returns>
+        public override int GetHashCode()
+        {
+            return Value.GetHashCode();
+        }
+
+        /// <summary>
         /// Compares the current <see cref="Bool32"/> to another <see cref="object"/>. Returns true if they are equal.
         /// </summary>
         /// <param name="obj"></param>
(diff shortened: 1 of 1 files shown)


src/Core/Silk.NET.Core/Miscellaneous/Bool32.cs
```diff
@@ -1,18 +1,56 @@
...
+        /// <summary>
+        /// Compares the current <see cref="Bool32"/> to another <see cref="object"/>. Returns true if they are equal.
+        /// </summary>
+        /// <param name="obj"></param>
+        /// <returns>True if <paramref name="obj"/> is equal to the current <see cref="Bool32"/>. 
+        /// This will always be false if <paramref name="obj"/> is not a <see cref="Bool32"/>.</returns>
+        public override bool Equals(object obj)
+        {
+            if (obj is Bool32 other)
+                return Equals(other);
```
Perksey, line 27: Should this also check for the other types so this just acts as a generic trampoline? i.e.:
```cs
public override bool Equals(object obj) => obj switch
{
    Bool32 val => Equals(val),
    bool val => Equals(val),
    uint val => Equals(val),
    _ => base.Equals(obj)
}
```

Syncaidius: This seems like a good idea. Makes it consistent with the explicit Equals() overloads.

Khitiara:
Should also override GetHashCode by convention when overriding Equals - that should pretty easy though you can just forward it to Value

Perksey (approved):

PR #1599, opened 2023-08-03
[net7] Use high-resolution sleep primitives to enhance CPU efficiency

Perksey:


Perksey:
This has been heavily reworked in response to Discord discourse. This is now ready for review.

Perksey:
SharpLab evidence of JIT efficacy: https://sharplab.io/#v2:C4LghgzgtgPgAgJgIwFgBQcDMACR2DC2A3utmdgA4BOAlgG5jACmuSAdACI1gDmAdgHsIwGgGMIbAMrABFAO6NRAC2wB9Kkz4ATJlWmyFwZdgC82PkzkAKAJQBuUuWr1GLOOy69BwsRP3zFFVUAVwotV39DYzMLa3tHMmcGZmwtAWCAIwAbFnVNHSoABV0aAS1TbCRsAHpsADYABgc0ckpaZJYOdOzc0PDmYtoyiqraxubWrGwMgQEs7ABJCHwsoRo+HmJsAG13NgAlYL4RKCY2fAEoChocvV06MSYJAFkmYCUyhausqz3D45op3Ol2ut0k90eLzeHy0XwoWQA8hQRAI+BIAHICBZ8LLrdY8GwAXR4b1MAD5sAAzMBZCBMOzYAC+CVwOBmc2wADVJABPPiiLa7dj/E5nC5XG66cFUB6iJ5sV7vT7fX7Co6i4ESsEQuVQpWw75IlFotiY7G4vj4okk4Dkqk0ukM5ktJztVzTWbzSQfYJZLSSa58QV/dWAsUgyV3GWQhXQ5Xw1UHUNA8WgqU6+WKmFwxHI0oms04vEbInYG12lmtKvVqsAQiWKzWG2wADIW5Wa53WrXuXyBW2O12u3ltFKZAEjEo2ABRLJgCh0rRsAAqAmANPBolRWgg2AAPGoNKOiiVhgOXUPL2o+hFx1Ep7P54uV2uN0wt9pdweQmFXINSloToslMcAACzYP8qoINgqIAGJUGApw2MQg5VnISiStgVj1ssqwQFaqGtCQF5XtgNCUlhIYAkC2LALQaK+GwAAaAAcdTMWxABCkBnEskihBQAhUMwWg2IRNbEaRNZpOJQ6SVJXZURqtH0fh4gcexrF1NxdJsIUYDBHStgTAprTOqZrToZhVjeukfoBus8QkVe5kKXBCGnMZslMiyzqMkAA=

HurricanKai:
I'm happy with this, thanks for writing the code also.

It's a bit unclear to me right now how much we still want to support the whole "custom Run" idea? This would put the final nail in the coffin for this. (But we've added other features in the past that were in this direction)

Perksey:
The thing is in 1.X Run was indeed fully custom, and made an explicit decision in 2.0 to regress here in the name of write-once-run-everywhere. There's nothing stopping our users implementing their own `Run` at all today, but had to change for iOS (and now WASM) compatibility. Users are still free to completely ignore our `Run` function if they wish.

Perksey:
@Beyley @HurricanKai this builds now, is this good for merge?

HurricanKai:
I think given this conflicts with custom `Run` we should resolve that first - what about making a new overload that takes `ShouldSpin` as a delegate parameter?

Perksey:
> I think given this conflicts with custom Run we should resolve that first - what about making a new overload that takes ShouldSpin as a delegate parameter?

If the user wants to customise this stuff they should implement the whole run loop themselves. This is the write-once-run-anywhere path where the fact that there is a while loop at all is an implementation detail.

HurricanKai:
Fair enough - can we make the ShouldSpin virtual? Would that be a problem? That's less obvious but is at least an escape hatch?

Perksey:
That'd be an API contract change and will probably need to go through the proposal system unless you disagree. I agree this would be a nice addition, but out of scope of this PR I think.

Perksey:
Actually, the user should be able to opt-out of this behaviour. Thinking about things that need regular DoEvents calls, like rhythm games.

Perksey:
This is going into draft, and would welcome being picked up by the community.

This contains an experimental attempt at making our frame timing more CPU efficient by using high-resolution sleep primitives. However, we were unable to do so in a way that is consistent and that wasn't "jittery". Please see this Discord thread: https://discord.com/channels/521092042781229087/587346162802229298/1142876986537160865

This is now **up for grabs**

Perksey:
Closing and adding to the community epic.

PR #1600, opened 2023-08-04
Reduce string allocations in Path-related calls

Youssef1313:
image


HurricanKai (approved):

src/Core/Silk.NET.Core/Loader/DefaultPathResolver.cs
```diff
@@ -260,7 +260,8 @@ private static bool TryLocateNativeAssetFromDeps
                     foreach (var runtimeLib in defaultContext.RuntimeLibraries)
                         foreach (var nativeAsset in runtimeLib.GetRuntimeNativeAssets(defaultContext, rid))
                         {
-                            if (Path.GetFileName(nativeAsset) == name || Path.GetFileNameWithoutExtension(nativeAsset) == name)
+                            var nativeAssetSpan = nativeAsset.AsSpan();
```
Perksey, line 263: This API isn't everywhere, can you add an `#ifdef NETSTANDARD2_1`? This should make the build pass :)

Perksey (changes requested):
Thanks, good spot!

PR #1610, opened 2023-08-15
Proper frame timing in BreakneckSleep

Khitiara:
# Summary of the PR
Fixes BreakneckSleep in the existing CPU usage PR to store the last tick timestamp and compare to that instead of to when Sleep is called.
The branch still doesnt actually use BreakneckSleep in the window loop but still this is in theory an improvement

### DIFF
--- a/src/Core/Silk.NET.Core/Miscellaneous/BreakneckSleep.cs
+++ b/src/Core/Silk.NET.Core/Miscellaneous/BreakneckSleep.cs
@@ -45,24 +45,15 @@ public readonly struct BreakneckSleep : IDisposable
     public nint Handle { get; }
 
     /// <summary>
-    /// The duration of the sleep.
-    /// </summary>
-    public TimeSpan Duration { get; }
-
-    /// <summary>
     /// The accuracy of the underlying implementation. 
     /// </summary>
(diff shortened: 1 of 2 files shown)


Perksey:
This does not build, but also I’m not sure I agree with the implementation. BreakneckSleep is a readonly struct, being a thin wrapper over a sleep configuration essentially. Adding mutable state like this counteracts that. Moreover, I’d rather data structs like these expose the data therein.

I think maybe we should look at the Sleep method. Maybe add a new TimeSpan parameter called “lastOvershoot” (default it as well) and change the return type to TimeSpan to allow the user to control the tracking of frame time deficits. This type has no concept of a frame time, and it may not necessarily be the case that Sleep is called in sequence/regular intervals. This current implementation would not be appropriate for that generalisation in this case.

There is an outstanding todo on my original PR to add a MaxThrottle property to IViewProperties to integrate BreakneckSleep and allow the user to control how fast they loop.

Perksey (approved):
If you wanted to add a dedicated limiter type lmk, otherwise I’m happy to merge this.

Khitiara:
im good for now, can always do another go with a proper limiter type later but im not sure of how best to handle the intricacies there

PR #1612, opened 2023-08-15
macOS WebGPU support

AsgardXIV:
# Summary of the PR
Adds WebGPU support for macOS. It's basically a port/copy of the Veldrid implementation.

Veldrid implementation: https://github.com/veldrid/veldrid/tree/master/src/Veldrid.MetalBindings

# Further Comments
Given this is basically a straight port/copy, not sure if you'll want to take this or how the license might need adjusting etc (Veldrid is also MIT so it should be compatible if it's included correctly).

It is also possible to pull in the whole Veldrid.MetalBindings projects as a dependency and it works fine, but I assumed that wouldn't be desirable so I grabbed the minimum set needed to make this work.

I implemented this for a project and figured others might benefit from it so I'd open the PR.

This works on my M1 Max MacBook Pro which is the only machine I can test it on, I believe it should also work on Intel Macs.


AsgardXIV:
> @dotnet-policy-service agree

@dotnet-policy-service agree

src/WebGPU/Silk.NET.WebGPU/Platforms/MacOS/Bool8.cs
```diff
@@ -0,0 +1,23 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Silk.NET.WebGPU.Platforms.MacOS;
```
Beyley, line 4: This should be in the same place as Bool32

src/WebGPU/Silk.NET.WebGPU/Platforms/MacOS/CAMetalLayer.cs
```diff
@@ -0,0 +1,17 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+
+namespace Silk.NET.WebGPU.Platforms.MacOS;
+
+internal struct CAMetalLayer
+{
+    public readonly IntPtr NativePtr;
```
Beyley, line 10: Silk.NET code tends to prefer `nint` or `void*` over `IntPtr`

src/WebGPU/Silk.NET.WebGPU/Platforms/MacOS/Selector.cs
```diff
@@ -0,0 +1,31 @@
...
+    public Selector(string name)
+    {
+        int byteCount = Encoding.UTF8.GetMaxByteCount(name.Length);
+        byte* utf8BytesPtr = stackalloc byte[byteCount];
+        fixed (char* namePtr = name)
+        {
+            Encoding.UTF8.GetBytes(namePtr, name.Length, utf8BytesPtr, byteCount);
+        }
+
+        NativePtr = ObjectiveCRuntime.sel_registerName(utf8BytesPtr);
```
Beyley, line 27: We have SilkMarshal to handle this type of string marshalling

src/WebGPU/Silk.NET.WebGPU/Platforms/MacOS/Selector.cs
```diff
@@ -0,0 +1,31 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+using System.Text;
+
+namespace Silk.NET.WebGPU.Platforms.MacOS;
+
+internal unsafe struct Selector
+{
+    public readonly IntPtr NativePtr;
```
Beyley, line 11: same thing with `nint` or `void*` here

Beyley (changes requested):
mostly small stuff

src/WebGPU/Silk.NET.WebGPU/Platforms/MacOS/CAMetalLayer.cs
AsgardXIV: Switched to nint

src/WebGPU/Silk.NET.WebGPU/Platforms/MacOS/Selector.cs
AsgardXIV: Switched to nint

src/WebGPU/Silk.NET.WebGPU/Platforms/MacOS/Selector.cs
```diff
@@ -0,0 +1,26 @@
...
+    public readonly nint NativePtr;
+
+    public Selector(nint ptr)
+    {
+        NativePtr = ptr;
+    }
+
+    public Selector(string name)
+    {
+        NativePtr = ObjectiveCRuntime.sel_registerName(SilkMarshal.StringToPtr(name));
```
Beyley, line 19: this string should be freed somewhere, as this is a memory leak

Beyley (changes requested):

src/WebGPU/Silk.NET.WebGPU/Platforms/MacOS/Selector.cs
AsgardXIV: Ah, my bad. I will address that.

AsgardXIV: Fixed both instances.

Perksey (approved):
I think there's a wider discussion to be had about our approach to Objective-C (namespace construction mainly), but given you've made all of these types internal I think it's fine for now. Thank you for this, this is an exceptional contribution! I'll let Beyley re-review but LGTM.

Beyley (approved):

PR #1623, opened 2023-08-24
Support MoltenVK

zvasya:
Fixed loading of the "__Internal" library for iOS, iOS Simulator and MacCatalyst using the GetMainProgramHandle method from Net7.0.
And add package with MoltenVK libraries


zvasya:
Working examples can be found here: https://github.com/zvasya/VulkanExample

Updated packages are in the LocalPackages folder.

zvasya:
@dotnet-policy-service agree

src/Core/Silk.NET.Core/Loader/LibraryLoader.cs
```diff
@@ -353,6 +353,11 @@ private class NetNextNativeLibraryLoader : LibraryLoader
         {
             protected override nint CoreLoadNativeLibrary(string name)
             {
+#if NET7_0_OR_GREATER
+                if (name == null)
```
Perksey, line 357:
```suggestion
                if (string.IsNullOrEmpty(name) || name == "__Internal")
```

src/Native/Silk.NET.MoltenVK.Native/build/netstandard2.0/Silk.NET.MoltenVK.Native.targets
```diff
@@ -0,0 +1,17 @@
+<Project ToolsVersion="15.0">
+  <PropertyGroup>
+    <_LibMoltenVKPath Condition="'$(RuntimeIdentifier)' == 'maccatalyst-x64' OR '$(RuntimeIdentifier)' == 'maccatalyst-arm64'">$(MSBuildThisFileDirectory)..\..\runtimes\MoltenVK.xcframework\ios-arm64_x86_64-maccatalyst\libMoltenVK.a</_LibMoltenVKPath>
+    <_LibMoltenVKPath Condition="$(RuntimeIdentifier) == 'iossimulator-x64' OR '$(RuntimeIdentifier)' == 'iossimulator-arm64'">$(MSBuildThisFileDirectory)..\..\runtimes\MoltenVK.xcframework\ios-arm64_x86_64-simulator\libMoltenVK.a</_LibMoltenVKPath>
+    <_LibMoltenVKPath Condition="'$(RuntimeIdentifier)' == 'ios-arm64'">$(MSBuildThisFileDirectory)..\..\runtimes\MoltenVK.xcframework\ios-arm64\libMoltenVK.a</_LibMoltenVKPath>
```
Perksey, line 5: This is not the standard layout for .NET native binaries. I don't think there's a precedent in .NET where xcframeworks are used. `runtimes/ios-arm64/libMoltenVK.a` should be sufficient.

src/Native/Silk.NET.MoltenVK.Native/Silk.NET.MoltenVK.Native.csproj
```diff
@@ -0,0 +1,22 @@
+<Project Sdk="Microsoft.NET.Sdk">
+    <Import Project="..\..\..\build\props\common.props" />
```
Perksey, line 2: Thank you for the binaries, however we are trying to move away from having binaries from an unknown (potentially untrusted) source in the repo. Recently, we have moved to defining all of our build scripts in this directory: https://github.com/dotnet/Silk.NET/tree/main/build/nuke/Native with automation here: https://github.com/dotnet/Silk.NET/tree/main/.github/workflows

I think we would want similar for MoltenVK as well. If you're happy to add a build script and a workflow, that would be amazing! But otherwise, let's just remove this import which will prevent these binaries from being shipped until someone can add trusted build scripts and CI workflows. It's nothing personal, we just try to always have binaries built from CI for security reasons.

src/Core/Silk.NET.Core/Loader/LibraryLoader.cs
Perksey, line 357: I'm not convinced this will work, or at least not work in a way that results in approval from Apple in the App Store. This is because all Vulkan calls will still be going through the dynamic loader (albeit to load functions from itself). I think you'll have to add what is called a `PInvokeOverride` which will force our loader to use `DllImport` and thereby directly link the appropriate functions. You can see an example of this here: https://github.com/dotnet/Silk.NET/blob/main/src/Windowing/Silk.NET.SDL/Sdl.cs#L14

Perksey (changes requested):
This is an amazing contribution I wasn't expecting to see given the state of our iOS support! Thank you so much! There's a few comments mostly relating to process and deployment, and a couple of code comments. Hopefully shouldn't take too much time, again thank you so much for your work here.

src/Native/Silk.NET.MoltenVK.Native/build/netstandard2.0/Silk.NET.MoltenVK.Native.targets
zvasya: I took this xcframework from https://github.com/KhronosGroup/MoltenVK/releases/tag/v1.2.5 and preferred to leave its structure as is to make it easier to maintain.

src/Native/Silk.NET.MoltenVK.Native/Silk.NET.MoltenVK.Native.csproj
zvasya: Yes, I understand your concerns about binaries.
I think I can write build script to load MoltenVK-all.tar from official MoltenVK repo, unpack and take this .xcframework. Or even pick up only the required libraries, if you think it's better that way. 
But I cannot guarantee how long The Khronos Group will continue to maintain this structure.

src/Core/Silk.NET.Core/Loader/LibraryLoader.cs
zvasya: Yes, I have not yet been able to check whether Apple will allow this on their review.
I first tried using dylib but The Khronos Group does not provide them for the Simulator and Mac Catalyst. Using the PInvokeOverride attribute is the second thing I tried. But with it I can't compile for Mac and Android (an error at the linking stage). I think it is necessary to surround these generated classes with Defines. But in other cases, this option works.

src/Native/Silk.NET.MoltenVK.Native/Silk.NET.MoltenVK.Native.csproj
zvasya: I added the MoltenVK repository as a submodule and added the build script. But there are several difficulties:
1) MoltenVK is built in Xcode, so I only build if OperatingSystem.IsMacOS()
2) Building MoltenVK requires cmake, python3 and ninja, as well as brew to install them (https://github.com/KhronosGroup/MoltenVK#install). I'm not sure if I need to add these installation steps or not?

src/Native/Silk.NET.MoltenVK.Native/build/netstandard2.0/Silk.NET.MoltenVK.Native.targets
zvasya: I left only libraries, without xcframework

src/Core/Silk.NET.Core/Loader/LibraryLoader.cs
zvasya: Done

Perksey (commented):
This is awesome! Please keep it coming, super excited to get Silk.NET.Vulkan working on iOS out of the box.

src/Core/Silk.NET.Core/Loader/LibraryLoader.cs
Perksey: I've just checked the runtime source code and GetMainProgramHandle uses `dlopen`. libdl is banned by Apple on iOS, and apps using these symbols will not pass their review.

> Using the PInvokeOverride attribute is the second thing I tried. But with it I can't compile for Mac and Android (an error at the linking stage).

That's odd that Android is affected, perhaps we need to use `#if __IOS__` on the attribute and add a `net7.0-ios` TFM on Vulkan... Might not be the worst idea anyway, and have us automatically ProjectReference the new native package so users automatically download this.

But I know what's causing this, can you send me the compilation errors? Chances are we need to make some functions exempt from this override so we don't try and link them, as we had to do this with SDL [here](https://github.com/dotnet/Silk.NET/blob/main/generator.json#L881)

After those changes have been made BuildTools need to be rerun by `dotnet run --project src/Core/Silk.NET.BuildTools/Silk.NET.BuildTools.csproj -- generator.json --profile=Vulkan`. This will change the generated code.

If you need any help with any of this or would rather not do this, please let me know - this is quite nitty-gritty Silk.NET stuff which I'd imagine you didn't think you were signing up for! I'm also always happy to talk anything over here or in the Silk.NET Discord.

src/Native/Silk.NET.MoltenVK.Native/Silk.NET.MoltenVK.Native.csproj
zvasya: I looked into the capabilities of GitHub Actions and I think the new version should work

src/Core/Silk.NET.Core/Loader/LibraryLoader.cs
zvasya: Rework it to PInvokeOverride. But I encountered a problem with incorrect generation by SilkTouch. So I add new attribute PreventSilkTouchBugAttribute).

I also made sure that this is not a Roslyn problem: https://github.com/dotnet/roslyn/issues/59496
Perhaps it can be solved in SilkTouch simply by not using the all old class modifiers, but by taking only partial.

build/nuke/Native/MoltenVK.cs
```diff
@@ -0,0 +1,42 @@
...
+                    {
+                        var runtimes = RootDirectory / "src" / "Native" / "Silk.NET.MoltenVK.Native" / "runtimes";
+
+                        InheritedShell($"./fetchDependencies --ios --iossim --maccat", MoltenVKPath).AssertZeroExitCode();
+                        InheritedShell($"make ios iossim maccat", MoltenVKPath).AssertZeroExitCode();
+
+                        var xcFrameworkDir = MoltenVKPath / "Package" / "Release" / "MoltenVK" / "MoltenVK.xcframework";
+                        const string iosDir = "ios-arm64";
+                        const string simulatorDir = "ios-arm64_x86_64-simulator";
+                        const string macCatalystDir = "ios-arm64_x86_64-maccatalyst";
```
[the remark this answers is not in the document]
zvasya: Replace libraries in new folders.
But I chose "ios" instead of "ios-arm64", because otherwise maccatalyst tries to link it to itself...

src/Native/Silk.NET.MoltenVK.Native/Silk.NET.MoltenVK.Native.csproj
Perksey: Looks good, thanks!

src/Vulkan/Silk.NET.Vulkan/PreventSilkTouchBugAttribute.cs
```diff
@@ -0,0 +1,11 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System;
+
+namespace Silk.NET.Vulkan;
+
+public class PreventSilkTouchBugAttribute : Attribute
```
Perksey, line 8:
```suggestion
internal class PreventSilkTouchBugAttribute : Attribute
```

Perksey (changes requested):
Good to go after this, thank you so much for your work!

HurricanKai:
I'm probably missing context here - I'm not really surprised that this works, but I'd appreciate if the relevant context is added here or even better the new attribute has some comment to explain why it's necessary.

(Using __Internal works fine for Sdl, right? Why is the attribute not necessary there?)

zvasya:
> I'm probably missing context here - I'm not really surprised that this works, but I'd appreciate if the relevant context is added here or even better the new attribute has some comment to explain why it's necessary.
> 
> (Using __Internal works fine for Sdl, right? Why is the attribute not necessary there?)

We discuss this with [Perksey](https://github.com/Perksey) at https://discord.com/channels/521092042781229087/607634593201520651/1149730272754020384

Perksey (approved):

PR #1647, opened 2023-09-10
Update SDL2 binaries

silkdotnet:


Playboi2023:
Thanks Silk Net nem binaries 1s and 0s and a KiloNova

PR #1678, opened 2023-09-24
Update vkd3d + vkd3d-compiler

### LINKED ISSUE #1634: Possible calling convention issue with ID3D11Blob (vkd3d) on linux x64 using dxvk
# Summary
Having an issue running the d3d sample using dxvk.

The main issue stems from ID3D10Blob returning garbage. 
From investigating it looks like the virtual functions used in ID3D10Blob from vkd3d uses microsoft calling conventon, while im presuming it gets called using a linux one.

debugging
values

**NOTE: images are not the same process instance.** If it was it should have returned, 0x5600192d8970 for data ptr and 0x258 for size.

# Steps to reproduce
- Platform: Desktop Linux(arch) x86_64
- Framework Version: net7.0, Not sure if rider uses mono or dotnet to compile.
- API: D3D11(DXVK),  GLFW window
- API Version: D3D_FEATURE_LEVEL_11_0

1. Copied the code from the [d3d11 sample](https://github.com/dotnet/Silk.NET/blob/main/examples/CSharp/Direct3D11%20Tutorials/Tutorial%201.2%20-%20Hello%20quad/Program.cs) into a rider c# project.
2. install nuget libraries (for dxvk with GLFW window)
3. run code

# Comments
Issue isnt techinically directly in this project as its in vkd3d this project uses, specfiically with [`STDMETHODCALLTYPE`](https://github.com/Beyley/vkd3d/blob/master/include/vkd3d_windows.h#L189)

Beyley:
# Summary of the PR
Update vkd3d to my latest fork, and to Zig `0.12.0-dev.405+c1e94b28a`

Closes #1634

# Further Comments
My fork now has the ms_abi patch applied, so it should fix the ABI issues we sometimes saw


src/Microsoft/Vkd3dCompiler/lib.c
```diff
@@ -132,6 +195,17 @@ extern int32_t D3DCompile2(const void *data, void* data_size, const char *filena
         option->value = true;
     }
 
+    if (flags & (D3DCOMPILE_PACK_MATRIX_ROW_MAJOR | D3DCOMPILE_PACK_MATRIX_COLUMN_MAJOR))
```
Perksey, line 198: Where does the code in this file actually come from?

Beyley: Comes from this file

https://github.com/Beyley/vkd3d/blob/9c2d740607a3234a5a98c3f084c2326569f6e183/libs/vkd3d-utils/vkd3d_utils_main.c#L265

All i have done is strip out the irrelevant parts

Perksey: Feels like a bit too much human intervention (and code that isn't ours, in our repo, effectively untracked)

I know it's cursed, but we could do something like
```c
#include "vkd3d_utils_main.c"
#undef D3D12GetDebugInterface
#undef D3D12CreateDeviceVKD3D
#undef D3D12CreateRootSignatureDeserializer
#undef D3D12CreateVersionedRootSignatureDeserializer
// etc...
```

Alternatively we could leave the irrelevant parts in... just not really keen on us having our own "buffer" of upstream

Beyley: > Feels like a bit too much human intervention (and code that isn't ours, in our repo, effectively untracked)
> 
> I know it's cursed, but we could do something like
> 
> ```c
> #include "vkd3d_utils_main.c"
> #undef D3D12GetDebugInterface
> #undef D3D12CreateDeviceVKD3D
> #undef D3D12CreateRootSignatureDeserializer
> #undef D3D12CreateVersionedRootSignatureDeserializer
> // etc...
> ```
> 
> Alternatively we could leave the irrelevant parts in... just not really keen on us having our own "buffer" of upstream

The irrelevant parts more meant it complicated the build process, since stuff like the logging macros were causing a billion undefined symbol errors i was not too keen on debugging/fixing, but i can give using the original source a shot with some well placed `def` and `undef`  if desired

Perksey: Yeah, the main thing is I had no idea this existed (yes I know clearly I don't pay that much attention) and wouldn't have thought to check here when updating vkd3d, and the same will definitely be said for whoever comes after you & I.

Eliminating as much human error as possible should be key, and this I feel like is definitely human-error-prone

Beyley: > Yeah, the main thing is I had no idea this existed (yes I know clearly I don't pay that much attention) and wouldn't have thought to check here when updating vkd3d, and the same will definitely be said for whoever comes after you & I.
> 
> Eliminating as much human error as possible should be key, and this I feel like is definitely human-error-prone

to be fair i *also* had to reference my own commits to figure out what i had done, so its a bit error prone yeah, i'll try to get the original source working at some point soon (and also update past `.405` so we get them LLVM 17 benefits)

Perksey: Notably 4 months ago they removed ms_abi so we are likely to have fixed #1634

Perksey: ...although you probably already knew that as you've already linked the issue 🤦

Perksey:
Converting back to draft for the time being

Perksey:
2.18 snap is on Tuesday, is this likely to make it in?

Beyley:
> 2.18 snap is on Tuesday, is this likely to make it in?

uhhhhh maybe? ive been very lazy and tired the past week so it depends on if i have the energy to work on it before then (which itself depends on my AC getting fixed, i dont have the energy to fight C when its already like 30 degrees C in my house)

Perksey:
Assuming this is probably stale?

Perksey:
closing as stale

PR #1680, opened 2023-09-27
This is the October 2023 Update (v2.18.0)

Perksey:
The `Use` functions rely on reflection doing the right things, which isn't necessarily correct on AOT.


Perksey:
Need to do input as well

Perksey:
Need to write the patch notes tomorrow and regen bindings. This shall be the last manual update, I decree it!

Perksey:
This is ready for review & release

Beyley (approved):
yas bros... i'ts lookoing good

PR #1684, opened 2023-10-01
Add CI commit glob for MoltenVK binaries

yodasoda1219:
# Summary of the PR
Adds a glob for MoltenVK binaries built during CI runs. This fixes the issue of CI workflows excluding built binaries from commits.

No related issues.

# Further Comments


Perksey (approved):
Thanks

PR #1692, opened 2023-10-02
Use Release on feed-bound builds only

Perksey:
Evaluating whether this will make PR checks faster.


Perksey:
No noticeable improvement

PR #1700, opened 2023-10-09
Add *SOME* WinRT support

Perksey:
Adds IInspectable and friends, winstring support, and a few WinRT interop APIs.


Perksey:
Ready for review.

src/Core/Silk.NET.Core/Native/Structs/WinString.cs
```diff
@@ -0,0 +1,90 @@
...
+        [DllImport("kernel32.dll")]
+        static extern void* HeapAlloc(void* heap, uint dwFlags, nuint dwBytes);
+
+        HString = (Header*)HeapAlloc(GetProcessHeap(), 0, (nuint)sizeof(Header) + 2 * nChars);
+        HString->Flags = 0;
+        HString->Length = (uint)nChars;
+        HString->Reserved0 = 0;
+        HString->Reserved1 = 0;
+        HString->Data = &HString->BufferStart;
+        HString->RefCount = 1;
```
Beyley, line 43: Is there a reason we are using `HeapAlloc` specifically here? If we are required to, id prefer if we cache GetProcessHeap somewhere, to remove the need for the call to it *every* WinString creation

Beyley (commented):
Only one small question, aside from that i have no gripes with the implementation

src/Core/Silk.NET.Core/Native/Structs/WinString.cs
Perksey: _Technically_ you can replace `malloc`, though please don't. This is for consistency with the windows-rs implementation of HSTRING.

Beyley (approved):
i'ts looking good bros....

PR #1708, opened 2023-10-12
Update all bindings

Beyley:
# Summary of the PR
Update all bindings


Perksey (commented):
Seems legit, can you just do a couple of things:
- Make sure the native packages that aren't automatically versioned have their version numbers incremented
- Investigate the impact of not cloning submodules recursively for the Git-versioned ones? Perhaps we should just do a non-recursive `submodule init` for CI. It's supposed to take the date of the latest commit in the cloned submodule but I fear that it's probably taking the silk commit date, which isn't bad but it's probably not what we want. Sorry I've only just thought of this now.

The latter is not a show-stopper as it's not a You Problem, we can merge without this.

Beyley:
> Seems legit, can you just do a couple of things:
> 
>     * Make sure the native packages that aren't automatically versioned have their version numbers incremented
> 
>     * Investigate the impact of not cloning submodules recursively for the Git-versioned ones? Perhaps we should just do a non-recursive `submodule init` for CI. It's supposed to take the date of the latest commit in the cloned submodule but I fear that it's probably taking the silk commit date, which isn't bad but it's probably not what we want. Sorry I've only just thought of this now.
> 
> 
> The latter is not a show-stopper as it's not a You Problem, we can merge without this.

I still need to go through and fix all the failed native CI runs, a lot of them seem to just be "out of space" or a simple "build process changed", so i'll mark as draft until then

Perksey:
ty

Perksey:
CI is greeeeeeeeeeen, do we need the native lib updates?

Beyley:
> CI is greeeeeeeeeeen, do we need the native lib updates?

Looking through vkd3d is the main thing still missing, but that will come in the other PR ig, so yeah, we can open this for proper review

Perksey:
I think my original comments still apply.

Beyley:
> I think my original comments still apply.

Completely forgot about that yeah, i'll bump the versions when i get the time

Perksey (approved):

Perksey:
@dotnet-policy-service rerun

PR #1723, opened 2023-10-12
Add support for OpenAL EffectTarget extension

okaniku:
# Summary of the PR
This change adds support for the OpenAL Soft EffectTarget extension, which allows effects to be chained.


Perksey (approved):
Thank you so much!

PR #1727, opened 2023-10-12
Proposal containing the results of the SilkX experiment

Perksey:
This shall remain in draft until the SilkX experiment has concluded!


documentation/proposals/Proposal - Generation of Library Sources and PInvoke Mechanisms.md
```diff
@@ -6,283 +6,268 @@ Proposal design for a platform invoke (P/Invoke) mechanism for Silk.NET 3.0.
...
 
-Note to the reader: Given preprocessor directives are processed at parse time in Roslyn, both of those last requirements are basically benign.
+The pivotal proposal around which the SilkX experiment revolves is solving the excessive overload problem by having no overloads at all. Namely, Silk.NET.Core will have a set of "pointer-like" types that can be implicitly casted to by a multitude of different types. This gives the user the same flexibility as having all different permutations of overloaded parameters generated without having to do so, and thereby alleviates the requirement of having overloads at all.
 
-#### Custom Call Style: Procedure Address Expressions
+`Ptr` and `Mut` `ref struct`s **shall** be defined wrapping a `ref` and `ref readonly` respectively. These types **shall** have generic and non-generic variants. This shall be repeated up to 3 pointer dimensions, taking into account C's mutability rules in each instantiation. This means that there will be 14 top-level generic structs and 14 top-level non-generic structs. These **should** be auto-generated to ensure they remain maintainable. IL weaving **may** be required for the higher dimensions.
```
Khitiara, line 73: backwards

documentation/proposals/Proposal - Generation of Library Sources and PInvoke Mechanisms.md
```diff
@@ -6,283 +6,268 @@ Proposal design for a platform invoke (P/Invoke) mechanism for Silk.NET 3.0.
...
 
-Custom code may be used as a call style by providing a pointer to SilkTouch using a Procedure Address Expression. This is useful in scenarios such as COM interop.
+Unless deemed inappropriate, inapplicable, and/or infeasible by the Silk.NET team (the team reserved the right to do so without Working Group approval), the types **should** have the following characteristics:
+- An instance **must** be constructable from a `ref T` or `ref readonly T` (depending on the outer dimension mutability) where `T` is either the generic type (or `byte` for non-generic variants), a "pointer-like" type of the inner dimension of type being constructed.
+- An indexer **must** be present accepting a `nuint` for the index, returning a `ref T` or `ref readonly T` (depending on the outer dimension mutability) where `T` is either the generic type (or `byte` for non-generic variants), a "pointer-like" type of the inner dimension of type being indexed.
```
Khitiara, line 77: clarify ability to return other Ptr-like types

Khitiara, line 77: permit multiple-argument indexers on higher-arity Ptr types

documentation/proposals/Proposal - Generation of Library Sources and PInvoke Mechanisms.md
```diff
@@ -6,283 +6,268 @@ Proposal design for a platform invoke (P/Invoke) mechanism for Silk.NET 3.0.
...
+- A raw pointer of the same dimensions **must** be implicitly convertible to the type.
+- `NullPtr` **must** be implicitly convertible to the type, and the returned pointer must represent a null reference.
+- An explicit operator **must** be present to unsafely convert the reference represented by the pointer to a raw pointer without the user using `fixed`/`GetPinnableReference`.
+- An explicit operator **must** be present to unsafely convert the reference represented by the pointer to a void pointer without the user using `fixed`/`GetPinnableReference`.
+- For generic pointer types, an implicit operator **must** be present to convert the pointer type to a `string` (or an array of strings of the inner dimension's jaggedness i.e. `PtrPtrPtr<byte>` becomes `string[][]`). Much like `ReadOnlySpan<char>`, `ToString` **must** call this implicitly if the type is a single-dimension pointer to a string pointee type. The implicit cast **may** throw an exception if the pointee type is not a string pointee type. This is because we can't constrain the type used on implicit operators on generic types.
```
Khitiara, line 86: still feel uneasy with implicit string conversions for allocation/perf reasons but that would need benchmarking and further analysis and consideration

documentation/proposals/Proposal - Generation of Library Sources and PInvoke Mechanisms.md
```diff
@@ -6,283 +6,268 @@ Proposal design for a platform invoke (P/Invoke) mechanism for Silk.NET 3.0.
...
 
-## SilkTouch Overloader
-The Overloader, one of the two final stages whose resources **MUST** be entirely independent of eachother, creates overloads of functions that expose a more user-friendly interface than the function it overloads, and do appropriate marshalling to lower the parameter types used down to the original function's types.
+The binding class **shall** itself implement the top-level interface, using the `getProcAddress` delegate stored from the `Create` method to implement the native calls using a direct function pointer call.
 
-The Overloader **MUST** be able to be used on any function and not be tied to any of the Emitter's constraints.
+The binding class **shall** also contain shorthand functions for calling the static functions contained within the static subinterface using the *static default*.
 
-The Overloader **MUST** be able to be invoked via the SilkTouch CLI and **MAY** be able to be invoked via an incremental Roslyn source generator.
+**INFORMATIVE:** It is undecided whether we want to implement the static subinterface on the binding class itself as well.
```
Khitiara, line 258: I vote yes to this

Khitiara (commented):
overall LGTM

documentation/proposals/Proposal - Generation of Library Sources and PInvoke Mechanisms.md
Perksey: Yeah fair, just wasn't sure whether we wanted what is essentially a proxy to implement the interface itself, but yeah I don't see why not.

documentation/proposals/Proposal - Generation of Library Sources and PInvoke Mechanisms.md
Perksey: If you don't want to marshal a string just don't pass a string 🤷

documentation/proposals/Proposal - Generation of Library Sources and PInvoke Mechanisms.md
Perksey: I thought this was implied by `inner dimension`, but alright will reword

Khitiara: yeah i just wanted that clarified in the wording, i think the parens around the `byte`/non-generic version hides it a bit at least to my reading

documentation/proposals/Proposal - Generation of Library Sources and PInvoke Mechanisms.md
```diff
@@ -6,283 +6,268 @@ Proposal design for a platform invoke (P/Invoke) mechanism for Silk.NET 3.0.
...
+    // Static Interface
+    public static byte* ToLower(byte* str) => DllImport.ToLower(str);
+    public static Ptr<byte> ToLower(Ptr<byte> str) => DllImport.ToLower(str);
+    public static void FreeResult(byte* str) => DllImport.FreeResult(str);
+    public static void FreeResult(Ptr<byte> str) => DllImport.FreeResult(str);
 }
 ```
 
-Procedure Address Methods are method groups (or an otherwise callable expression) within the scope of the method that **MUST** return a `void*`, `nint`, or `IntPtr` when invoked. This is the actual address in memory of the function being invoked.
+**INFORMATIVE:** `getProcAddress` delegates replace "native contexts".
```
Perksey, line 221: Maxine has expressed interest in Discord in not doing this, and maintaining an INativeContext interface in some form rather than forcing delegates to be used.

documentation/proposals/Proposal - Generation of Library Sources and PInvoke Mechanisms.md
Perksey:
```suggestion
`Ptr` and `Mut` `ref struct`s **shall** be defined wrapping a `ref readonly` and `ref` respectively. These types **shall** have generic and non-generic variants. This shall be repeated up to 3 pointer dimensions, taking into account C's mutability rules in each instantiation. This means that there will be 14 top-level generic structs and 14 top-level non-generic structs. These **should** be auto-generated to ensure they remain maintainable. IL weaving **may** be required for the higher dimensions.
```

documentation/proposals/Proposal - Generation of Library Sources and PInvoke Mechanisms.md
Perksey:
```suggestion
- An indexer **must** be present accepting a `nuint` for the index, returning a `ref T` or `ref readonly T` (depending on the outer dimension mutability) where `T` is either the generic type (or `byte` for non-generic variants), a "pointer-like" type of the inner dimension (i.e. a `PtrPtr<T>`'s indexer will return a `Ptr<T>`) of type being indexed.
```

documentation/proposals/Proposal - Generation of Library Sources and PInvoke Mechanisms.md
Perksey:
```suggestion
The binding class **shall** implement the static subinterface (i.e. to proxy calls to the *static default* for ease of use).
```

uwx:
Notes from Discord chat:

> > > For this requirement, how do we implicitly handle this without a known length? For a C-string we can search for the null-terminator, but not all strings will necessarily follow that pattern, and not every byte ptr will be a c-string leading to a crash when VS inspects it. This also feels like something that should be in an extension method since really it should only be defined for byte, sbyte, short, ushort, int, uint as other generic types will not convert nicely. 
> > 
> > what is the motivation for the implicit operator? having to do new string(x) makes more sense to me
> > like, in C++ there is no implicit cast from char* to std::string
> > because the process of that conversion is not determinable
> > you have a pointer to memory, and i think it would be unwise to make assumptions about what that pointer means
> 
> ok lets make it an explicit operator for now 

Also, need a thorough discussion about left-to-right order of PtrPtrPtr chains, and what Mut and Ptr should be named for clarity

documentation/proposals/Proposal - Generation of Library Sources and PInvoke Mechanisms.md
```diff
@@ -6,283 +6,268 @@ Proposal design for a platform invoke (P/Invoke) mechanism for Silk.NET 3.0.
...
+`Ptr` and `Mut` `ref struct`s **shall** be defined wrapping a `ref readonly` and `ref` respectively. These types **shall** have generic and non-generic variants. This shall be repeated up to 3 pointer dimensions, taking into account C's mutability rules in each instantiation. This means that there will be 14 top-level generic structs and 14 top-level non-generic structs. These **should** be auto-generated to ensure they remain maintainable. IL weaving **may** be required for the higher dimensions.
 
-Custom code may be used as a call style by providing a pointer to SilkTouch using a Procedure Address Expression. This is useful in scenarios such as COM interop.
+Unless deemed inappropriate, inapplicable, and/or infeasible by the Silk.NET team (the team reserved the right to do so without Working Group approval), the types **should** have the following characteristics:
+- An instance **must** be constructable from a `ref T` or `ref readonly T` (depending on the outer dimension mutability) where `T` is either the generic type (or `byte` for non-generic variants), a "pointer-like" type of the inner dimension of type being constructed.
```
curin, line 76: Regarding dimension mutability, more explicit definition maybe good. As of right now inner-most is first and outermost is last. This may also not be the most intuitive as I read it in the reverse. So MutPtr<byte> would be byte** const rather than byte * const *.

documentation/proposals/Proposal - Generation of Library Sources and PInvoke Mechanisms.md
curin: Maxine brought this up in the discord, this may be better as an explicit operator as assuming a Ptr<T> equates to a string is dangerous.

curin: It is also probably good to limit this functionality to only String types (byte, sbyte, ushort, short, uint, int). Using extensions would work for a method, but we can't directly override an operator. It may be best to use an explicit function like ReadToString(), then the operator can call the function and we override that for specific types otherwise throw an exception.

Perksey: Agreed, I think we could define `ReadToString` in `SilkMarshal` as an extension method for `Ptr`, `Mut`, `Ptr<T>`, and `Mut<T>` where T is one of: byte, sbyte, ushort, short, char, uint, int

documentation/proposals/Proposal - Generation of Library Sources and PInvoke Mechanisms.md
```diff
@@ -6,283 +6,268 @@ Proposal design for a platform invoke (P/Invoke) mechanism for Silk.NET 3.0.
...
 
-#### Custom Call Style: Procedure Address Expressions
+`Ptr` and `Mut` `ref struct`s **shall** be defined wrapping a `ref readonly` and `ref` respectively. These types **shall** have generic and non-generic variants. This shall be repeated up to 3 pointer dimensions, taking into account C's mutability rules in each instantiation. This means that there will be 14 top-level generic structs and 14 top-level non-generic structs. These **should** be auto-generated to ensure they remain maintainable. IL weaving **may** be required for the higher dimensions.
 
-Custom code may be used as a call style by providing a pointer to SilkTouch using a Procedure Address Expression. This is useful in scenarios such as COM interop.
+Unless deemed inappropriate, inapplicable, and/or infeasible by the Silk.NET team (the team reserved the right to do so without Working Group approval), the types **should** have the following characteristics:
```
Perksey, line 75: In the case of a generic pointer type, implicitly convertible to its non-generic counterpart?

curin: This should be easy enough as we would just be passing the InteriorRef as the Ref for the non-generic consturctor.

Perksey:
# Maths

- We agree with the addition to add scalar operations over vectors (i.e. Vector4<T> * T does X*T, Y*T, ...)
- Add an analyser for encouraging the most correct and most efficient type instead of using sub-optimal types. 
- Ensure we've documented that.
- Vector * Matrix?
    - assuming row major, not allowing matrix * vector (only vector * matrix)
    - does it make sense to just do it the way game engines do it and the way 2.0 does it?
    - possibly technically invalid mathematically, even if common programmatically.
    - easy to hit corner case that not all users may understand.
- Ensuring we have no upcasting and downcasting operators that could have unexpected behaviour.
- System.Numerics is strictly right hand, which is the same as XNA and somewhat default for OpenGL users. This differs with DirectX that is typically left hand.
    - Happy with having explicit LH and RH functions.
- Missing matrix APIs that were added in .NET 8 to System.Numerics
    - CreateLookTo (additional to CreateLookAt)
    - CreateViewport
    - CreateBillboardLH
    - CreateConstrainedBillboardLH
    - CreateBillboardRH
    - CreateConstrainedBillboardRH
- Investigate newer less-hated alternatives to quaternions for future (see Freya Holmer talks)
- Possibly have a general Transform struct in the future?
- Rectangles and boxes are obscured with regards to internal layout, bindings to have interchange types that implicitly cast to the mathematical constructs that are represented by the Box/Rectangle types.
- Investigate in the future if we can either add an Origin property for compatibility or an analyser for users to know that Min is the new Origin.
- Approved notwithstanding the missing APIs.

Perksey:
# SilkTouchX

- We discussed a particular problematic case where RegisterClassEx returns an atom which is later reinterpreted to be a pointer - a debugger will explode when inspecting this pointer as it is not necessarily 
- Where ReadOnlySpan<char> represents a string and we don't just want to pass the ref as-is (i.e. we want to add the null terminator like we do for string).
- Generally we think that providing a tool that works 90% of the time is fine, the unsafe overloads are always there, but we'd worry about users making incorrect assumptions and we can probably do implicit behaviour for that final 10%.
- Require that users manually encoding strings add that null terminator and document this. Our implicit ones do the right thing.
- Don't allow ref types to throw an error when being handed something that isn't a valid pointer.
- Approved provided that we:
    - make unsafe available
    - special case ROSpan<char> as above
- Future discussions need to be had on Vulkan implementation intricacies (getProcAddr) and also the addition of "complex" overloads.

Perksey:
@curin @uwx @khitiara If any of you want to put up a PR for the CreateLookTo (additional to CreateLookAt), CreateViewport, CreateBillboardLH, CreateConstrainedBillboardLH, CreateBillboardRH, CreateConstrainedBillboardRH it'd be much appreciated sop we can get this in.

I'll update SilkTouchX now.

Perksey:
@dotnet/silk-dotnet @uwx @Khitiara This is ready for merge I think.

Khitiara (approved):
LGTM

Perksey:
Is it possible to get an approval here? The revisions to the proposals are as discussed in the WG meeting.

curin (approved):

Perksey:
NOOOOOO NOT THE NATIVE BUILD JOBS

Perksey:
Build is not triggering because there have been no source changes, merging manually.

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,387 @@
...
+- System.Numerics is strictly right hand, which is the same as XNA and somewhat default for OpenGL users. This differs with DirectX that is typically left hand.
+    - Happy with having explicit LH and RH functions.
+- Missing matrix APIs that were added in .NET 8 to System.Numerics
+    - CreateLookTo (additional to CreateLookAt)
+    - CreateViewport
+    - CreateBillboardLH
+    - CreateConstrainedBillboardLH
+    - CreateBillboardRH
+    - CreateConstrainedBillboardRH
+- Investigate newer less-hated alternatives to quaternions for future (see Freya Holmer talks)
```
alexrp, line 383: @Perksey don't suppose you have a link? Quaternions are, of course, ontologically evil, so I would be interested to hear about viable alternatives. :eyes:

Perksey: I think it was @HurricanKai or @ThomasMi that raised this point (and alluded to things like rotors as well)

PR #1791, opened 2023-11-11
Proposal - 3.0 Managed Chain Changes

Khitiara:
# Summary of the PR
A proposal for changes, some of them breaking, to the managed chain utilities for 3.0.

As discussed in discord with the SilkX team.


Perksey:
Can you add .md to the filename

Perksey:
No pre-WG comments here

curin:
No comments from me, either, but VK is also not my area of expertise, so I'm just going off of the proposal working as explained.

Perksey:
- We believe that the chain abstractions break down significantly anyway when used as an output with layers that can possibly inject additional PNexts that aren't reflected in these abstractions.
- As a result, this chain abstraction only really works with inputs in every case.
- UntypedHeadPtr being changed from BaseInStructure is contrary to this assumption.
- Why is it IStructuredType instead of ITypedStructure? Changing this doesn't affect too much, but leaving it as is doesn't affect too much either.
- 3.0 is the opportunity to fix long-standing issues like this as it's already massively breaking, so we can do silly changes like this.
- IStructuredType is already broken with this proposal, so let's just rename it anyway.
- We believe that IBaseStructure is probably the most sensible name given how Vulkan names this.
- Everything else looks fine, approved notwithstanding the above.

Perksey:
# Maths

- We agree with the addition to add scalar operations over vectors (i.e. Vector4<T> * T does X*T, Y*T, ...)
- Add an analyser for encouraging the most correct and most efficient type instead of using sub-optimal types. 
- Ensure we've documented that.
- Vector * Matrix?
    - assuming row major, not allowing matrix * vector (only vector * matrix)
    - does it make sense to just do it the way game engines do it and the way 2.0 does it?
    - possibly technically invalid mathematically, even if common programmatically.
    - easy to hit corner case that not all users may understand.
- Ensuring we have no upcasting and downcasting operators that could have unexpected behaviour.
- System.Numerics is strictly right hand, which is the same as XNA and somewhat default for OpenGL users. This differs with DirectX that is typically left hand.
    - Happy with having explicit LH and RH functions.
- Missing matrix APIs that were added in .NET 8 to System.Numerics
    - CreateLookTo (additional to CreateLookAt)
    - CreateViewport
    - CreateBillboardLH
    - CreateConstrainedBillboardLH
    - CreateBillboardRH
    - CreateConstrainedBillboardRH
- Investigate newer less-hated alternatives to quaternions for future (see Freya Holmer talks)
- Possibly have a general Transform struct in the future?
- Rectangles and boxes are obscured with regards to internal layout, bindings to have interchange types that implicitly cast to the mathematical constructs that are represented by the Box/Rectangle types.
- Investigate in the future if we can either add an Origin property for compatibility or an analyser for users to know that Min is the new Origin.
- Approved notwithstanding the missing APIs.

PR #1792, opened 2023-11-12
Update SilkX Proposals

curin:
# Summary of the PR
This PR updates the proposals to reflect work being done on SilkX

Proposal - Generation Of Library Sources and PInvoke Mechanism
Proposal - Generic Math


documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1182 @@
...
+
+# Vector Types
+
+The main types defined for this proposal are two sets of vector types, `VectorNI<T>` and `VectorNF<T>` where `N` defines the dimensionality of the vector between 2 and 5. Aside from some notable exceptions the vector types will be expected to implement a similar api to the correlating one below (`VectorNI<T>` will follow `Vector2I<T>` and the `VectorNF<T>` will follow `Vector2F<T>`).
+
+Aside from the api below, the following constraints should also be followed:
+- Three Dimensional Vectors should implement a CrossProduct function which takes another vector and returns the cross product with our original vector. A static implementation of this function should be available as well.
```
uwx, line 30: The function should be called Cross for consistency with 2.x and System.Numerics.

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1182 @@
...
+- Every Dimension should have the relevant number of properties (X and Y for Vector2, and X,Y,Z,W,V for Vector5) and relevant unit vectors
+- Constructors for higher dimensions should include lower dimension variants that use the lower dimensions for their specific components (vector2 -> X,Y)
+
+```cs
+public readonly record struct Vector2I<TScalar> : IEquatable<Vector2I<TScalar>>, 
+                           IEquatable<Vector2>,
```
uwx, line 36: Equality comparison to System.Numerics.Vector2 might still be debatable.

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1182 @@
...
+    public Vector2I<TScalar> Abs();
+    public void CopyTo(TScalar[] array);
+    public void CopyTo(TScalar[] array, int index);
+
+    public static Vector2I<TScalar> Dot (Vector2I<TScalar> left, Vector2I<TScalar> right);
+    public static Vector2I<TScalar> DistanceSquared (Vector2I<TScalar> left, Vector2I<TScalar> right);
+    public static Vector2I<TScalar> Clamp (Vector2I<TScalar> value, Vector2I<TScalar> min, Vector2I<TScalar> max);
+    public static Vector2I<TScalar> Min (Vector2I<TScalar> value1, Vector2I<TScalar> value2);
+    public static Vector2I<TScalar> Max (Vector2I<TScalar> value1, Vector2I<TScalar> value2);
+    public static Vector2I<TScalar> Abs (Vector2I<TScalar> value);
```
uwx, line 97: These methods will be provided on a static class. They will have explicit implementations of the IVector static abstract methods that simply delegate to the static class.

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1182 @@
...
+    public static Vector2I<TScalar> operator +(Vector2I<TScalar> value);
+    public static Vector2I<TScalar> operator -(Vector2I<TScalar> value);
+
+    public static Vector2I<TScalar> operator &(Vector2I<TScalar> left, Vector2I<TScalar> right);
+    public static Vector2I<TScalar> operator |(Vector2I<TScalar> left, Vector2I<TScalar> right);
+    public static Vector2I<TScalar> operator ^(Vector2I<TScalar> left, Vector2I<TScalar> right);
+
+    public static Vector2I<TScalar> operator ~(Vector2I<TScalar> value);
+
+    public TScalar LengthSquared { get; }
```
uwx, line 84: This may be a method or property, we haven't decided yet.

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1182 @@
...
+
+    public Vector2I<TScalar> Dot(Vector2I<TScalar> other);
+    public TScalar DistanceSquared(Vector2I<TScalar> other);
+    public Vector2I<TScalar> Abs();
+    public void CopyTo(TScalar[] array);
+    public void CopyTo(TScalar[] array, int index);
```
uwx, line 90: These methods will be provided via extension methods to reduce minimum JIT generic specialization code size. They will have explicit implementations of the IVector methods that simply delegate to the extension methods.

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1182 @@
...
+
+    public static Vector2I<TScalar> UnitX { get; }
+    public static Vector2I<TScalar> UnitY { get; }
+
+    public static Vector2I<TScalar> Zero { get; }
+    public static Vector2I<TScalar> One { get; }
+
+    public TScalar this[int index] { get; }
+
+    public Vector2I(TScalar x, TScalar y);
```
uwx, line 61: Missing:

`Vector2I(TScalar value)` repeating value constructor
`Vector2I(ReadOnlySpan<TScalar> span)` span constructor

also worth noting for larger vectors that this will also be available:
`Vector3I(Vector2I<TScalar> vector, TScalar z)` etc.

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1182 @@
...
+    public void CopyTo(TScalar[] array);
+    public void CopyTo(TScalar[] array, int index);
+
+    public static Vector2F<TScalar> Dot (Vector2F<TScalar> left, Vector2F<TScalar> right);
+    public static Vector2F<TScalar> DistanceSquared (Vector2F<TScalar> left, Vector2F<TScalar> right);
+    public static Vector2F<TScalar> Distance (Vector2F<TScalar> left, Vector2F<TScalar> right);
+    public static Vector2F<TScalar> Clamp (Vector2F<TScalar> value, Vector2F<TScalar> min, Vector2F<TScalar> max);
+    public static Vector2F<TScalar> Min (Vector2F<TScalar> value1, Vector2F<TScalar> value2);
+    public static Vector2F<TScalar> Max (Vector2F<TScalar> value1, Vector2F<TScalar> value2);
+    public static Vector2F<TScalar> Abs (Vector2F<TScalar> value);
```
uwx, line 161: Ditto above notes.

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1182 @@
...
+    public TScalar Y { get; }
+
+    public static Vector2I<TScalar> UnitX { get; }
+    public static Vector2I<TScalar> UnitY { get; }
+
+    public static Vector2I<TScalar> Zero { get; }
+    public static Vector2I<TScalar> One { get; }
+
+    public TScalar this[int index] { get; }
+
```
uwx, line 60: Several methods missing here, including AsSpan(). See the previous proposal.

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1182 @@
...
+    public static Vector2F<TScalar> Dot (Vector2F<TScalar> left, Vector2F<TScalar> right);
+    public static Vector2F<TScalar> DistanceSquared (Vector2F<TScalar> left, Vector2F<TScalar> right);
+    public static Vector2F<TScalar> Distance (Vector2F<TScalar> left, Vector2F<TScalar> right);
+    public static Vector2F<TScalar> Clamp (Vector2F<TScalar> value, Vector2F<TScalar> min, Vector2F<TScalar> max);
+    public static Vector2F<TScalar> Min (Vector2F<TScalar> value1, Vector2F<TScalar> value2);
+    public static Vector2F<TScalar> Max (Vector2F<TScalar> value1, Vector2F<TScalar> value2);
+    public static Vector2F<TScalar> Abs (Vector2F<TScalar> value);
+}
+```
+At the end of this proposal is the definition for the new interfaces used in the these classes.
```
uwx, line 164: Missing the static utility classes.

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1182 @@
...
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static Quaternion<T> operator -(Quaternion<T> value);
+
+    /// <summary>Adds each element in one quaternion with its corresponding element in a second quaternion.</summary>
+    /// <param name="value1">The first quaternion.</param>
+    /// <param name="value2">The second quaternion.</param>
+    /// <returns>The quaternion that contains the summed values of <paramref name="value1" /> and <paramref name="value2" />.</returns>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static Quaternion<T> Add<T>(Quaternion<T> value1, Quaternion<T> value2)
+        where T : IRootFunctions<T>, ITrigonometricFunctions<T>;
```
uwx, line 361: It makes no sense for this to be a generic method in a generic type.

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1182 @@
...
+    
+    /// <summary>Calculates the length of the quaternion.</summary>
+    /// <returns>The computed length of the quaternion.</returns>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static T Length<T>(this Quaternion<T> self) where T : IRootFunctions<T>, ITrigonometricFunctions<T>;
+
+    /// <summary>Calculates the squared length of the quaternion.</summary>
+    /// <returns>The length squared of the quaternion.</returns>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static T LengthSquared<T>(this Quaternion<T> self) where T : IRootFunctions<T>, ITrigonometricFunctions<T>;
```
uwx, line 500: See prior notes about static methods and how they are to be implemented.

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1182 @@
...
+- CircleF
+- CircleI
+- CubeF
+- CubeI
+- PlaneF
+- PlaneI
+- Ray2DF
+- Ray2DI
+- Ray3DF
+- Ray3DI
```
uwx, line 520: Drop the D, as we're already doing so for vectors. Possibly rename Box to Aabb or AxisAlignedBoundingBox, as the current name is rather confusing as we have both Box and Rectangle.

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1182 @@
...
+
+The following Geometric Types are defined:
+- Box2DF
+- Box2DI
+- Box3DF
+- Box3DI
+- CircleF
+- CircleI
+- CubeF
+- CubeI
```
uwx, line 514: Really this is a rectangular prism, not a cube.

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1182 @@
...
+{
+    /// <summary>
+    /// The type of scalars in this vector type.
+    /// </summary>
+    protected static abstract Type ScalarType { get; }
+    
+    /// <summary>
+    /// The amount of scalars in this vector type.
+    /// </summary>
+    protected internal static abstract int Count { get; }
```
uwx, line 558: I think this has to be changed due to naming conflict with IReadOnlyList.Count.

uwx (changes requested):

Perksey:
Paging @HurricanKai now we have a formal maths proposal.

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1182 @@
...
+- [x] Proposed
+- [ ] Discussed with API Review Board (ARB)
+- [ ] Approved
+- [ ] Implemented
+
+# Design Decisions
+- This proposal should replace the 2.X implementation of `Silk.Net.Maths`, matching `System.Numerics` where possible, with concessions for design oversights in that api.
+- This proposal assumes no knowledge of the 2.x Math library.
+- Text herein marked **INFORMATIVE** does not form a normative part of this proposal, and is for background only.
+- Within this proposal, the key words **must**, **required**, **shall**, **should**, **recommended**, **may**, **could**, and **optional** are to be interpreted as described in [RFC 2119 - Key words for use in RFCs to Indicate Requirement Levels](https://www.ietf.org/rfc/rfc2119.txt). The additional key word **optionally** is an alternate form of **optional**, for use where grammatically appropriate. These key words are highlighted in the proposal for clarity.
```
HurricanKai, line 20: I don't think this is applied correctly, if at all?
I'm a fan of this if you are motivated to put in the effort, but as we generally don't expect proposals to go unchanged through more than a handful of contributors I'm fine not doing this also.
Marking it and then not doing it isn't great imo.

HurricanKai (commented):
Unfortunately I'm not available to review for the next few days. At a glance this looks good and seems to follow the discord discussions.
I'll review this as soon as I can. Also thank you for putting in all the effort!

documentation/proposals/Proposal - Generic Math.md
curin: This is an implementation detail, which isn't meant to be part of the proposal. Whether they are done via extension methods or not shouldn't matter just that they exist on this type.

documentation/proposals/Proposal - Generic Math.md
curin: This is just to save space. I combined the static classes with their relevant type. This does not have to match the implementation. It is just to get across what functions we expect to be available.

documentation/proposals/Proposal - Generic Math.md
curin: The 2.X type was called Cube, are we changing that?

documentation/proposals/Proposal - Generic Math.md
curin: This feels more like an implementation detail for how some functions are given. I don't think we need it included in the proposal.

documentation/proposals/Proposal - Generic Math.md
curin: Fixing the first two, the third is described in a comment at the beginning of this section.

documentation/proposals/Proposal - Generic Math.md
curin: Fixed up some of the language to better match this. This was mostly included for the Informative and optional text, but I do realize some of my other phrasing didn't match, so I pulled it in line.

documentation/proposals/Proposal - Generic Math.md
Perksey: I think this is specifically referring to `Vector2F<T>.Abs` vs `Vector2F.Abs<T>` - the latter the compiler can infer the T

documentation/proposals/Proposal - Generic Math.md
curin: I did remove the D, but Box2D and Rectangle are 2 separate representations of the same object. If they sound similar that's because they are.

Perksey (commented):
Agree with all the comments above. Where I have written a comment that is applicable to all API definitions, assume that it is indeed applicable (cba to repeat myself).

documentation/proposals/Proposal - Generic Math.md
curin: Fixed

documentation/proposals/Proposal - Generic Math.md
curin: fixed

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1182 @@
...
+
+# Vector Types
+
+The main types defined for this proposal are two sets of vector types, `VectorNI<T>` and `VectorNF<T>` where `N` defines the dimensionality of the vector between 2 and 5. Aside from some notable exceptions the vector types will be expected to implement a similar api to the correlating one below (`VectorNI<T>` will follow `Vector2I<T>` and the `VectorNF<T>` will follow `Vector2F<T>`).
+
+Aside from the api below, the following constraints should also be followed:
+- Three Dimensional Vectors should implement a CrossProduct function which takes another vector and returns the cross product with our original vector. A static implementation of this function should be available as well.
+- Every Dimension should have the relevant number of properties (X and Y for Vector2, and X,Y,Z,W,V for Vector5) and relevant unit vectors
```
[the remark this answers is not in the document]
curin: Vector 5 generally gives V as the last component.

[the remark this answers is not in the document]
curin: you can do this as a property that returns a readonly ref.
`public readonly ref TScalar X => ref _X;`
or something like that.

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1182 @@
...
+- Matrix5x4I
+
+Integer Variants do not require any functions which interact with Quaternions
+
+Matricies should fulfill the following requirements:
+- be a struct
+- Stored in row major format
+- F matricies work with F vectors, and I Matricies work with I vectors
+- Both column and row vectors and individual values accessible via properties
+- An indexer that takes column indicies and outputs the value
```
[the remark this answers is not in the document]
curin: This was based on a comment Tanner made about Matricies on the CPU generally being row major for Ram and simd efficiency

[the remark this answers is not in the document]
curin: fixing this, I'm missing a few words. It was meant to say row and column indicies

documentation/proposals/Proposal - Generic Math.md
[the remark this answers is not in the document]
Perksey: I'm not sure how this works with `with`ers though, and it's worth noting if we use `record struct` parameters then this won't be generated. I'm fine either way, @tannergooding do you have thoughts?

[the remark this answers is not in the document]
uwx: > We won't be able to take a readonly ref to them unless they're fields
Why would we want to?

documentation/proposals/Proposal - Generic Math.md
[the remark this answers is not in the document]
uwx: Both are available on my test impl. System.Numerics has both array and span overloads. I guess due to implicit casting the array overloads could be dropped?

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1182 @@
...
+    ///The X component of the Vector
+    ///</summary>
+    [DataMember]
+    public TScalar X { get; }
+
+    ///<summary>
+    ///The Y component of the Vector
+    ///</summary>
+    [DataMember]
+    public TScalar Y { get; }
```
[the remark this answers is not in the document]
uwx: Doesn't work with the span constructor as when you have a primary constructor, all other constructors have to call it.

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1182 @@
...
+    /// <summary>Gets or sets the element at the specified index.</summary>
+    /// <param name="index">The index of the element to get or set.</param>
+    /// <returns>The element at <paramref name="index" />.</returns>
+    /// <exception cref="ArgumentOutOfRangeException"><paramref name="index" /> was less than zero or greater than the number of elements.</exception>
+    public T this[int index];
+
+    /// <summary>Gets a value that indicates whether the current instance is the identity quaternion.</summary>
+    /// <value><see langword="true" /> if the current instance is the identity quaternion; otherwise, <see langword="false" />.</value>
+    /// <altmember cref="Identity"/>
+    public bool IsIdentity;
```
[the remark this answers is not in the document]
uwx: Feels like it should be a method. This is a comparison operation, is it not?

documentation/proposals/Proposal - Generic Math.md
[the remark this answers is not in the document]
uwx: Euclidean coordinates go X, Y, Z, W
Texture coordinates go U, V, W

The natural progression for Euclidean coordinates seems to be X, Y, Z, W, V, U

documentation/proposals/Proposal - Generic Math.md
uwx: We're changing everything, so now is a great time to decide on better names if there are any.

documentation/proposals/Proposal - Generic Math.md
uwx: > This feels more like an implementation detail for how some functions are given. I don't think we need it included in the proposal.

The implementation for our primitive types will be explicit on the interface implementations and the actual code will be in the static classes. So this changes how you use the library unless you're implementing the interfaces yourself.

documentation/proposals/Proposal - Generic Math.md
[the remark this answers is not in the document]
Perksey: Passing to native is one reason (without pointer casting), but that could probably just be solved with a GetPinnableReference or perhaps even a Ref property so users can do .Ref[0].

documentation/proposals/Proposal - Generic Math.md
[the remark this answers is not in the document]
Perksey: a) This is a record struct. If we don’t want to use record features this should not be a record.

b) I see no reason why this shouldn’t be the case anyway.

documentation/proposals/Proposal - Generic Math.md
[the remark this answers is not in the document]
Perksey: Same argument as Length/LengthSquared I guess, but in any case it shouldn’t be a field as the proposal indicates.

documentation/proposals/Proposal - Generic Math.md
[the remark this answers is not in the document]
Perksey: BCL generally has a policy to accept an array where there’s a span to ensure users that may not be aware of the implicit cast know they can pass an array (and also for historical reasons)

This is not of concern for us specifically.

documentation/proposals/Proposal - Generic Math.md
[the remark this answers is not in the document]
Perksey: I feel like it’d be nice to have some more academic reasoning. But then again I’m still struggling with Vector5 even existing haha.

documentation/proposals/Proposal - Generic Math.md
Perksey: Could rename to Extent?

documentation/proposals/Proposal - Generic Math.md
Perksey: I feel like this may be a bit too much bikeshedding and also removing some familiarity without great reasoning for it. Our Rectangle could also be a Square, just like Cube could be a Cuboid. I don’t have strong feelings here, and because of this I think familiarity (for me) is overriding.

documentation/proposals/Proposal - Generic Math.md
Perksey: I’d like to keep Box2I and Box2F named as such, not calling it AABB was explicit on my part (for no particular reason I just don’t like it)

Agree with renaming Ray3DI to just Ray3I as well, the Ds were for consistency and that’s no longer the case here.

documentation/proposals/Proposal - Generic Math.md
Perksey: Why would they be explicit? This feels like it’s unnecessarily limiting usability - if they want to call it on the type directly, why not? I _personally_ think that using the static class makes for cleaner code, but who am I to tell users how to write their code…

documentation/proposals/Proposal - Generic Math.md
Perksey: I’m not sure I agree with the assessment on JIT code size here, but concur that this is an implementation detail.

documentation/proposals/Proposal - Generic Math.md
curin: Aside from just Silk familiarity, Cube is pretty commonly used for any AABB. Unreal uses this terminology, where as Unity uses Box which we already have used for the D3D representation.
On a side note, I don't think the logic of a rectangle can be a square is a good argument as mathematically a square is a rectangle, a cuboid is not a cube by definition, but the reverse is true.

documentation/proposals/Proposal - Generic Math.md
curin: Dimensions would probably be a better name here.

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1182 @@
...
+  - CreateConstrainedBillboard
+  - CreateLookAt
+  - CreateOrthographic
+  - CreateOrthographicOffCenter
+  - CreatePerspective
+  - CreatePerspectiveFieldOfView
+  - CreatePerspectiveOffCenter
+  - CreateReflection
+  - CreateWorld
+- **Optionally** include any relevant transformation for Matrix5x4 (Color Matrix)
```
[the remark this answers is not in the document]
curin: removed

documentation/proposals/Proposal - Generic Math.md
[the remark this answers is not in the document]
curin: Moved to a property for now until something more concrete is decided

documentation/proposals/Proposal - Generic Math.md
curin: I feel like Resharper assuming length refers to the number of elements is its problem, context matters. Similarly if we wrote a LineSegment class we would probably have a length property, and in context it makes sense to call it that. The length of a vector is generally considered the length mathematically. Length matches the 2.X and System.Numerics naming.

If we feel we need to rename Unreal uses Size for this property, while unity uses Magnitude, either would be good alternatives.

Perksey: An automated tool only makes this assumption because a human thought that it was a safe assumption to make. The same could be true for any of our users. Properties don’t elicit the idea of performing a calculation, rather accessing something that already exists about the value.

I’m fine with length if it’s a method per this argument. However I’d also be open to calling it something like magnitude (size would be confusing as well) if we do want to keep it as a property.

curin: I think assuming a user with context would make the same assumptions as a tool which has none is a bit of a stretch. For example, a length property on a Rectangle would not be assumed to relate to number of elements. Similarly, I don't think my first assumption with a Vector's Length is that it relates to number of elements as vectors since Mathematics gives meaning to the length of a vector.

Properties also do calculations all the time. For example, the IsIdentity property in matricies requires several comparisons. Properties are by definition obfuscation of accessor and mutator methods which may require any number of calculations or checks.

Also does a Length method not get picked up by Resharper?

I do agree that Size is a bad name.

curin: Also the Length property should have a summary which tells a user that it refers to the Euclidean Length of the Vector.

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1189 @@
...
+
+# Vector Types
+
+The main types defined for this proposal are two sets of vector types, `VectorNI<T>` and `VectorNF<T>` where `N` defines the dimensionality of the vector between 2 and 5. Aside from some notable exceptions the vector types will be expected to implement a similar api to the correlating one below (`VectorNI<T>` will follow `Vector2I<T>` and the `VectorNF<T>` will follow `Vector2F<T>`).
```
tannergooding, line 27: I raised this on Discord, but is `5` correct and needed?

I would assume this is for a `Matrix5x4` in which case it is `5x Vector4` and not `4x Vector5`, typically speaking.

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1189 @@
...
+
+Aside from the api below, the following constraints **must** also be followed:
+- Three Dimensional Vectors **must** implement a Cross function which takes another vector and returns the cross product with our original vector. A static implementation of this function should be available as well.
+- Every Dimension **must** have the relevant number of properties (X and Y for Vector2, and X,Y,Z,W,V for Vector5) and relevant unit vectors
+- Constructors for higher dimensions **must** include lower dimension variants that use the lower dimensions for their specific components (vector2 -> X,Y)
+
+```cs
+public readonly record struct Vector2I<TScalar> : IEquatable<Vector2I<TScalar>>,
```
tannergooding, line 35: nit: It doesn't really need to be `record` here.

You're really going to want to implement custom equality (perf) and formatting (globalization, and other features), so I don't think you actually "get" anything from records.

documentation/proposals/Proposal - Generic Math.md
tannergooding: I agree. It's not typical for types to be equatable with anything other than themselves (for many reasons).

Users wanting to compare with a Vector2 should rely on conversions instead.

documentation/proposals/Proposal - Generic Math.md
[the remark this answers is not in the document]
tannergooding: > I see no reason why this shouldn’t be the case anyway.

Efficiency namely. You can write a more efficient implementation for a `new Vector2I(Span<TScalar>)` API than one which defers to a primary constructor assigning fields individually.

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1189 @@
...
+    public static Vector2I<TScalar> Zero { get; }
+    public static Vector2I<TScalar> One { get; }
+
+    public TScalar this[int index] { get; }
+
+    public Vector2I(TScalar x, TScalar y);
+    public Vector2I(TScalar value);
+    public Vector2I(ReadOnlySpan<TScalar> span);
+
+    public Span<TScalar> AsSpan();
```
tannergooding, line 65: This needs `[UnscopedRef]` (maybe `[return: UnscopedRef]`, I forget) to ensure that the created span is correctly tracked.

documentation/proposals/Proposal - Generic Math.md
tannergooding: Its going to be a property in the generic ones for System.Numerics.

Length is the industry standard term (more so than magnitude) and it is an incredibly cheap computation

documentation/proposals/Proposal - Generic Math.md
tannergooding: > I’m not sure I agree with the assessment on JIT code size here

I gave some descriptions in Discord previously. But it overal impacts trimability and VM overhead due to the forced method table entries per type.

Making it a static method in another class means that you're not taking `this` as an implicit byref, means you can make it an extension method who prefer `x.Dot(y)` over `Vector2.Dot(x, y)`, improve trimming support, reduce VM overhead, etc.

tannergooding: -- Notably, it's unclear how these are an implementation detail if they are `public`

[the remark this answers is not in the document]
tannergooding: We do it to ensure non C# languages can call the code, namely.

If you aren't concerned with that, then just `Span<T>` is sufficient. If you are concerned, then both `T[]` and `T[], int start` (or `T[], int start, int length` if the size isn't fixed) should be provided.

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1189 @@
...
+    public TScalar LengthSquared { get; }
+
+    public Vector2I<TScalar> Dot(Vector2I<TScalar> other);
+    public TScalar DistanceSquared(Vector2I<TScalar> other);
+    public Vector2I<TScalar> Abs();
+    public void CopyTo(TScalar[] array);
+    public void CopyTo(TScalar[] array, int index);
+
+    public static Vector2I<TScalar> Dot (Vector2I<TScalar> left, Vector2I<TScalar> right);
+    public static Vector2I<TScalar> DistanceSquared (Vector2I<TScalar> left, Vector2I<TScalar> right);
```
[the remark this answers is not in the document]
tannergooding: Nevermind, this is the integer one. Ignore me 😄

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1189 @@
...
+    ///The Y component of the Vector
+    ///</summary>
+    [DataMember]
+    public TScalar Y { get; }
+
+    public static Vector2F<TScalar> UnitX { get; }
+    public static Vector2F<TScalar> UnitY { get; }
+
+    public static Vector2F<TScalar> Zero { get; }
+    public static Vector2F<TScalar> One { get; }
```
tannergooding, line 128: We've approved and are adding `PositiveInfinity`, `NegativeInfinity`, `NaN`, `Epsilon`, `NegativeZero`, `Pi`, `Tau`, and `E` in System.Numerics.

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1189 @@
...
+    public static Vector2F<TScalar> operator /(Vector2F<TScalar> left, TScalar right);
+    public static Vector2F<TScalar> operator %(Vector2F<TScalar> left, TScalar right);
+
+    public static Vector2F<TScalar> operator *(TScalar left, Vector2F<TScalar> right);
+
+    public static Vector2F<TScalar> operator +(Vector2F<TScalar> value);
+    public static Vector2F<TScalar> operator -(Vector2F<TScalar> value);
+
+    public TScalar LengthSquared { get; }
+    public TScalar Length();
```
tannergooding, line 154: It seems counter intuitive for `LengthSquared` to be a property but `Length` to be a method. I'd make both properties and plan to do that for the generic ones in System.Numerics when we eventually expose them.

documentation/proposals/Proposal - Generic Math.md
tannergooding: You can't do `Vector2F<T>.Abs` in a generic context, you can only do `TVector.Abs(x)` or `x.Abs()` if you've defined a relevant extension method.

So having them be explicit makes it more consistent for non-generic code as well. You just always do `Vector2F.Abs(x)` and that can optionally be an extension, allowing `x.Abs()`.

It notably also helps with trimming and overall usability of the APIs, while making it more consistent for how people access it. -- I believe the static method would also block the extension method from working, if you didn't implement it explicitly.

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1189 @@
...
+
+# Matrix Types
+
+This proposal includes the following matrix types:
+- Matrix2x2F
+- Matrix2x2I
+- Matrix2x3F
+- Matrix2x3I
+- Matrix2x4F
+- Matrix2x4I
```
tannergooding, line 182: What are `2x3` and `2x4` used for?

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1189 @@
...
+- Matrix3x2F
+- Matrix3x2I
+- Matrix3x3F
+- Matrix3x3I
+- Matrix3x4F
+- Matrix3x4I
+- Matrix4x2F
+- Matrix4x2I
```
tannergooding, line 190: Same question here. What is `3x4` used for? `4x3` is definitely common and is `4x Vector3` with the final column being implied `0, 0, 0, 1`.

I'm not familiar with any common use cases for a `4x2`, however.

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1189 @@
...
+Matricies **must** fulfill the following requirements:
+- be a struct
+- Stored in row major format
+- F matricies work with F vectors, and I Matricies work with I vectors
+- Both column and row vectors and individual values accessible via properties
+- An indexer that takes row and column indicies and outputs the value
+- Add, subtract, and multiply operators defined with Matricies of the same size
+- Multiply operators defined with compatible matricies, if the output matrix type already exists (AxB * BxC = AxC)
+- Multiply operator defined with the appropriate Vector type (vector4F * Matrix4x4F is valid)
+  - **optionally**, include multiplication operators for Vector types smaller than the appropriate type, assuming 1 for the missing dimensions (vector2F * Matrix4x4F -> (Vector2F.x, Vector2F.y, 1, 1) * Matrix4x4F)
```
tannergooding, line 209: I think assuming 1 is problematic and would differ from the behavior if you were to explicitly convert to the higher dimensioned vector.

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1189 @@
...
+- Negate Operator defined
+- Fulfills `IEquatable<T>` where `T` is the same matrix class and the System.Numerics version of the matrix if available
+- Implicit conversion to and from the System.Numerics matrix type, if available
+- Invert function for square matricies
+- GetDeterminant function for square matricies and Matrix3x2
```
tannergooding, line 214: Shouldn't this also be for `Matrix4x3` and `Matrix5x4`? Much like `Matrix3x2`, they exist as a memory/perf optimization since the last column can be inferred to be `0, 0, 0, 1` (since matrices expanded up take on the same positional value as `Identity` would have)

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1189 @@
...
+- Transpose function
+- static lerp function
+- static identity property
+- For Matrix3x2, Matrix3x3, Matrix4x3, and Matrix4x4 include the following static functions
+  - CreateBillboard
```
tannergooding, line 219: Some of these should probably have left and right handed versions.

System.Numerics is modeled after XNA, which used "right handed" by default. IIRC this is also what GLM uses.
DirectX uses left handed by default

Beyond left vs right handed, there is then the consideration on whether Z is up or Y is up (a lot of games make Z up, so that top down works for 2D and the coordinates make sense when displayed in UI) and whether the screen's top left is 0, 0 or the bottom left is 0, 0. (others are possible too)

These other nuances can sometimes influence what people think about when doing various operations, but GLM and DX do agree themselves on how values work for left vs right handed, so Silk should match that.

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1189 @@
...
+  - CreateWorld
+
+# Quaternion
+
+Quaternion is required only to fulfill the following api:
+
+```cs
+[StructLayout(LayoutKind.Sequential), DataContract, Serializable]
+public readonly struct Quaternion<T> : IEquatable<Quaternion<T>>
+    where T : INumberBase<T>, ITrigonometricFunctions<T>, IRootFunctions<T>
```
tannergooding, line 250: I don't think a quaternion should be a number anymore than a vector should.

Functionally, its basically a Vector4 that does a couple minor operations differently.

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1189 @@
...
+    public readonly T Y { get; }
+
+    /// <summary>The Z value of the vector component of the quaternion.</summary>
+    [DataMember]
+    public readonly T Z { get; }
+
+    /// <summary>The rotation component of the quaternion.</summary>
+    [DataMember]
+    public readonly T W { get; }
+
```
tannergooding, line 267: A number of libraries also expose `Vector3 Axis` (X, Y, Z) and `float Angle` (`2 * Acos(W)`) as "properties"

Many even parity most of the Vector4 surface area, including `Length`

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1189 @@
...
+    /// <altmember cref="Identity"/>
+    public bool IsIdentity { get; }
+
+    /// <summary>Adds each element in one quaternion with its corresponding element in a second quaternion.</summary>
+    /// <param name="value1">The first quaternion.</param>
+    /// <param name="value2">The second quaternion.</param>
+    /// <returns>The quaternion that contains the summed values of <paramref name="value1" /> and <paramref name="value2" />.</returns>
+    /// <remarks>The <see cref="op_Addition" /> method defines the operation of the addition operator for <see cref="Quaternion" /> objects.</remarks>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static Quaternion<T> operator +(Quaternion<T> value1, Quaternion<T> value2);
```
tannergooding, line 307: nit: Prefer `left` and `right` over `value1` and `value2` for these, since they are clearly the left and right sides of the `+` operator.

documentation/proposals/Proposal - Generic Math.md
tannergooding: Does this even need to exist given `operator +` exists?

documentation/proposals/Proposal - Generic Math.md
tannergooding: If you're tracking things like `origin` (the position) in addition to the `size`, then `AABB` simply drops the `rotation` field of the transform

GLM and DXM define `BoundingBox` as its own thing, since its often used in physics: https://learn.microsoft.com/en-us/windows/win32/api/DirectXCollision/ns-directxcollision-boundingbox

They also have variants (like `BoundingOrientedBox`, which is AABB) that drop fields for this efficiency purpose.

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1189 @@
...
+    /// </summary>
+    protected internal static abstract int Count { get; }
+}
+
+public interface IVectorEquatable<TVector, TScalar>
+    : IEquatable<TVector>
+    where TVector : IVector<TVector, TScalar>
+    where TScalar : INumberBase<TScalar>
+{
+    public TVector ScalarsEqual(TVector other);
```
tannergooding, line 573: Is this returning `AllBitsSet` for match and `Zero` for no match?

If so, do the other types need to expose this concept explicitly? Does there then need to be some `IndexOfFirstMatch`, `IndexOfLastMatch`, etc?

Should this just be called `Equals` (like `Vector<T>` and `Vector128<T>`) or maybe `CompareEqual`?

tannergooding: What also, about `IVectorComparable` for the other comparison operators?

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1189 @@
...
+
+    static virtual TVector Zero => TVector.Create(TScalar.Zero);
+
+    static virtual TVector One => TVector.Create(TScalar.One);
+
+    int IReadOnlyCollection<TScalar>.Count => TVector.Count;
+
+    TScalar IReadOnlyList<TScalar>.this[int index];
+    
+    public new TScalar this[int index];
```
tannergooding, line 608: why do you need this, shouldn't it coming from `IReadOnlyList` be an identical signature?

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1189 @@
...
+    /// <param name="left1">The value which multiplies <paramref name="right" />.</param>
+    /// <returns>The product of <paramref name="right" /> multiplied-by <paramref name="left1" />.</returns>
+    static virtual TVector operator *(TScalar left1, TVector right) => right * left1;
+
+    /// <summary>Multiplies two values together to compute their product.</summary>
+    /// <param name="right">The value which <paramref name="left" /> multiplies.</param>
+    /// <param name="left">The value which multiplies <paramref name="right" />.</param>
+    /// <returns>The product of <paramref name="right" /> multiplied-by <paramref name="left" />.</returns>
+    /// <exception cref="OverflowException">The product of <paramref name="right" /> multiplied-by <paramref name="left" /> is not representable by <typeparamref name="TVector" />.</exception>
+    static virtual TVector operator checked *(TScalar left, TVector right) => right * left;
```
tannergooding, line 660: nit: this should be `checked(right * left)`, the other checked operators need to likewise reuse `checked` so that overflow checking can happen as expected.

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1189 @@
...
+    protected static abstract bool TryConvertToTruncating<TOther, TOtherScalar>(TVector value, [MaybeNullWhen(false)] out TOther result)
+#nullable disable
+        where TOther : IVector<TOther, TOtherScalar> where TOtherScalar : INumberBase<TOtherScalar>;
+#nullable restore
+
+
+    #endregion
+}
+
+public interface IModulusVector<TVector, TScalar> :
```
tannergooding, line 830: Do you really "need" these rather than just having them on `INumberVector`?

documentation/proposals/Proposal - Generic Math.md
curin: The part that is an implementation detail is whether these are from extension methods or live in the primary struct. From the user's perspective assuming they both are in the same namespace, it will looks almost the same.

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1189 @@
...
+    /// <param name="value">The value whose set bits are to be counted.</param>
+    /// <returns>The number of set bits in <paramref name="value" />.</returns>
+    static abstract TVector PopCount(TVector value);
+
+    /// <summary>Tries to read a two's complement number from a span, in big-endian format, and convert it to an instance of the current type.</summary>
+    /// <param name="source">The span from which the two's complement number should be read.</param>
+    /// <param name="isUnsigned"><c>true</c> if <paramref name="source" /> represents an unsigned two's complement number; otherwise, <c>false</c> to indicate it represents a signed two's complement number.</param>
+    /// <param name="value">On return, contains the value read from <paramref name="source" /> or <c>default</c> if a value could not be read.</param>
+    /// <returns><c>true</c> if the value was succesfully read from <paramref name="source" />; otherwise, <c>false</c>.</returns>
+    static abstract bool TryReadBigEndian(ReadOnlySpan<byte> source, bool isUnsigned, [MaybeNullWhen(false)] out TVector value);
```
tannergooding, line 888: How will this work? For scalars, it reads the entire `source` as one scalar `T` and returns false if its too big.

For a vector, you can't know where one T stops and the next begins.

documentation/proposals/Proposal - Generic Math.md
Perksey: True, per the approved SDP we only guarantee source compatibility.

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1189 @@
...
+        where TNewVector : IVector<TNewVector, TInt>
+        where TInt : IBinaryInteger<TInt>;
+    static abstract TNewVector CeilingToInt<TNewVector, TInt>(TVector vector)
+        where TNewVector : IVector<TNewVector, TInt>
+        where TInt : IBinaryInteger<TInt>;
+
+    static virtual Vector64<TScalar> AsVector64(TVector self) => Vector64.Create(TVector.AsSpan(self));
+    static virtual Vector128<TScalar> AsVector128(TVector self) => Vector128.Create(TVector.AsSpan(self));
+    static virtual Vector256<TScalar> AsVector256(TVector self) => Vector256.Create(TVector.AsSpan(self));
+    static virtual Vector512<TScalar> AsVector512(TVector self) => Vector512.Create(TVector.AsSpan(self));
```
tannergooding, line 1114: nit: `As` is generally reserved for bitcasting where there can be no loss of data and its 1-to-1

`To` is generally used for converting, such as when it might not match up.

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1189 @@
...
+}
+
+public interface IVector2<TVector, TScalar> : IVector<TVector, TScalar>
+    where TVector : IVector2<TVector, TScalar>
+    where TScalar : INumberBase<TScalar>
+{
+    static virtual TVector UnitX => { get; }
+    static virtual TVector UnitY => { get; }
+    
+    TScalar X { get; }
```
tannergooding, line 1124: `X` could be on `IVector`, since it wouldn't make sense to define a 0'd dimension vector >.>

documentation/proposals/Proposal - Generic Math.md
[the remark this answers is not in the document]
tannergooding: Exposing fields publicly is "always" a mistake.

If you *really* need the ability to return a ref, expose a different ref returning property, an `AsSpan()` method, or something similar.

In the case of float/double, which are hardware accelerated, manipulating the individual fields is typically a mistake and you want to treat the entire vector as a single primitive.

----

The reason its always a mistake, is you can never change it in the future. Consider that `static readonly` may be less efficient than constructing a new value type each time, due to what the JIT can easily infer.

Consider that a `Matrix4x4` may be better represented as `4x Vector4` than `16x float` and you want to be able to change that in the future for efficiency reasons.

The properties give the minimal abstraction that lets you change things without hurting codegen, so you should always use them (outside very special cases like internal types where you can break the ABI; or interop types where you're matching an ABI stable contract with another language for lowlevel users).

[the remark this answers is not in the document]
Perksey: A ref indexer may be the way to go here.

documentation/proposals/Proposal - Generic Math.md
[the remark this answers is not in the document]
Perksey: If that's true, can we just call this primary ctor with all defaults and then do the efficient impl after and the JIT optimise it out? I just want to make sure we are actually wielding the benefits of a record struct if we're using it, otherwise this should just be a regular struct.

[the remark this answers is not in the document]
tannergooding: I think it should just be a regular struct.

A record struct's primary benefit is implicitly generating things like GetHashCode, Equals, ToString, etc

But Silk.NET has many reasons to override or even extend those default implementations with additional functionality, so I think the benefits of `record` are lost almost completely (if not entirely)

documentation/proposals/Proposal - Generic Math.md
curin: Its not for storage it is for Vector5 * Matrix5x4

documentation/proposals/Proposal - Generic Math.md
curin: Will fix this with the next revision, as I agree

documentation/proposals/Proposal - Generic Math.md
curin: Will remove 👍

documentation/proposals/Proposal - Generic Math.md
curin: Is this just setting every component to these values?

documentation/proposals/Proposal - Generic Math.md
curin: Agreed

documentation/proposals/Proposal - Generic Math.md
curin: ¯\_(ツ)_/¯ I have no clue, I just was taking from the list I was given,

documentation/proposals/Proposal - Generic Math.md
curin: 4x2, No clue.
3x4 may just be for transpose.

documentation/proposals/Proposal - Generic Math.md
curin: This is how Affine transformation matricies are applied to vectors (Vector3, 1) * Matrix 4x4. Whether that should stay as a transform method I guess would be the debate here.

documentation/proposals/Proposal - Generic Math.md
curin: 👍

documentation/proposals/Proposal - Generic Math.md
tannergooding: Yep!

Notably `AllBitsSet` is also a common one

documentation/proposals/Proposal - Generic Math.md
tannergooding: I would expect maybe a specialized `WriteTransposed` or similar in that case, since the primary scenario here is to store to GPU memory as part of a copy, you don't actually need the `2x3` at any point.

documentation/proposals/Proposal - Generic Math.md
curin: T is what INumberBase is being applied to not Quaternion. 
Also I am planning to restrict this to IBinaryFloatingPointIeee754 instead so that we match Vector4F, which makes a lot of assumptions easier, and integer Quaternions don't make much sense.

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,1189 @@
...
+
+    /// <summary>The Z value of the vector component of the quaternion.</summary>
+    [DataMember]
+    public readonly T Z { get; }
+
+    /// <summary>The rotation component of the quaternion.</summary>
+    [DataMember]
+    public readonly T W { get; }
+    
+    public ReadOnlySpan<T> Components { get; }
```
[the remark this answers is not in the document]
curin: Agreed

documentation/proposals/Proposal - Generic Math.md
curin: This makes sense, I'm updating to add this to the requirements

documentation/proposals/Proposal - Generic Math.md
curin: Adding

documentation/proposals/Proposal - Generic Math.md
Khitiara: update: opengl bindings use matrix types 2x2 up to 4x4 with all combinations thereof in parameters and returns so they are all needed from that perspective

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,366 @@
...
+
+The main types defined for this proposal are two sets of vector types, `VectorNI<T>` and `VectorNF<T>` where `N` defines the dimensionality of the vector between 2 and 4.
+
+For each vector struct, the following requirements **must** fulfill the following requirements:
+- Implements IEquatable with itself as the generic parameter
+- Implements IReadonlyList with the components as the list elements
+- The relevant number of properties to represent the mathematical vector's components (X and Y for Vector2) and relevant unit vectors
+- Constructors which take either a single parameter and uses it for every component, a parameter for each component, or a ReadOnlySpan of values which has the same number of elements as our vector has components.
+- Constructors for 3 dimensions and up **must** include lower dimension variants that use the lower dimensions for their specific components (vector2 -> X,Y).
+- A ref indexer that takes a int index and returns the corresponding component value (0 -> x, 1 -> y, etc.)
```
Khitiara, line 38: did we decide whether byref or settable byval indexer?

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,366 @@
...
+- An AsSpan function which returns this vector as a Span of the generic type
+- A LengthSquared property which returns the dot product of the vector with itself.
+- A Dot function which takes another vector and returns the dot product with our original vector. 
+  - A static implementation of this function **must** be available as well.
+- For 3D Vectors, a Cross function which takes another vector and returns the cross product with our original vector. 
+  - A static implementation of this function **must** be available as well.
+- `+`, `-`, `*`, `/`, and `%` operators defined between two vectors of the same type which returns a vector which has had each operation applied component-wise.
+- `+`, `-`, `*`, `/`, and `%` operators defined between a vector and a scalar value that matches the generic type which returns a vector which has had each operation applied component-wise with the scalar value. Both vector first and scalar first should be implemented.
```
Khitiara, line 46: scalar-vector only makes sense for `*`,`/`,`%` and im not sure scalar first makes sense for the latter two

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,366 @@
...
+- A `-` unary operator which returns the negated vector.
+- A `+` unary operator which returns the vector.
+- Overrides ToString to show component values.
+- Include additional ToString implementations which take a `string` format and a `string` format and `IFormatProvider`
+- TryFormat functions which output a formatted string of the vector to a `Span<char>` or `Span<byte>`, in UTF-16 and UTF-8, respectively.
+- Static Parse and TryParse functions which take a `string`, `Span<char>`, or `Span<byte>` and `IFormatProvider`.
```
Khitiara, line 52: notate the implementation of ISpanFormattable, ISpanParsable, and IUtf8... interfaces?

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,366 @@
...
+  - A Static implementation of this function **must** be available as well.
+- Max and Min functions, which takes a scalar value which matches the generic type and returns a new vector which component-wise has the Max or Min value with the scalar value, respectively.
+  - A Static implementation of this function **must** be available as well.
+- A Clamp function which takes a Max vector and a Min vector and returns a vector which has its components bounded between the Min and Max vectors.
+  - A Static implementation of this function **must** be available as well.
+- A Clamp function which takes a Max scalar and a Min scalar, both which match the generic type, and returns a vector which has its components bounded between the Min and Max scalars.
+  - A Static implementation of this function **must** be available as well.
+- An Abs function which returns a vector where each component is the absolute value of the original
+  - A Static implementation of this function **must** be available as well.
+- CopyTo functions which copy to an array, with or without a starting index
```
Khitiara, line 63: copyto span

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,366 @@
...
+- A Static Zero Vector with zero for all components
+- A Static One Vector with one for all components
+- A static AllBitsSet Vector with all bits set for all components
+- Define static CreateChecked, CreateSaturating, and CreateTruncating which converts other vector types to this type
+  - Try variants of these methods should also be defined which out the resulting vector and return a bool representing success or failure of the operation.
+- Define Transform functions which take a Matrix of higher dimensionality (Vector 2 can use Matrix2xn, Matrix3xn, and matrix4xn) and return a vector containing the output (type should match the outer type e.g. Vector2.Transform(Matrix4x4) returns Vector2)
+  - A Static implementation of these functions **must** be available
+  - Define matching `*` operators which function the same
+- Define TransformNormal functions which take a Matrix of higher dimensionality (Vector 2 can use Matrix2xn, Matrix3xn, and matrix4xn) and return a vector containing the output (type should match the outer type e.g. Vector2.Transform(Matrix4x4) returns Vector2)
+  - A Static implementation of these functions **must** be available
```
Khitiara, line 83: clarify behavior of these functions as assuming 1 for missing components

Khitiara, line 83: should this be permitted for more than one extra dimension?

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,366 @@
...
+
+For I types, the following requirements **must** also be fulfilled:
```
Khitiara, line 85: following additional requirements and remove the also maybe? would read better to me that way
```suggestion
For I types, the following additional requirements **must** be fulfilled:
```

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,366 @@
...
+- the bitwise `&`, `|`, and `^` operators defined between two vectors which returns a vector which has had these operators applied on a component-wise basis.
+- the bitwise `&`, `|`, and `^` operators defined between a vectors and a scalar value that matches the generic type which returns a vector which has had these operators applied on a component-wise basis with the scalar.
+- the unary bitwise `~` operator defined which negates the bits of the vector components.
+- AllBitsSet property which returns a Vector with all bits set to 1
```
Khitiara, line 89: duplicate with line 76 above

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,366 @@
...
+  - WriteLittleEndian(Span<byte> destination, int startIndex)
+    - Returns int representing the bytes written
+  - WriteLittleEndian(byte[] destination)
+    - Returns int representing the bytes written
+- Define the following static functions for these types to match IBinaryInteger (Vector replaced with type, e.g. `Vector2I<T>`) which returns a new vector with these operations applied component-wise, unless otherwise specified:
+  - Log2(Vector x)
+  - DivRem(Vector left, Vector right)
+    - Returns tuple of 2 Vectors (Vector Quotient, Vector Remainder)
+  - PopCount(Vector x)
+    - returns
```
Khitiara, line 116: returns what?

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,366 @@
...
+  - Atan2(Vector x, TScalar y)
+  - Atan2Pi(Vector x, TScalar y)
+  - BitDecrement(Vector x)
+  - BitIncrement(Vector x)
+  - FusedMultiplyAdd(Vector left, Vector right, Vector addend)
+  - FusedMultiplyAdd(Vector left, Vector right, TScalar addend)
+  - FusedMultiplyAdd(Vector left, TScalar right, Vector addend)
+  - FusedMultiplyAdd(Vector left, TScalar right, TScalar addend)
+  - ReciprocalEstimate(Vector x)
+  - ReciprocalSqrtEstimate(TVector x)
```
Khitiara, line 199: TVector
```suggestion
  - ReciprocalSqrtEstimate(Vector x)
```

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,366 @@
...
+  - WriteLittleEndian(byte[] destination)
+    - Returns int representing the bytes written
+- Define the following static functions for these types to match IBinaryInteger (Vector replaced with type, e.g. `Vector2I<T>`) which returns a new vector with these operations applied component-wise, unless otherwise specified:
+  - Log2(Vector x)
+  - DivRem(Vector left, Vector right)
+    - Returns tuple of 2 Vectors (Vector Quotient, Vector Remainder)
+  - PopCount(Vector x)
+    - returns 
+
+For F types, the following requirements **must** also be fulfilled:
```
Khitiara, line 118: same suggestion as with I types
```suggestion
For F types, the following additional requirements **must** be fulfilled:
```

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,366 @@
...
+
+# I types versus F Types
+Each type in this proposal, aside from `Quaternion` ends in I or F, defining whether it is an integer type or floating point type. Integer types **must** use a generic type argument `T` with the constraint of `IBinaryInteger<T>`. On the other hand, floating point types **must** use a generic type argument `T` with the constraint of `IFloatingPointIeee754<T>`.
```
Khitiara, line 26: should have comma after quaternion
```suggestion
Each type in this proposal, aside from `Quaternion`, ends in I or F, defining whether it is an integer type or floating point type. Integer types **must** use a generic type argument `T` with the constraint of `IBinaryInteger<T>`. On the other hand, floating point types **must** use a generic type argument `T` with the constraint of `IFloatingPointIeee754<T>`.
```

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,366 @@
...
+  - ScaleB(Vector x, int n)
+  - RoundToInt(Vector x)
+    - Returns `VectorNI`, where N matches the dimensionality of the vector
+    - **INFORMATIVE** This may require multiple methods depending on implementation
+  - FloorToInt(Vector x)
+    - Returns `VectorNI`, where N matches the dimensionality of the vector
+    - **INFORMATIVE** This may require multiple methods depending on implementation
+  - CeilingToInt(Vector x)
+    - Returns `VectorNI`, where N matches the dimensionality of the vector
+    - **INFORMATIVE** This may require multiple methods depending on implementation
```
Khitiara, line 213: suggest specifying that these should take a type parameter for the particular int scalar type to use

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,366 @@
...
+Matrix structs **must** fulfill the following requirements:
+- Fulfills `IEquatable<T>` where `T` is the same matrix class
+- Stored in row major format
+- F matricies work with F vectors, and I Matricies work with I vectors
+- Both row vectors and individual values (M11, etc.) accessible via properties
+- A ref indexer that takes row and column indicies and outputs the value
+- Add, subtract, and multiply operators defined with Matricies of the same size
+- Multiply operators defined with compatible matricies, if the output matrix type already exists (AxB * BxC = AxC)
+- Multiply operator defined with the appropriate Vector type (vector4F * Matrix4x4F is valid)
+  - **optionally**, include multiplication operators for Vector types smaller than the appropriate type, assuming 1 for the missing dimensions (vector2F * Matrix4x4F -> (Vector2F.x, Vector2F.y, 1, 1) * Matrix4x4F)
```
Khitiara, line 258: this is specified required in the vector types above. additionally, im unsure if stepping up multiple dimensions should be permissible

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,366 @@
...
+- Ray2F
+- Ray2I
+- Ray3F
+- Ray3I
+- Rect2F
+- Rect2I
+- SphereF
+- SphereI
+
+The Box structs are defined by Min and Max Vectors, while Rectangle and Cube are defined by Origin and Size Vectors. These Types **must** be implicitly castable between each other.
```
Khitiara, line 350: update for RectN names for rectangle and cube
```suggestion
The Box structs are defined by Min and Max Vectors, while the Rect structs are defined by Origin and Size Vectors. These Types **must** be implicitly castable between each other.
```

Khitiara (changes requested):

documentation/proposals/Proposal - Generic Math.md
Khitiara: as above the additional types are needed for opengl bindings

documentation/proposals/Proposal - Generic Math.md
tannergooding: Can you give an example of such an API?

AFAIK, OpenGL itself doesn't have a built-in matrix type on its native side (it defers everything to GLM, which is separate). GLSL supports a little bit of arbitrary matrix sizes, but noting its `mat#x#` is `CxR`, which is backwards from standard convention. SPIR-V and HLSL then support them as well, but I don't think I've ever seen them used in practice.

Notably, even for float/double, they would basically be "anti-patterns" and users would easily fall into a perf pit of failure by using them, hence why I think it'd be "better" to have the shape that does work well, and to encourage exposing `LoadTransposed` and `StoreTransposed` to allow users to efficiently handling "transpose + copy to/from memory".

documentation/proposals/Proposal - Generic Math.md
curin: There is a reason to have one that does and one that does not, which matches Maxine's implementation currently (int can be done implicitly, the others need casts I believe). This is an mainly an implementation detail, that I want to leave up to whoever is implementing it.

documentation/proposals/Proposal - Generic Math.md
curin: Good catch, changing to just implementing those interfaces now

documentation/proposals/Proposal - Generic Math.md
curin: It is component-wise so (4,5) + 5 = (9,10)
This isn't necessary, but more as an easy of use. If generally agreed we shouldn't have these I'll remove

documentation/proposals/Proposal - Generic Math.md
curin: We decided by ref

documentation/proposals/Proposal - Generic Math.md
curin: fixed

documentation/proposals/Proposal - Generic Math.md
curin: removed from here, but I did agree that this functionality should match Transform as Petre mentioned in the discord. I could go either way, but it was the decision I made at the time.

documentation/proposals/Proposal - Generic Math.md
curin: fixed

documentation/proposals/Proposal - Generic Math.md
curin: fixed in both places

documentation/proposals/Proposal - Generic Math.md
curin: This should definitely be discussed, but for now I don't see a reason why not.

documentation/proposals/Proposal - Generic Math.md
curin: fixed

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,366 @@
...
+
+# **INFORMATIVE** Integer and Floating Point Types
+While investigating the use of generic math we came to the conclusion that making types which supports both integer and floating point types would not be optimal. This was discussed at length on the discord [here](https://discord.com/channels/521092042781229087/587346162802229298/1167705816812498974). Ultimately it was decided to provide both an integer and floating point variant for each vector type and every type built from them. These types are generic where `Vector2I<T>` will be a 2D vector which takes any binary integer type for `T`. Similarly `Vector2F<T>` will be a 2D vector which takes any floating point type for `T`. By extension we get types like `CubeI<T>` and `RectangleF<T>`. The integer types are granted the bitwise operators `&`, `~`, `|`, and `^`. Floating point types will include some operations that require certain functions unavailable to integer types like `Length` which requires `Sqrt`.
```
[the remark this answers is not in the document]
curin: fixed

documentation/proposals/Proposal - Generic Math.md
curin: Fixed

documentation/proposals/Proposal - Generic Math.md
curin: whoops I must have got distracted writting this, fixed

documentation/proposals/Proposal - Generic Math.md
curin: fixed

documentation/proposals/Proposal - Generic Math.md
curin: fixed, also Transform normal assumes 0, as normals should not be translated.

documentation/proposals/Proposal - Generic Math.md
```diff
@@ -0,0 +1,366 @@
...
+
+Each type **must** include the following:
+- Intersect functions with both another instance of the type and a point
+- GetDistanceToNearest(Point,Edge,etc) functions **must** be available for a given point
+- For all but the rays and planes, GetInflated function that takes a point and returns the scaled object that is closest to the original and contains the given point
+- Include Scale and Translation transformation functions
+- For Planes and Rays, Normalize functions
+- For Planes include the following static functions
+  - CreateFromVerticies
+  - CreateFromVertexAndNormal
```
Khitiara, line 360: more standard name would be `CreateFromPointNormal` as that is what the form is usually called
```suggestion
  - CreateFromPointNormal
```

Khitiara (changes requested):

documentation/proposals/Proposal - Generic Math.md
tannergooding: My brain apparently wasn't working. I swore I read this as `Quaternion<T> : INumberBase<T>` 😆

Perksey:
I'll merge this for simplicity in the meeting but make sure someone writes notes on the things we need to discuss a la the outstanding comments.

Issue #1823, opened 2023-12-04
OpenAL ReopenDevices extension doesn't work


okaniku:
# Summary
The OpenAL ReopenDevices extension doesn't seem to work.

# Steps to reproduce
After setting a current context, and calling
`reopenDevices.ReopenDevice(device, deviceName, attrList);`
I get 'Native symbol not found (Symbol: alcReopenDeviceSOFT)'

# Comments
I'm wondering whether it's because this is an ALC prefixed extension, and the extension class derives from `NativeExtension<AL>` rather than `ContextExtensionBase`.

Ref: https://github.com/dotnet/Silk.NET/blob/main/src/OpenAL/Extensions/Silk.NET.OpenAL.Extensions.Soft/ReopenDevices.cs

Interestingly, getting the extension in the first place works fine from the AL API:
```
al = AL.GetApi(true);
al.TryGetExtension(out ReopenDevices reopenDevices);
```

Is this a bug or something I'm doing wrong?

okaniku:
Unfortunately this is still happening, even after changing it to a context extension.

Strangely, I'm able to successfully call the underlying function by doing this:

```
using var alc = ALContext.GetApi(true);
var deviceName = string.Empty;
var dev = alc.OpenDevice(deviceName);
var ctx = alc.CreateContext(dev, null);
alc.MakeContextCurrent(ctx);

if (dev != null && alc.TryGetExtension(dev, out ReopenDevices ext))
{
    var maxName = SilkMarshal.GetMaxSizeOf(deviceName, NativeStringEncoding.UTF8);
    var nameNative = deviceName.Length > 256 ? new byte[maxName] : stackalloc byte[maxName];
    fixed (byte* namePtr = nameNative)
    {
        ((delegate* unmanaged[Cdecl]<Device*, byte*, int*, bool>) alc.GetProcAddress(dev, "alcReopenDeviceSOFT"))(dev, namePtr, null);
    }
}
```

Perksey:
@okaniku Can you compare the results of `alc.Context.GetProcAddress("alcReopenDeviceSOFT")` vs `alc.GetProcAddress("alcReopenDeviceSOFT")`

Relevant code:

https://github.com/dotnet/Silk.NET/blob/main/src/OpenAL/Silk.NET.OpenAL/ALC/ALContext.cs#L98

okaniku:
[...]

PR #1824, opened 2023-12-05
Fix AL ReopenDevices extension

okaniku:
# Summary of the PR
Fixes the AL ReopenDevices extension to derive from `ContextExtensionBase` since it is a context extension.

[Issue](https://github.com/dotnet/Silk.NET/issues/1823)


Perksey (approved):
Thanks

PR #1868, opened 2023-12-28
[3.0] Cache ClangSharp outputs & use cache if we shouldn't run ClangSharp

Perksey:
This introduces a caching mechanism in SilkTouch 3.0 so that we can cache the raw outputs from ClangSharp to allow the rest of the SilkTouch pipeline to be run & developed on platforms other than the ones in which ClangSharp can run. For example, we'd cache raw outputs for the Windows SDK bindings so that if we don't have access to the Windows SDK for example (e.g. we're on macOS), we'd open the cache and feed the cached outputs through the pipeline.

The way I envision this looking is:
- The bindings that can't be regenerated outside of a specific environment are cached upon regeneration in that environment.
- We'd have a CI job that will run in that specific environment to regenerate the cache when there's an update to the underlying dependency, similar to how dependabot works.
- This keeps all developers regardless of platform to be able to clone Silk.NET and regenerate all the bindings, cached or uncached as appropriate.

This was also how the BuildTools cache was intended to work, but I don't know of a single instance of it working properly and even if it did, the cache was output at a pretty late stage in the pipeline such that the only changes observable without regenerating the cache were pretty much just overloads.


Perksey:
Oh and this also lets us use Git repos & NuGet packages as ClangSharp inputs, alleviating the need for submodules which have to be manually synchronised.

Perksey:
I think this is ready for review, I'll do the NUKE/CI stuff in a later PR.

The LOC count may seem excessive, but this is because TerraFX has updated and as a result so has our Windows bindings.

The bit I'm interested in scrutiny for is the stuff in sources/SilkTouch as this has been written over many sleep-deprived evenings.

sources/SilkTouch/Caching/FileSystemCacheProvider.cs
```diff
@@ -0,0 +1,243 @@
...
+
+    internal static string CommonDir { get; set; } = Environment.CurrentDirectory;
+
+    /// <inheritdoc />
+    public async Task<(string Path, bool IsNew)?> GetDirectory(
+        string cacheKey,
+        CacheIntent intent,
+        CacheFlags flags
+    ) =>
+        await CoreGetDirectory(cacheKey, intent, intent == CacheIntent.ResolvedForeignInput, flags);
```
Khitiara, line 29: why not return the task directly instead of using await here?

sources/SilkTouch/Caching/FileSystemCacheProvider.cs
```diff
@@ -0,0 +1,243 @@
...
+                                {
+                                    options.UnixCreateMode = unixFileMode;
+                                }
+
+                                var dstPath = Path.Combine(path, entry.FullName);
+                                Directory.CreateDirectory(Path.GetDirectoryName(dstPath) ?? path);
+                                await using var destination = new FileStream(dstPath, options);
+                                await using var stream = entry.Open();
+                                await stream.CopyToAsync(destination, ct);
+                                await destination.FlushAsync(ct);
```
Khitiara, line 119: does disposeasync not flush?

sources/SilkTouch/Caching/FileSystemCacheProvider.cs
curin: Awaiting a async method is different than just returning the output. This is also standard practice for calling async methods as far as I'm aware.

sources/SilkTouch/Caching/FileSystemCacheProvider.cs
curin: From what I can find, CopyToAsync does not Flush automatically, so this does seem correct

curin (approved):
Looks good from what I can tell. Didn't really look at the d3d12 code, since I'm assuming that is just raw clangsharp output atm. The actual code for Caching, Sources, and the new AddIncludes looks all pretty sound to me. 
The generated OpenGL code also looks a bit better formed.

sources/SilkTouch/Caching/FileSystemCacheProvider.cs
Khitiara: its being await using'd though, my question is about DisposeAsync not the copyto

sources/SilkTouch/Caching/FileSystemCacheProvider.cs
Khitiara: everything ive seen in bcl is if youre going to have the inner core method you just return directly to save the overhead of a second state machine, usually in the context of an outer method that does arg checking before forwarding to keep the arg checking out of the async though

Perksey: Awaiting is required to link the stacks correctly as I understand it, which impacts exception handling and will probably has some async2 implications in the future as well. If this were best practice, there’d probably be an optimisation in Roslyn and/or a ReSharper codefix to do essentially what you describe, which I don’t believe is the case(?)

sources/SilkTouch/Caching/FileSystemCacheProvider.cs
curin: I don't think dispose will necessarily flush either, but I'm unsure

Perksey: In my experience it’s not necessarily the case, and indeed before I added flushing for the _noHostZips I was getting corrupted archives, so I think it is technically required.

sources/SilkTouch/Caching/FileSystemCacheProvider.cs
Khitiara: rider suggests it for me but that might not be standard idk

sources/SilkTouch/Caching/FileSystemCacheProvider.cs
Khitiara: alright, just wanted to check

Perksey:
Thank you all for reviewing and your continued support with the 3.0 effort!

PR #1882, opened 2024-01-07
Use ref readonly instead of in

Perksey:
Manually defines the RequiresLocationAttribute given that we can't upgrade away from the .NET 7 SDK and thereby get C# 12.


Perksey:
Ready for review.

src/Assimp/Silk.NET.Assimp/AssimpOverloads.gen.cs
```diff
@@ -3122,7 +3122,7 @@ public static unsafe Return GetMaterialProperty(this Assimp thisApi, [Flow(Silk.
 
         /// <summary>To be documented.</summary>
         [NativeName("Src", "Line 1477, Column 28 in material.h")]
-        public static unsafe Return GetMaterialProperty(this Assimp thisApi, [Flow(Silk.NET.Core.Native.FlowDirection.In)] Material* pMat, [Flow(Silk.NET.Core.Native.FlowDirection.In)] ReadOnlySpan<byte> pKey, uint type, uint index, [Flow(Silk.NET.Core.Native.FlowDirection.In)] in MaterialProperty* pPropOut)
+        public static unsafe Return GetMaterialProperty(this Assimp thisApi, [Flow(Silk.NET.Core.Native.FlowDirection.In)] Material* pMat, [Flow(Silk.NET.Core.Native.FlowDirection.In)] ReadOnlySpan<byte> pKey, uint type, uint index, [Flow(Silk.NET.Core.Native.FlowDirection.In)] [RequiresLocation] in MaterialProperty* pPropOut)
```
tannergooding, line 3125: The C# encoding of `ref readonly` is `[RequiresLocation] ref`, not `[RequiresLocation] in`

Perksey: SharpLab seems to have the `[in]` IL modifier set?
```
    .method public hidebysig 
        instance void M (
            [in] int32& x
        ) cil managed 
    {
        .param [1]
            .custom instance void [System.Runtime]System.Runtime.CompilerServices.RequiresLocationAttribute::.ctor() = (
                01 00 00 00
            )
```

Khitiara: the draft feature spec in the official docs additionally states:
> `ref readonly` parameters will be emitted as `[in] T&`, plus annotated with the following attribute:
> ```cs
> namespace System.Runtime.CompilerServices
> {
>    [AttributeUsage(AttributeTargets.Parameter, AllowMultiple = false, Inherited = false)]
>    public sealed class RequiresLocationAttribute : Attribute
>    {
>    }
> }
> ```
though that is the draft spec so im unsure if it actually matches the impl

tannergooding: `[in] T` is not the same as `in T`, `[in]` is an IL modifier used as part of interop, it equates to `System.Runtime.InteropServices.InAttribute`

* `ref T` is encoded in IL as `T&`
  * The raw C# equivalent: `ref T`
* `out T` is encoded in IL as `[out] T&`
  * The raw C# equivalent: `[System.Runtime.InteropServices.OutAttribute] ref T`
* `in T` is encoded as `[System.Runtime.CompilerServices.IsReadOnlyAttribute] T&`
  * The raw C# equivalent: `[System.Runtime.CompilerServices.IsReadOnlyAttribute] ref T`
  * The compiler currently also emits `[in]` (that is `[System.Runtime.InteropServices.InAttribute]`) to ensure interop behaves correctly
  * The compiler did emit `[in]` in the first official release of the feature and it is not part of the required recognition
* `ref readonly T` is encoded as `[System.Runtime.CompilerServices.RequiresLocationAttribute] T&`
  * The raw C# equivalent: `[System.Runtime.CompilerServices.RequiresLocationAttribute] ref T`
  * The compiler currently also emits `[in]` (that is `[System.Runtime.InteropServices.InAttribute]`) to ensure interop behaves correctly
  * The compiler has always emitted `[in]` for the feature, but it is not part of the required recognition
  * It is unclear what the behavior the compiler will have for `[RequiresLocation, IsReadOnly] ref T`, but afair there was an explicit decision to not do it this way

Khitiara: ah that makes more sense then; how does binary back compat work because i think `ref readonly` is then technically a breaking change for us if we dont maintain the `[IsReadOnly]`? and also does the `RequiresLocationAttribute` have to be public or does internal work ok for this?

Perksey: It will break old compilers tmk, I really want the warning that comes with ref readonly but it seems like there’s loads of tradeoffs.

cc @333fred (please feel free to ignore, I know you’re busy and this is unsolicited) I don’t suppose you know what the considerations were of ref readonly on older compilers? Is it intended to default to ref behaviour instead of in behaviour? Is there any implications of not doing this i.e. is it okay to use both attributes?

333fred: I have no idea. @jjonescz would have a better idea.

tannergooding: > how does binary back compat work because i think ref readonly is then technically a breaking change for us if we dont maintain the [IsReadOnly]

Attributes, outside of `modreq` and `modopt`, are not part of the signature and so from a binary perspective, `ref T`, `out T`, `in T`, and `ref readonly T` are all simply `T&` and therefore "compatible".

It's only the source compatibility that comes into question and older compilers would indeed see it as `ref T` now rather than `ref readonly T` or `in T`, they would have to switch to using `ref` explicitly. The exact semantics of the parameter type and callsite annotation are here: https://github.com/dotnet/csharplang/blob/main/proposals/csharp-12.0/ref-readonly-parameters.md#summary

Perksey: I suggest we go ahead with this for today's update. Source compatibility is what matters most to us, and for old compilers using `in` makes sense. It is true that it is unknowable what adding this attribute to `in` instead of `ref` means, but worst case scenario is we just have the current behaviour - I doubt this will actually brick the compuler.

Perksey: Okay so I've looked in the spec as it says: 

> Note that unlike `in` parameters, no `[IsReadOnly]` will be emitted for `ref readonly` parameters to avoid increasing metadata size and also to make older compiler versions interpret `ref readonly` parameters as `ref` parameters (and hence `ref` → `ref readonly` won't be a source breaking change even between different compiler versions).

Now obviously this is UB, but we would want older compilers to interpret `ref readonly` as `in` instead of `ref` primarily because this is what user code uses today. If modern compilers interpret `[RequiresLocation, IsReadOnly]` as `in`, this is fine as it just means this is a failed experiment. If it interprets it as `ref readonly`, great as well. The only bad case will be if it gets confused and instead of making a decision, interprets it as `ref`. We'll use the experimental feed to derisk this.

tannergooding: Its worth noting the the BCL didn't do anything fancy like this. We took the "break" such that older compilers (which are really unsupported to use with newer language features anyways) would see it as `ref`

The language also considered the implications around this when designing the feature in the first place, so if Silk does something different here, it will probably be the only thing doing that.

Perksey: Is there an assumption that developers using older compilers are also using older .NETs? Or rather, aren’t using the newest BCL? I don’t think we’d like to take the break, but also want to avoid writing our own analyser for something the compiler in theory has built in. We support all the way back to .NET Framework 4.6.2 so we probably shouldn’t affect source compatibility, especially because this is the only compatibility we guarantee on paper.

tannergooding: >  that developers using older compilers are also using older .NETs? 

Yes. Each version of .NET comes with a matched version of Roslyn, so you can't (without extreme effort) reference a .NET 8 binary with say the C# 11 compiler.

Perksey:
Confirmed working by testing against `vk.CreateInstance`
```cs
unsafe
{
    var vk = Vk.GetApi();
    vk.CreateInstance(new InstanceCreateInfo(), null, null);
}
```
```
0>------- Started building project: Scratch
0>Program.cs(8,23): Warning CS9193 : Argument 1 should be a variable because it is passed to a 'ref readonly' parameter
Scratch -> /Users/dylan/Documents/RiderProjects/Scratch/Scratch/bin/Debug/net8.0/Scratch.dll
0>------- Finished building project: Scratch. Succeeded: True. Errors: 0. Warnings: 1
```

Perksey:
@Beyley Ready for review. I'm intending to update the non-Clang bindings at some point, did you want me to do that here or in a separate PR?

Beyley:
i think im too stupid to understand the reason for this, but the code itself looks good
after this is merged in i'll do a full update/regen of all the bindings including clang stuff

Perksey:
Force merging per Beyley's comment

PR #1907, opened 2024-01-21
[3.0] Fix OpenGL type-mappings

Perksey:
So there's a bunch of weird things that my recent GL PR pulled up with the headers, so this adds the type mappings as per https://www.khronos.org/opengl/wiki/OpenGL_Type.

This should hopefully ensure that outputs are the same across platforms.


Perksey:
Ready for review, raw diff is here: https://patch-diff.githubusercontent.com/raw/dotnet/Silk.NET/pull/1907.diff

curin (approved):

Issue #1550, opened 2023-07-02
SDL multi-window event-based input is inconsistent


desiwko:
# Summary
When utilizing event-based input via SDL in a multi-window environment events appear to get "lost".  MouseDown/MouseUp and KeyDown/KeyUp events do not consistently fire.  Presumably other input events are equally affected.  This is especially visible when trying use the "Close" widget on the window where clicking multiple times may never get processed.

# Steps to reproduce
- Platform: Desktop (Windows)
- Framework Version: .NET Core 7
- API: SDL / OpenGL
- API Version: OpenGL 4.5 Core

1. Instantiate multiple Silk windows using the SDL backend
2. Attempt to read mouse clicks in various windows
3. Observe inconsistent input event messages

# Comments
This is specific to the SDL backend only.  GLFW does not appear to exhibit this issue.

Minimum reproduceable example (This should look somewhat familiar):

```
using Silk.NET.Windowing;
using Silk.NET.Windowing.Extensions;
using Silk.NET.Input;
using Silk.NET.OpenGL;

namespace ogl;

public class OpenGL
{
    private WindowManager windowManager;
    private List<IInputContext> inputContexts = new List<IInputContext>();

    public OpenGL()
    {
        windowManager = new WindowManager();

        Window.PrioritizeSdl();

        for (int i = 0; i < 4; i++)
        {
            windowManager.CreateWindow(WindowOptions.Default);
        }
    }


    public void Start()
    {
        foreach(var window in windowManager.Windows)
        {
            window.Load += () =>
            {
                var inputContext = window.CreateInput();
                inputContexts.Add(inputContext);

                inputContext.Mice[0].MouseDown += OnMouseDown;
                inputContext.Mice[0].MouseUp += OnMouseUp;

                inputContext.Keyboards[0].KeyDown += OnKeyDown;
                inputContext.Keyboards[0].KeyUp += OnKeyUp;
            };

            window.Render += (time) =>
            {
                var context = GL.GetApi(window.GLContext);

                context.Enable(EnableCap.DepthTest);
                context.ClearColor(0f, .2f, .2f, 1f);
                context.Clear(ClearBufferMask.DepthBufferBit | ClearBufferMask.ColorBufferBit);
            };
        }

        windowManager.Run();
    }

    private void OnKeyUp(IKeyboard keyboard, Key key, int arg3)
    {
        System.Diagnostics.Debug.WriteLine("Key Up : " + key);
    }

[...]

PR #1909, opened 2024-01-21
Fix SDL2 platform losing events when using multiple windows/views

ZingBallyhoo:
# Summary of the PR
Currently the SDL windowing platform loses events when using multiple windows.

Flow of the bug:
- 1st Window DoEvents
  - Previous events cleared
  - Platform DoEvents
    - SDL PollEvent
    - Polled events dispatched to every window
  - Events handled for this window
- 2nd Window DoEvents
  - Previous events cleared **(and everything we just collected)**
  - Platform DoEvents
    - SDL PollEvent **(likely gives nothing since we just polled already)**
    - Polled events dispatched to every window
  - Events handled for this window **(there are none)**

The solution in this pr is to add a new (internal) AfterProcessingEvents method to (the internal) ViewImplementationBase, which performs the clearing of the event queue. Clearing at the end of DoEvents is too early for the input module.

I also took the liberty of consolidating event removal logic. I thought its best to be safe around free handling for SDL_DropEvent.

Fixes https://github.com/dotnet/Silk.NET/issues/1550
Fixes https://github.com/dotnet/Silk.NET/issues/350

# Further Comments
Feedback/reshaping welcome.
Maybe AfterProcessingEvents -> ClearEvents? and remove non-virtual ClearEvents from sdl view.
Or maybe this multi-poll logic is too cursed already? idk

### DIFF
--- a/src/Input/Silk.NET.Input.Sdl/SdlInputContext.cs
+++ b/src/Input/Silk.NET.Input.Sdl/SdlInputContext.cs
@@ -254,7 +254,7 @@ namespace Silk.NET.Input.Sdl
 
                 if (!skipped)
                 {
-                    _sdlView.Events.RemoveAt(i);
+                    _sdlView.RemoveEvent(i);
                 }
                 else
                 {
--- a/src/Windowing/Silk.NET.Windowing.Common/Internals/ViewImplementationBase.cs
+++ b/src/Windowing/Silk.NET.Windowing.Common/Internals/ViewImplementationBase.cs
@@ -289,7 +289,13 @@ namespace Silk.NET.Windowing.Internals
             _inRenderLoop = true;
             DoEvents();
             ProcessEvents?.Invoke();
+            AfterProcessingEvents();
             _inRenderLoop = false;
+        }
+        
+        internal virtual void AfterProcessingEvents()
+        {
+            // used by sdl2 windows to clear events
         }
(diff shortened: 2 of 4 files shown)


ZingBallyhoo:
@dotnet-policy-service agree

Perksey:
Wow! Thank you so much, haven't even reviewed it yet but I'm excited to! Before I do, do you think this could knock off #350 as well?

ZingBallyhoo:
I tried the zip attached in that issue seeing the same behavior both for SDL & GLFW now.

The "When resizing a window events still get mixed up." part is definitely an API usage issue, but I'm not a GL expert. There's nowhere in the code that specifies which window/gl context to use. If you swap the order of the IWindow.Initalize calls the other window gets drawn to and the centering makes sense.

Adding some names and comments makes it more clear what's happening.. try this if you want
[ReproIssue350.zip](https://github.com/dotnet/Silk.NET/files/14003199/ReproIssue350.zip)

Perksey (approved):
No issues here, thank you so much!

azeno:
You're right, my example project was missing a `MakeCurrent` call to ensure the Skia surface gets bound to the content window and not the blank window. Thanks for pointing that out. Did some more tests and can confirm that the events get routed correctly now. Thanks so much!

PR #1915, opened 2024-01-24
[3.0] Refactor the DSL mod to allow new DSL types to be plugged in

Perksey:
This is one way to approach the adding of things like GLEnum and GL handles. It may not be the final one, keeping in draft until I've proved it out a bit more.

In any case, this final form of this PR will include:
- DSL mods are now applied to the interfaces where applicable, instead of the classes derived from those interfaces
- The original unsafe overloads have been added back per the Working Group's request.


Perksey:
@curin (and any other interested parties for 3.0) This is now ready for review.

Please see the PR description. This will form the basis for API-specific overloading and transformation.

As per @aquagoose's suggestion which was approved by the Working Group, this also resurrects the raw overloads and defines the DSL overloads in terms of those functions.

Next up:
- Per-stage hard and soft mod dependencies, so some mods can execute their functionality before or after a given stage (topologically sorted). This is necessary to ensure that the KHR transformers which will convert `GL.Constants` into `GLEnum` and to sort out the enum groups. The functionality added in this PR wrt DSL transformers will be used to actually modify the functions.
- Sort out the SupportedApiProfile logic, possible with the above.
- OpenGL bindings done?
- Allow scrapers to be pluggable/mods in their own right i.e. ClangScraper will be a mod. This isn't related to the OpenGL line of work, but it's basically so we can ingest the TerraFX.Interop.Windows checked-in bindings as-is rather than regenerating them when we don't need to (a good generation will produce identical outputs, a bad one will fail for other reasons).
- Start work on Windows SDK bindings?

curin (approved):
Looks good, in the future it would be nice if global formatting changes were a separate review. There are a lot of whitespace changes in files irrelevant to the changes this PR addresses. It is a little difficult to always tell what changes were relevant at a glance.

Perksey:
Will do, thanks!

PR #1952, opened 2024-02-19
Added SetFocus method to the Window

Arugin:
# Summary of the PR
Added SetFocus method to the Window

As was discussed here https://github.com/dotnet/Silk.NET/discussions/1526, both GLFW and SDL support this functionality and it is very usefull for the ImGui support, so it can be added to the window.

# Further Comments
Unfortunately I wasn't be able to build the project, to check if everything is working correctly, could someone to pick it up? (I am asked for some reason to enter credentials for https://chrome-internal.googlesource.com that I don't have while recursive submodule update. Also I have tons of  errors on nuke compile like  Restore: \Silk.NET\src\Windowing\Silk.NET.Windowing.Sdl\S
ilk.NET.Windowing.Sdl.csproj : error NU1301: Failed to load service index for source https://gitlab.com/api/v4/projects/47661606/packages/nuget/index.js
on. Silk.NET\Silk.NET.gen.sln)

### DIFF
--- a/src/Windowing/Silk.NET.Windowing.Common/Interfaces/IView.cs
+++ b/src/Windowing/Silk.NET.Windowing.Common/Interfaces/IView.cs
@@ -104,6 +104,11 @@ namespace Silk.NET.Windowing
         void Reset();
 
         /// <summary>
+        /// Sets focus to current window.
+        /// </summary>
+        void SetFocus();
+
+        /// <summary>
         /// Close this window.
         /// </summary>
         void Close();
--- a/src/Windowing/Silk.NET.Windowing.Common/Interfaces/IView.cs
+++ b/src/Windowing/Silk.NET.Windowing.Common/Interfaces/IView.cs
@@ -106,7 +106,7 @@ namespace Silk.NET.Windowing
         /// <summary>
         /// Sets focus to current window.
         /// </summary>
-        void SetFocus();
+        void Focus();
 
         /// <summary>
         /// Close this window.
--- a/src/Windowing/Silk.NET.Windowing.Common/Internals/ViewImplementationBase.cs
+++ b/src/Windowing/Silk.NET.Windowing.Common/Internals/ViewImplementationBase.cs
@@ -71,6 +71,7 @@ namespace Silk.NET.Windowing.Internals
         public abstract void ContinueEvents();
         public abstract Vector2D<int> PointToClient(Vector2D<int> point);
         public abstract Vector2D<int> PointToScreen(Vector2D<int> point);
+        public abstract void SetFocus();
         public abstract void Close();
         protected abstract void RegisterCallbacks();
         protected abstract void UnregisterCallbacks();
--- a/src/Windowing/Silk.NET.Windowing.Common/Internals/ViewImplementationBase.cs
+++ b/src/Windowing/Silk.NET.Windowing.Common/Internals/ViewImplementationBase.cs
@@ -71,7 +71,7 @@ namespace Silk.NET.Windowing.Internals
         public abstract void ContinueEvents();
         public abstract Vector2D<int> PointToClient(Vector2D<int> point);
         public abstract Vector2D<int> PointToScreen(Vector2D<int> point);
-        public abstract void SetFocus();
+        public abstract void Focus();
         public abstract void Close();
         protected abstract void RegisterCallbacks();
         protected abstract void UnregisterCallbacks();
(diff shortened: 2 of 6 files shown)


Perksey:
This has been added to the API Review queue.

Perksey:
Can you add the API to the PublicAPI files please?
```
  [ERR] Pack: /Users/runner/work/Silk.NET/Silk.NET/src/Windowing/Silk.NET.Windowing.Common/Interfaces/IView.cs(109,14): error RS0016: Symbol 'SetFocus' is not part of the declared public API [/Users/runner/work/Silk.NET/Silk.NET/src/Windowing/Silk.NET.Windowing.Common/Silk.NET.Windowing.Common.csproj::TargetFramework=netstandard2.0]
  [ERR] Pack: /Users/runner/work/Silk.NET/Silk.NET/src/Windowing/Silk.NET.Windowing.Common/Interfaces/IView.cs(109,14): error RS0016: Symbol 'SetFocus' is not part of the declared public API [/Users/runner/work/Silk.NET/Silk.NET/src/Windowing/Silk.NET.Windowing.Common/Silk.NET.Windowing.Common.csproj::TargetFramework=netstandard2.1]
```

Arugin:
Sure.

Arugin:
Any ETA when it can be merged? Or I need to do something else so it would be ready for merge?

Perksey:
We were waiting for another API review session, but given this is just one API it probably isn't worth the wait.

src/Windowing/Silk.NET.Windowing.Common/Interfaces/IView.cs
```diff
@@ -103,6 +103,11 @@ public interface IView : IViewProperties, IGLContextSource, IVkSurfaceSource, ID
         /// </summary>
         void Reset();
 
+        /// <summary>
+        /// Sets focus to current window.
+        /// </summary>
+        void SetFocus();
```
Perksey, line 109: I personally would like this renamed to BringToFront

curin: Personally SetFocus is slightly different in my mind than BringToFront. SetFocus means it is the currently active window, BringToFront would just make it the top window, not necessarily the active one.

Perksey: Note that while GLFW calls this glfwFocusWindow, SDL calls this SDL_RaiseWindow which is likely a more descriptive of what it does.

Personally, I do not want a method prefixed with `Set` with no parameters. If we're keeping the focus-focused name, then let's just call it `Focus`.

curin: Focus seems fine to me

Perksey: Seems agreeable enough.

ThomasMiz: `Focus` sounds good.

`Raise` sounds like it would could be ambiguous

HurricanKai:
I haven't tested this, but according to docs SDLs RaiseWindow both requests the input focus (so from my understanding typing would go directly into the window, without any further interaction) *and also brings it to the front*
~~GLFW does not document any changing of the window "layer". 
SDLs SetWindowInputFocus seems to be more in line with GLFWs (documented) functionality.~~
EDIT: I learned that although docs are confusing GLFW does in fact also couple both in FocusWindow

~~It could be considered to rename this to InputFocus and limit this to only that.~~
EDIT: Given new info I'm fine with this PR as-is.
The below is still useful information, and I wonder how NoBringToFrontOnFocus works on GLFW, but whatever. ImGUI pop-ups are cool, but maybe not the most important thing.

Looking at the original issue this seems to be about ImGUI multi-window.
I've never hooked this up myself, so not sure what the hooks look like, but ImGUI does also make this distinction quite explicitly. There's a difference between being on top and having input focus.
ImGUI controls this using ImGuiWindowFlags_NoBringToFrontOnFocus
This is explicitly useful when implementing pop ups, especially things like Auto-complete where it's important input focus is retained.

My suggestion would be to explicitly split this into BringToFront (naming?) and [Request]InputFocus
Although it's unclear how to implement this in GLFW.
EDIT: As said above, I'm also fine with this PR as-is. The suggestion stands, although it may be impossible.

src/Windowing/Silk.NET.Windowing.Common/Interfaces/IView.cs
Arugin: I've renamed the method to `Focus`.

Arugin:
I've renamed the method according to comments and rebased the branch.

ThomasMiz (approved):

ThomasMiz:
👌👌

PR #1953, opened 2024-02-20
extra features to the input proposal

lumi2021:
some new features and a little changing on the structure.

in reality, it's not finished, i'm posting here to have a better revision and discursion about the features that i'm trying to propose

(also some helps with possible english mistakes)

all feedback will be very well received

### DIFF
--- a/documentation/proposals/Proposal - Multi-Backend Input.md
+++ b/documentation/proposals/Proposal - Multi-Backend Input.md
@@ -182,8 +182,6 @@ public partial class InputContext
     public Keyboards Keyboards { get; }
     public Gamepads Gamepads { get; }
     public Joysticks Joysticks { get; }
-    public InputMap? Map { get; set; }
-    public event Action<dynamic>? InputEvent;
     public IReadOnlyList<IInputDevice> Devices { get; }
     public IList<IInputBackend> Backends { get; }
     public event Action<ConnectionEvent>? ConnectionChanged;
@@ -206,13 +204,11 @@ By virtue of the `State` properties not updating until `IInputBackend.Update` is
 ## Custom List Types
 
 These are relatively simple list wrappers with the events fired when state changes.
-All of them have one same event called `InputEvent`. This event is fired for any state changed in the device.
 
 ```cs
 public partial class Mice : IReadOnlyList<IMouse>
 {
     public MouseClickConfiguration ClickConfiguration { get; set; }
-    public event Action<dynamic>? InputEvent;
     public event Action<MouseDownEvent>? ButtonDown;
     public event Action<MouseUpEvent>? ButtonUp;
     public event Action<MouseClickEvent>? Click;
@@ -223,7 +219,6 @@ public partial class Mice : IReadOnlyList<IMouse>
 
 public partial class Keyboards : IReadOnlyList<IKeyboard>
 {
-    public event Action<dynamic>? InputEvent;
     public event Action<KeyDownEvent>? KeyDown;
     public event Action<KeyUpEvent>? KeyUp;
     public event Action<KeyCharEvent>? KeyChar;
@@ -231,7 +226,6 @@ public partial class Keyboards : IReadOnlyList<IKeyboard>
 
 public partial class Gamepads : IReadOnlyList<IGamepad>
 {
-    public event Action<dynamic>? InputEvent;
     public event Action<GamepadDownEvent>? ButtonDown;
     public event Action<GamepadUpEvent>? ButtonUp;
     public event Action<GamepadThumbstickMoveEvent>? ThumbstickMove;
@@ -240,7 +234,6 @@ public partial class Gamepads : IReadOnlyList<IGamepad>
 
 public partial class Joysticks : IReadOnlyList<IJoystick>
 {
-    public event Action<dynamic>? InputEvent;
     public event Action<JoystickDownEvent>? ButtonDown;
     public event Action<JoystickUpEvent>? ButtonUp;
     public event Action<JoystickAxisMoveEvent>? AxisMove;
@@ -265,24 +258,26 @@ This will be configurable on `Mice` (i.e. via `InputContext.Mice.ClickConfigurat
 Unlike 1.0 and 2.0, this proposal uses `readonly record struct`s as their only argument for the event action. This allows us to provide more information to the event handlers without breaking in the future. These types are farily simple:
 
 ```cs
-public readonly record struct ConnectionEvent(IInputDevice Device, DateTime Timestamp, bool IsConnected);
-public readonly record struct KeyDownEvent(IKeyboard Keyboard, DateTime Timestamp, Key Key, bool IsRepeat);
-public readonly record struct KeyUpEvent(IKeyboard Keyboard, DateTime Timestamp, Key Key);
-public readonly record struct KeyCharEvent(IKeyboard Keyboard, DateTime Timestamp, char Character);
-public readonly record struct MouseDownEvent(IMouse Mouse, DateTime Timestamp, Vector2 Position, MouseButton Button);
-public readonly record struct MouseUpEvent(IMouse Mouse, DateTime Timestamp, Vector2 Position, MouseButton Button);
-public readonly record struct MouseMoveEvent(IMouse Mouse, DateTime Timestamp, Vector2 Position, Vector2 Delta);
-public readonly record struct MouseScrollEvent(IMouse Mouse, DateTime Timestamp, Vector2 Position, Vector2 WheelPosition, Vector2 Delta);
-public readonly record struct MouseClickEvent(IMouse Mouse, DateTime Timestamp, Vector2 Position, MouseButton Button);
-public readonly record struct JoystickDownEvent(IJoystick Joystick, DateTime Timestamp, JoystickButton Button);
-public readonly record struct JoystickUpEvent(IJoystick Joystick, DateTime Timestamp, JoystickButton Button);
-public readonly record struct JoystickHatMoveEvent(IJoystick Joystick, DateTime Timestamp, Vector2 Value, Vector2 Delta);
-public readonly record struct JoystickAxisMoveEvent(IJoystick Joystick, DateTime Timestamp, int Axis, float Value, float Delta);
-public readonly record struct GamepadDownEvent(IGamepad Gamepad, DateTime Timestamp, JoystickButton Button);
-public readonly record struct GamepadUpEvent(IGamepad Gamepad, DateTime Timestamp, JoystickButton Button);
-public readonly record struct GamepadThumbstickMoveEvent(IJoystick Joystick, DateTime Timestamp, Vector2 Value, Vector2 Delta);
-public readonly record struct GamepadTriggerMoveEvent(IJoystick Joystick, DateTime Timestamp, int Axis, float Value, float Delta);
+public readonly record struct ConnectionEvent(IInputDevice Device, DateTimeOffset? Timestamp, bool IsConnected);
+public readonly record struct KeyDownEvent(IKeyboard Keyboard, DateTimeOffset? Timestamp, Key Key, bool IsRepeat);
+public readonly record struct KeyUpEvent(IKeyboard Keyboard, DateTimeOffset? Timestamp, Key Key);
+public readonly record struct KeyCharEvent(IKeyboard Keyboard, DateTimeOffset? Timestamp, char Character);
+public readonly record struct MouseDownEvent(IMouse Mouse, DateTimeOffset? Timestamp, Vector2 Position, MouseButton Button);
+public readonly record struct MouseUpEvent(IMouse Mouse, DateTimeOffset? Timestamp, Vector2 Position, MouseButton Button);
+public readonly record struct MouseMoveEvent(IMouse Mouse, DateTimeOffset? Timestamp, Vector2 Position, Vector2 Delta);
+public readonly record struct MouseScrollEvent(IMouse Mouse, DateTimeOffset? Timestamp, Vector2 Position, Vector2 WheelPosition, Vector2 Delta);
+public readonly record struct MouseClickEvent(IMouse Mouse, DateTimeOffset? Timestamp, Vector2 Position, MouseButton Button);
+public readonly record struct JoystickDownEvent(IJoystick Joystick, DateTimeOffset? Timestamp, JoystickButton Button);
+public readonly record struct JoystickUpEvent(IJoystick Joystick, DateTimeOffset? Timestamp, JoystickButton Button);
+public readonly record struct JoystickHatMoveEvent(IJoystick Joystick, DateTimeOffset? Timestamp, Vector2 Value, Vector2 Delta);
+public readonly record struct JoystickAxisMoveEvent(IJoystick Joystick, DateTimeOffset? Timestamp, int Axis, float Value, float Delta);
+public readonly record struct GamepadDownEvent(IGamepad Gamepad, DateTimeOffset? Timestamp, JoystickButton Button);
+public readonly record struct GamepadUpEvent(IGamepad Gamepad, DateTimeOffset? Timestamp, JoystickButton Button);
+public readonly record struct GamepadThumbstickMoveEvent(IJoystick Joystick, DateTimeOffset? Timestamp, Vector2 Value, Vector2 Delta);
+public readonly record struct GamepadTriggerMoveEvent(IJoystick Joystick, DateTimeOffset? Timestamp, int Axis, float Value, float Delta);
 ```
+
+This is the part of this proposal that incorporates the ideas in Enhanced Input Events, and is why this proposal supersedes that one.
 
 One final point to note is that throughout the rest of the proposal the following type will be used:
 
@@ -720,7 +715,16 @@ public readonly record struct GamepadState
 
 `Thumbsticks` contain the two thumbsticks on this gamepad. The X and Y values within this list range from -1 to 1: -1 being leftmost, and 1 being rightmost.
 
-`Triggers` contains the two triggers on thsi gamepad. The values within this list range from 0 to 1: 0 being unpressed, and 1 being fully pressed.
+`Triggers` contains the two triggers on this gamepad. The values within this list range from 0 to 1: 0 being unpressed, and 1 being fully pressed.
+
+Note the use of the `DualReadOnlyList` type. This is basically just:
+```cs
+public readonly struct DualReadOnlyList<T> : IReadOnlyList<T>
+{
+    public readonly T Left;
+    public readonly T Right;
+}
+```
 
 This is used where the list will only ever have exactly two elements, mainly because the "gamepad" form factor is standard and it doesn't make sense to have multiple thumbsticks or triggers given a human only has two thumbs or index fingers. More exotic devices should be exposed using the joystick API.
 
@@ -830,103 +834,6 @@ public interface IJoystickInputHandler : IInputHandler
 `HandleAxisMove` must be called when any value of `JoystickState.Axes` changes.
 
 `HandleHatMove` must be called when any value of `JoystickState.Hats` changes.
-
-# Input Maps
-
-## Presentation
-Input maps are designed to be easyer to the user to handle actions with more than one input.
-At first view it can looks useless and a bad practice, but this is complemtelly useful for cross-plataform programs, being games an example.
-
-This is an pseudocode example of what the Input Map system solves:
-
-Without inputMaps:
-```cs
-inputContext.Update();
-
-private Dictionary<string, bool> pressedKeys = new();
-
-inputContext.Gamepads.ThumbstickMove += @event => {Move(@event.Value)};
-inputContext.Joysticks.AxisMove += @event => {Move(@event.Value)};
-
-private Move(Vector2D<float> value)
-{
-    // Code for movement here
-}
-```
-
-With inputMaps:
-```cs
-inputContext.Update();
-
-// This single line get rid of the two in the example above. TWO ISN'T THE LIMIT!
-inputContext.Map.GetAction("Move").Axis += @event => {Move(@event.Value)};
-
-private Move(Vector2D<float> value)
-{
-    // Code for movement here
-}
-```
-
-## Implementation
-Each InputContext have one reference to an InputMap.
-This is how the InputMap class work:
-```cs
-public partial class InputMap
-{
-    public InputMapAction GetAction(string name);
-    public void AddAction(InputMapAction action, string name);
-    public void RemoveAction(string name);
-    public static InputMap Load(string path);
-    public static void Save(InputMap source, string path);
-    public static string Compile(InputMap source);
-}
-```
-
-`GetAction` method is the main entry point of the system. It allows the use to get an mapped action by it configurated name.
-
-```cs
-public partial class InputMapAction
-{
-    public bool IsPressed { get; }
-    public event Action<AxisInputEvent> Axis;
-}
-```
-
-## Configuration and saving
-The input map system's save system is one of the most important to it working. It helps to make the configuration easyer as faster.
-
-Instead of needing to hardcode all the actions, the user should create a file with a special syntax. This is an example of the sintax of the map system's saved data:
-```
-"move_up":
-    - KeyName.Up
-    - JoystickButton.DPadUp
-
-"move_down":
-- KeyName.Down
-- JoystickButton.DPadDown
-
-"move":
-    - Vector2D [
-        -KeyName.Down
-        -KeyName.Up
-        -KeyName.Left
-        -KeyName.Right
-    ]
-    - Joystick.Axis
-```
-
-The main syntax may somewhat resemble YAML, but it's much more symple and easy to parse.
-
-The main structure is based on a list, with it main feature having the line separation as the main way to separate the information. Also, the parsing is based on an finite-state model. The last data seted will recive the next configurations until a higher level information be setted.
-
-The label are wrote inside quotes ("") and ended with a collom (:). This starts a new `InputMapAction`.
-each new input are declared with a minnus sign (-) and a stace. After that, the name of the correct enum and the correct input.
-
-Input system can easly return vectors. after the minus sign, it's possible choose a vector type to recive (`Vector2D`, `Vector3D` or `Vector4D`) and the inputs that will math witch each axis, respectively.
-
-Comments are not allowed in these files.
-
-To Load an `InputMap` from a file, the static method `Load` should be used. in any case of edition during runtime, it's possible to save the changes using the method `Save`.
 
 # Meeting Notes
 
--- a/documentation/proposals/Proposal - Multi-Backend Input.md
+++ b/documentation/proposals/Proposal - Multi-Backend Input.md
@@ -3,11 +3,12 @@ Proposal API for backend-agnostic, refactored Input via keyboards, mice, and con
 
 # Contributors
 - Dylan Perks (@Perksey)
+- Lumi (@Lumi2021)
 
 # Current Status
 - [x] Proposed
-- [x] Discussed with Working Group
-- [x] Approved
+- [ ] Discussed with Working Group
+- [ ] Approved
 - [ ] Implemented
 
 # Design Decisions
@@ -181,6 +182,8 @@ public partial class InputContext
     public Keyboards Keyboards { get; }
     public Gamepads Gamepads { get; }
     public Joysticks Joysticks { get; }
+    public InputMap? Map { get; set; }
+    public event Action<dynamic>? InputEvent;
     public IReadOnlyList<IInputDevice> Devices { get; }
     public IList<IInputBackend> Backends { get; }
     public event Action<ConnectionEvent>? ConnectionChanged;
@@ -203,11 +206,13 @@ By virtue of the `State` properties not updating until `IInputBackend.Update` is
 ## Custom List Types
 
 These are relatively simple list wrappers with the events fired when state changes.
+All of them have one same event called `InputEvent`. This event is fired for any state changed in the device.
 
 ```cs
 public partial class Mice : IReadOnlyList<IMouse>
 {
     public MouseClickConfiguration ClickConfiguration { get; set; }
+    public event Action<dynamic>? InputEvent;
     public event Action<MouseDownEvent>? ButtonDown;
     public event Action<MouseUpEvent>? ButtonUp;
     public event Action<MouseClickEvent>? Click;
@@ -218,6 +223,7 @@ public partial class Mice : IReadOnlyList<IMouse>
 
 public partial class Keyboards : IReadOnlyList<IKeyboard>
 {
+    public event Action<dynamic>? InputEvent;
     public event Action<KeyDownEvent>? KeyDown;
     public event Action<KeyUpEvent>? KeyUp;
     public event Action<KeyCharEvent>? KeyChar;
@@ -225,6 +231,7 @@ public partial class Keyboards : IReadOnlyList<IKeyboard>
 
 public partial class Gamepads : IReadOnlyList<IGamepad>
 {
+    public event Action<dynamic>? InputEvent;
     public event Action<GamepadDownEvent>? ButtonDown;
     public event Action<GamepadUpEvent>? ButtonUp;
     public event Action<GamepadThumbstickMoveEvent>? ThumbstickMove;
@@ -233,6 +240,7 @@ public partial class Gamepads : IReadOnlyList<IGamepad>
 
 public partial class Joysticks : IReadOnlyList<IJoystick>
 {
+    public event Action<dynamic>? InputEvent;
     public event Action<JoystickDownEvent>? ButtonDown;
     public event Action<JoystickUpEvent>? ButtonUp;
     public event Action<JoystickAxisMoveEvent>? AxisMove;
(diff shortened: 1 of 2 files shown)


documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -257,27 +265,25 @@ This will be configurable on `Mice` (i.e. via `InputContext.Mice.ClickConfigurat
...
-public readonly record struct MouseClickEvent(IMouse Mouse, Vector2 Position, MouseButton Button);
-public readonly record struct JoystickDownEvent(IJoystick Joystick, JoystickButton Button);
-public readonly record struct JoystickUpEvent(IJoystick Joystick, JoystickButton Button);
-public readonly record struct JoystickHatMoveEvent(IJoystick, Vector2 Value, Vector2 Delta);
-public readonly record struct JoystickAxisMoveEvent(IJoystick Joystick, int Axis, float Value, float Delta);
-public readonly record struct GamepadDownEvent(IGamepad Gamepad, JoystickButton Button);
-public readonly record struct GamepadUpEvent(IGamepad Gamepad, JoystickButton Button);
-public readonly record struct GamepadThumbstickMoveEvent(IJoystick, Vector2 Value, Vector2 Delta);
-public readonly record struct GamepadTriggerMoveEvent(IJoystick Joystick, int Axis, float Value, float Delta);
+public readonly record struct ConnectionEvent(IInputDevice Device, DateTime Timestamp, bool IsConnected);
```
Perksey, line 268: These should be DateTimeOffset and should be nullable and the last parameter of these records. This is to ensure compatibility with backends that do not provide the timestamp at which the event occurred, and also means that this won't need to go through working group approval at all given that it is a modification the structs are explicitly designed for (see the already-approved comment about "This allows us to provide more information to the event handlers without breaking in the future.")

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -257,27 +265,25 @@ This will be configurable on `Mice` (i.e. via `InputContext.Mice.ClickConfigurat
...
+public readonly record struct JoystickHatMoveEvent(IJoystick Joystick, DateTime Timestamp, Vector2 Value, Vector2 Delta);
+public readonly record struct JoystickAxisMoveEvent(IJoystick Joystick, DateTime Timestamp, int Axis, float Value, float Delta);
+public readonly record struct GamepadDownEvent(IGamepad Gamepad, DateTime Timestamp, JoystickButton Button);
+public readonly record struct GamepadUpEvent(IGamepad Gamepad, DateTime Timestamp, JoystickButton Button);
+public readonly record struct GamepadThumbstickMoveEvent(IJoystick Joystick, DateTime Timestamp, Vector2 Value, Vector2 Delta);
+public readonly record struct GamepadTriggerMoveEvent(IJoystick Joystick, DateTime Timestamp, int Axis, float Value, float Delta);
 ```
 
-This is the part of this proposal that incorporates the ideas in Enhanced Input Events, and is why this proposal supersedes that one.
-
```
Perksey, line 280: This is important background information, please do not remove.

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -716,15 +722,6 @@ public readonly record struct GamepadState
...
 
-Note the use of the `DualReadOnlyList` type. This is basically just:
-```cs
-public readonly struct DualReadOnlyList<T> : IReadOnlyList<T>
-{
-    public readonly T Left;
-    public readonly T Right;
-}
-```
-
```
Perksey, line 727: Add this back please.

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -834,6 +831,103 @@ public interface IJoystickInputHandler : IInputHandler
 
 `HandleHatMove` must be called when any value of `JoystickState.Hats` changes.
 
+# Input Maps
```
Perksey, line 834: This should be a separate proposal given that there's no interest in having this as a core part of input, rather an extension to the existing input library. This also means we don't need to implement it for 3.0, rather we can do so at anytime in 3.X.

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -834,6 +831,103 @@ public interface IJoystickInputHandler : IInputHandler
...
+
+## Presentation
+Input maps are designed to be easyer to the user to handle actions with more than one input.
```
Perksey, line 837:
```suggestion
Input maps are designed to be easier to the user to handle actions with more than one input.
```

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -834,6 +831,103 @@ public interface IJoystickInputHandler : IInputHandler
...
+At first view it can looks useless and a bad practice, but this is complemtelly useful for cross-plataform programs, being games an example.
```
Perksey, line 838: This is a useless comment and contains unfounded claims you have not backed up.
- Why does it look useless?
- Why does it look like bad practice?
- Why specifically is it useful, rather than just saying it's useful?
- Why is it useful for games specifically?
- Why is it useful for cross-platform programs specifically?

Perksey, line 838:
```suggestion
At first view it can looks useless and a bad practice, but this is completely useful for cross-platform programs, being games an example.
```

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -834,6 +831,103 @@ public interface IJoystickInputHandler : IInputHandler
...
+
+This is an pseudocode example of what the Input Map system solves:
```
Perksey, line 840: Problems are best described by words. You haven't stated anywhere in this proposal what this actually achieves and why this is a valuable proposal for the working group to consider.

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -834,6 +831,103 @@ public interface IJoystickInputHandler : IInputHandler
...
+// This single line get rid of the two in the example above. TWO ISN'T THE LIMIT!
+inputContext.Map.GetAction("Move").Axis += @event => {Move(@event.Value)};
+
+private Move(Vector2D<float> value)
+{
+    // Code for movement here
+}
+```
+
+## Implementation
```
Perksey, line 870: Consider changing this header, implementation details should be INFORMATIVE TEXT only rather than a normative part of the proposal. I know that's probably not what you intend this section to specify, which is why I say change the header.

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -834,6 +831,103 @@ public interface IJoystickInputHandler : IInputHandler
...
+Each InputContext have one reference to an InputMap.
```
Perksey, line 871: Why have you done the ownership this way around? Surely it makes sense for an input map to own an input context and not the other way around given that the map is essentially an extension/grouping of the existing functionality? This also relates to my other comment about making this a separate proposal.

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -834,6 +831,103 @@ public interface IJoystickInputHandler : IInputHandler
...
+This is how the InputMap class work:
```
Perksey, line 872: An API declaration does not show _how_ something works.

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -834,6 +831,103 @@ public interface IJoystickInputHandler : IInputHandler
...
+```cs
+public partial class InputMap
+{
+    public InputMapAction GetAction(string name);
```
Perksey, line 876: Consider providing documentation comments and/or MoSCoW/RFC-style requirements as done throughout the 3.0 proposals. As it stands I have no idea what any of this must do and would have no idea how to implement it from this document alone.

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -834,6 +831,103 @@ public interface IJoystickInputHandler : IInputHandler
...
+    public void AddAction(InputMapAction action, string name);
+    public void RemoveAction(string name);
+    public static InputMap Load(string path);
```
Perksey, line 879: File format? Should this be included in the name?

Perksey, line 879: Do we want to make this exclusive to system IO? Why not provide a ROSpan overload?

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -834,6 +831,103 @@ public interface IJoystickInputHandler : IInputHandler
...
+    public static void Save(InputMap source, string path);
+    public static string Compile(InputMap source);
```
Perksey, line 881: What does this even mean?

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -834,6 +831,103 @@ public interface IJoystickInputHandler : IInputHandler
...
+}
+```
+
+`GetAction` method is the main entry point of the system. It allows the use to get an mapped action by it configurated name.
```
Perksey, line 885:
```suggestion
`GetAction` method is the main entry point of the system. It allows the use to get a mapped action by its configured name.
```

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -834,6 +831,103 @@ public interface IJoystickInputHandler : IInputHandler
...
+
+```cs
+public partial class InputMapAction
+{
+    public bool IsPressed { get; }
```
Perksey, line 890: What does this mean for actions that aren't boolean on/off actions?

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -834,6 +831,103 @@ public interface IJoystickInputHandler : IInputHandler
...
+    public event Action<AxisInputEvent> Axis;
```
Perksey, line 891: How does this work with mapping WASD as a substitute for a thumbstick?

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -834,6 +831,103 @@ public interface IJoystickInputHandler : IInputHandler
...
+    public static InputMap Load(string path);
+    public static void Save(InputMap source, string path);
+    public static string Compile(InputMap source);
+}
+```
+
+`GetAction` method is the main entry point of the system. It allows the use to get an mapped action by it configurated name.
+
+```cs
+public partial class InputMapAction
```
Perksey, line 888: Generally I think more definition needs to be done about what's in scope and what's out of scope for this model.

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -834,6 +831,103 @@ public interface IJoystickInputHandler : IInputHandler
...
+{
+    public bool IsPressed { get; }
+    public event Action<AxisInputEvent> Axis;
+}
+```
+
+## Configuration and saving
+The input map system's save system is one of the most important to it working. It helps to make the configuration easyer as faster.
```
Perksey, line 896: Please use a spellchecker (e.g. Grammarly)
```suggestion
The input map system's save system is one of the most important to it working. It helps to make the configuration easier and faster.
```

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -834,6 +831,103 @@ public interface IJoystickInputHandler : IInputHandler
...
+
+Instead of needing to hardcode all the actions, the user should create a file with a special syntax. This is an example of the sintax of the map system's saved data:
```
Perksey, line 898:
```suggestion
Instead of needing to hardcode all the actions, the user should create a file with a special syntax. This is an example of the syntax of the map system's saved data:
```

Perksey, line 898: Implying that this is the only route of entry?

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -834,6 +831,103 @@ public interface IJoystickInputHandler : IInputHandler
...
+```
+"move_up":
+    - KeyName.Up
+    - JoystickButton.DPadUp
+
+"move_down":
+- KeyName.Down
+- JoystickButton.DPadDown
```
Perksey, line 906: Are the padding differences significant? What are the parsing rules here?

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -834,6 +831,103 @@ public interface IJoystickInputHandler : IInputHandler
...
+
+"move":
+    - Vector2D [
+        -KeyName.Down
+        -KeyName.Up
+        -KeyName.Left
+        -KeyName.Right
+    ]
```
Perksey, line 914: What is the significant of Vector2D and why does the contained elements have more than 2 elements? Are you intending for this to be the WASD mapping I alluded to earlier? In which case, what are the rules for this? e.g. how do we map each key to a specific value on the axes.

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -834,6 +831,103 @@ public interface IJoystickInputHandler : IInputHandler
...
+    - Joystick.Axis
```
Perksey, line 915: Which joystick will this pick up on? Any?

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -834,6 +831,103 @@ public interface IJoystickInputHandler : IInputHandler
...
+```
+"move_up":
+    - KeyName.Up
+    - JoystickButton.DPadUp
+
+"move_down":
+- KeyName.Down
+- JoystickButton.DPadDown
+
+"move":
```
Perksey, line 908: How does this handle multiple controllers e.g. for multiple players?

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -834,6 +831,103 @@ public interface IJoystickInputHandler : IInputHandler
...
+    - Vector2D [
+        -KeyName.Down
+        -KeyName.Up
+        -KeyName.Left
+        -KeyName.Right
+    ]
+    - Joystick.Axis
+```
+
+The main syntax may somewhat resemble YAML, but it's much more symple and easy to parse.
```
Perksey, line 918: Why are we making our own file format? This is scope creep and beyond our engineering resources to execute. Using a well-supported format such as JSON is perfectly fine, a custom format is over engineering.

Perksey, line 918:
```suggestion
The main syntax may somewhat resemble YAML, but it's much more simple and easy to parse.
```

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -834,6 +831,103 @@ public interface IJoystickInputHandler : IInputHandler
...
+
+The main structure is based on a list, with it main feature having the line separation as the main way to separate the information. Also, the parsing is based on an finite-state model. The last data seted will recive the next configurations until a higher level information be setted.
```
Perksey, line 920: You can't say it's based on a finite state machine without specifying the states of said state machine! How are we supposed to implement that if so? Also, surely this is an implementation detail that should be non-normative?

Perksey, line 920: Grammar/spelling issues. I have no idea what the last part of this sentence is trying to say so I haven't attempted to reword.

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -834,6 +831,103 @@ public interface IJoystickInputHandler : IInputHandler
...
+
+The label are wrote inside quotes ("") and ended with a collom (:). This starts a new `InputMapAction`.
```
Perksey, line 922: Consider providing a grammar for this? e.g. ANTLR

Perksey, line 922:
```suggestion
The label is written inside of quotes ("") and delimited by a colon (:). This starts a new `InputMapAction`.
```

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -834,6 +831,103 @@ public interface IJoystickInputHandler : IInputHandler
...
+each new input are declared with a minnus sign (-) and a stace. After that, the name of the correct enum and the correct input.
```
Perksey, line 923:
```suggestion
each new input are declared with a minus sign (-) and a space. After that, the name of the correct enum and the correct input.
```

Perksey, line 923: But how do we discriminate between inputs from different devices or different types of devices?

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -834,6 +831,103 @@ public interface IJoystickInputHandler : IInputHandler
...
+
+Input system can easly return vectors. after the minus sign, it's possible choose a vector type to recive (`Vector2D`, `Vector3D` or `Vector4D`) and the inputs that will math witch each axis, respectively.
+
+Comments are not allowed in these files.
```
Perksey, line 927: This is such an arbitrary restriction and is one of the millions of reasons why I am reluctant to have a custom format

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -834,6 +831,103 @@ public interface IJoystickInputHandler : IInputHandler
...
+```
+
+The main syntax may somewhat resemble YAML, but it's much more symple and easy to parse.
+
+The main structure is based on a list, with it main feature having the line separation as the main way to separate the information. Also, the parsing is based on an finite-state model. The last data seted will recive the next configurations until a higher level information be setted.
+
+The label are wrote inside quotes ("") and ended with a collom (:). This starts a new `InputMapAction`.
+each new input are declared with a minnus sign (-) and a stace. After that, the name of the correct enum and the correct input.
+
+Input system can easly return vectors. after the minus sign, it's possible choose a vector type to recive (`Vector2D`, `Vector3D` or `Vector4D`) and the inputs that will math witch each axis, respectively.
```
Perksey, line 925: "return vectors" the example above uses keyboard buttons as well and does not specify in terms of requirements for how to map this into a vector. How is the axis matching done? (i.e. mathematically)

Perksey, line 925:
```suggestion
Input system can easily return vectors. After the minus sign, it's possible choose a vector type to receive (`Vector2D`, `Vector3D` or `Vector4D`) and the inputs that will match which each axis, respectively.
```

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -834,6 +831,103 @@ public interface IJoystickInputHandler : IInputHandler
...
+
+Comments are not allowed in these files.
+
+To Load an `InputMap` from a file, the static method `Load` should be used. in any case of edition during runtime, it's possible to save the changes using the method `Save`.
```
Perksey, line 929:
```suggestion
To load an `InputMap` from a file, the static method `Load` should be used. If the map is edited at runtime, it's possible to save the changes using the method `Save`.
```

Perksey (changes requested):

Perksey:
For what it's worth we have discussed an input action model before and have told the working group we'll look into it, but we don't even remotely have the bandwidth to do so before 3.0's RTM release.

documentation/proposals/Proposal - Multi-Backend Input.md
Beyley: imo it shouldnt be silk's job to do any IO here, the user should just supply a string for the map contents, and "Save" should just return a string for the saved map

documentation/proposals/Proposal - Multi-Backend Input.md
Beyley: i personally would lean towards ini due to its simplicity for non-developers to read/write.
Although the question is, are end users (eg. a player of a game built on Silk.NET) ever intended to modify this file? Or is this mapping file for developers only? The API being hardcoded to read/write to a path implies to me that its meant to be handled by end-users aswell, and JSON can be quite daunting to newcomers

documentation/proposals/Proposal - Multi-Backend Input.md
```diff
@@ -203,11 +206,13 @@ By virtue of the `State` properties not updating until `IInputBackend.Update` is
 ## Custom List Types
 
 These are relatively simple list wrappers with the events fired when state changes.
+All of them have one same event called `InputEvent`. This event is fired for any state changed in the device.
 
 ```cs
 public partial class Mice : IReadOnlyList<IMouse>
 {
     public MouseClickConfiguration ClickConfiguration { get; set; }
+    public event Action<dynamic>? InputEvent;
```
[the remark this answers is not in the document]
lumi2021: all the events are structs, that are faster but does not allow inheritance so i can't make a common parent to type it. For me is a really important option to can receive any input and can filtrate it as a type so i don't want to just delete it for now. do you know some way to solve his problem?

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: we can't provide an timestamp anyway? Maybe it could be wrong for some milliseconds or less, but in the case of don't receive a timestamp, it's not possible to send a DateTime.Now()?

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: i know, but i think this is really out of place because it's on the middle of the reading. You see, i don't deleted all the information and there's the credits of the Enhanced Input Events proposal in the start of the file
You think we can just move it or something? Part of my proposal is bellow and it can cause misunderstanding

Perksey: The Working Group disagrees given they have approved this information in its place. I don’t think this is what will cause misunderstanding.

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: it was already presented, why you want to i put it if you already explained the DualReadOnlyList<T> implementation because it would appear in the rest of the text?

documentation/proposals/Proposal - Multi-Backend Input.md
Beyley: > we can't provide an timestamp anyway? Maybe it could be wrong for some milliseconds or less, but in the case of don't receive a timestamp, it's not possible to send a DateTime.Now()?

If the backend doesnt provide timestamps for inputs, i'd rather us not lie about it doing so, if the user needs a timestamp for an event, they can do `Timestamp ?? DateTime.Now()` themselves

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: idk, ask for beyley, she said it or something of the kind lol

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: yeah, the section don't teach you to implement nothing, just maps the base of the system... i will fix

documentation/proposals/Proposal - Multi-Backend Input.md
Perksey: No idea what you’re saying, but I’m not intentionally removing an API declaration that is used throughout the rest of the proposal. It creates needless confusion.

documentation/proposals/Proposal - Multi-Backend Input.md
Beyley: What i had said "looks useless and bad practice" was the combination of multiple events into a single callback, i dont see a reason to combine keyboard and mouse events into a single callback, when you'd have to `if(ev is KeyboardEvent kbEvent)`, im not against input *maps* being in Silk.NET (_as an extension_, **not** core)

what i dont like is `public event Action<dynamic>? InputEvent;`

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: the map is an extension/grouping of the receiving of input events. It's on the same level of a ButtonDown or a ButtonUp, the only difference is because it will filtrate and operate the inputs before send it back.
i really can't see how a InputMap can own an InputContext because of it

documentation/proposals/Proposal - Multi-Backend Input.md
Perksey: No because it isn't clear what problem you're solving. e.g. what is an input action and why is it useful to hide the actual callbacks behind one?

documentation/proposals/Proposal - Multi-Backend Input.md
Perksey: Nonetheless, we are eager to ensure that this is strictly an extension to the core API and not part of the core API itself, so consider other ways to express this relationship. Inheritance maybe? Composition?

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: True, thx

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: If you're saying, ok

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: What I'm trying to say is you already put the implementation of the structure, a little above of where you said to i put back the info of where starts the part referencing the other proposal that I can't remember the name (sry) and i think make it two times is unnecessary

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: Oh, reality? Sry so, my fault. I will fix it

lumi2021: > So to clarify, you’re making a proposal you don’t personally understand the usefulness in?

I just don't see anything useless on it, i just misunderstood Beyley's comment. But anyway you have to say, it really looks a little useless at first look lol

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: Oh, sure sure! I really have not seen that i should follow that way when i wrote but now i can see it, thx

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: I really can't see how i can make it work with inheritance becuase the inputContext are really complex and there's a lot od useless information to the maps, so i will try to do composition, also because you will always get an InputContext from the core window so (i really don't want to make it using extension methods, i believe in composition supremacy)

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: I really want to use more comments but i was a little afraid because I was only using this proposal as a reference. I will check how it is done in another one and try to describe better

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: Yes the extension of the file is .File Format

Jokes apart, I don't have a specific file format yet, but it should be a problem for the system that should just read and parse its content. So while I don't have creativity to think of something to use as an standard you can have a inputMap.FileFormat 👍

lumi2021: Really? I was thinking about just returning the serialized content as a string but for me if the library does the IO work, better

If it's not silk's job i will change so, it's not like it's a problem

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: If you're talking about the `compile` method, it serializes and returns the result as a string. I know, the name is bad. I will change it, i really forgot the word for a moment

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: Good question... I was thinking about creating classes for each type of input, you think it's better in that way? As it's a class there's no reason to put all possible returns in the same type

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: It's on the input map syntax example.

If an input doesn't return a vector, you can write

```
- silkVectorClass [

]
```

And the specified inputs inside it, -1 +1, for each axis in the vector class, one input per line.
So like (pseudocode):

```
- Vector2D [
    - A // -x
    - D // +x
    - S // -y
    - W  // +y
]
```

If it's not explained enough in the proposal pls notify

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: This model is really raw, huh... I need to improve it a lot

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: Yeah, i have it installed on my work pc but not in my personal one lol i will do it (also back to Duolingo or a better English course probably)

documentation/proposals/Proposal - Multi-Backend Input.md
Perksey: Right, but this isn't an implementation detail, this is an API definition. It is normative and should be part of the specification.

documentation/proposals/Proposal - Multi-Backend Input.md
Perksey: For reasons such as this, I really wish you had waited to make this proposal to be honest, this isn't a one person job by any stretch of the imagination and we're likely to have to back-track on anything decided here once we actually start implementing it, which could be years from now.

Feel free to come up with an alternative design, but I don't think we can have enough confidence having so little available information. Both the windowing and input proposals came from multiple maintainers experimenting with potential designs and actively collaborating in a short space of time to come up with what feels the most natural both to use and implement, and we just don't have this information available for your proposal because it is a one person job. That's not to say that one person proposals aren't accepted, they definitely have been in the past, but historically these have been the proposals we've come to regret implementing the most. And we're just not interested in doing that experimentation right now because we have so many more pressing priorities before we can even look at extensions such as this. This is also one of the many reasons I put a comment to make this a separate proposal, so that it doesn't impact our much higher priority work.

documentation/proposals/Proposal - Multi-Backend Input.md
Perksey: I do not think it's better in that way. I don't know what could make it better to be honest.

documentation/proposals/Proposal - Multi-Backend Input.md
Perksey: Okay, this is not explained in the proposal (at all) particularly your latter code snippet.

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: These are the two options and the file configuration should be the best choice, i should considerate another option or I'm misunderstanding something?

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: I understand, i will put this section in a separated proposal and try to improve the maximum that i can

lumi2021: Also you want to hold it for a while until someone help to develop it better? Posting the proposal looks like the best way people can see and modify it, but it will be a problem if it gets rejected

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: Sure, bad word use

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: Should not*** be a problem

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: Tbh i really do not want to do it with separate classes also... I will try to think a way to everything works just inside the InputMap class, without unnecessary data access

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: I will try to improve it more. Also improve the syntax explanation with some other features like ranges and dead radius (this is the name? If it's not sry i will obviously search when i can)

Also some way of returning more customizable values or use modifiers like shift or ctrl

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: 905 & 906 are exactly to show padding isn't a rule, but i will write about the optional use of the white space in this case

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: Vector2D is to define the specified vector. It can be any vector, 2, 3 or 4

The question of why there's 4 keys is because the first is for the negative value and the second is for the positive value, being it for each axis

So the size of the list will aways be the axis count * 2

And yes, is what you asked before

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: I need to say, I don't understand so much what devices "joystick" and "gamepad" are referring. The idea should be any, but now I'm thinking about the fact that a ps4 controller have 2 thumbsticks and i forgot it completely lol

Can you explain me a little what exactly these devices represent? Like i really never head something called anything for these names (obviously in some cases but pretty obvious ones like the atari controller or the one that some flying machines uses, that's i think is out of where silk acts lol), just everything as controller

Also like what input will be returned by these weird car controllers or joysticks (are these joysticks that it refers?)?

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: Thing that i don't think again (or at least tried but can't) lol

In the case of the same input device, different bindings

In the case of different devices, It could return the data of each device separated (like a get mappedByDevice()) but it can broke a lot in some cases...

(Sry busy now i will add more data)

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: I don't want to use existing file formats like json or YAML (that is a base for syntax) because an entire parser for these ones will be unnecessarily expensive for something so simple. Also i want a file format easy for anyone to edit, being developers or users. Also i want something easy and fast to read and parse

This file format can use just a single loop to iterate per line and the state machine-based structure makes you and the machine easily understand the scopes

So i think it can be more optimized and cheaper if we do our own file format that does JUST what we want with an easy to parse syntax instead of use a complex data structure that needs a lot of operations to parse

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: Looks right for me, but i will try to rewrite the explanation of how the state machine works

Is it possible to create graphics and diagrams with github's markdown?

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: Lol no sry, it was late and i was tired, i correct it

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: I really can't see why put comments on it, it's just a structure to save data. Json also doesn't allow comments as standard, but should not be a problem to implement it, so i will try to add some rules sbout

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: Yeah, this parsing needs to be a lot more structured

lumi2021:
> For what it's worth we have discussed an input action model before and have told the working group we'll look into it, but we don't even remotely have the bandwidth to do so before 3.0's RTM release. So even if this is accepted, it won't be implemented for potentially over a year.

I was expecting. 3.0 is in development for some long years, right?

documentation/proposals/Proposal - Multi-Backend Input.md
Perksey: I can guarantee that System.Text.Json will be infinitely faster than any parser that the Silk.NET team has the ability to maintain itself. The same is likely true for extant YAML parsers.

This proposal creates unnecessary maintenance effort for the Silk.NET team that the team won't want to be dealing with, and will likely refuse to implement or merge on those grounds.

documentation/proposals/Proposal - Multi-Backend Input.md
Perksey: What?

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: understand. I will search about the YAML and JSON parsers and if you're right in this case, i will change to one of these two

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: i misunderstood the message sry. if i continue with this file type i will do one

documentation/proposals/Proposal - Multi-Backend Input.md
Perksey: I think placing it in a separate proposal and merging it with an `(On Hold)` prefix is the best course of action.

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: The input map system can't discriminate between input device types, the objective of it is exactly that

About the different device discrimination, i will work on it in the next commit

documentation/proposals/Proposal - Multi-Backend Input.md
lumi2021: i want to discuss a little more about it, can we do it on discord?

Perksey: No, here is fine.

Perksey (approved):
Title has been changed to `(On hold)`, merging.

Issue #241, opened 2020-07-15
Render on Drag or Resize


OAguinagalde:
Hello, this might not be the correct place to ask, but just in case. How would I go about rendering to the window even if I'm resizing/dragging the window around?

I've tried playing around with the Move event and the UseSingleThreadedWindow window option but can't manage.

Thanks!

Perksey:
Hi there, thanks for using Silk.NET!

Sadly this is caused by a fatal flaw baked into the message loop of the Desktop Window Manager that GLFW, our underlying window library, is subject to.

Basically Windows has a message loop that all events are put into, however when you drag or resize a window it actually halts that message loop until you've put it back down, so yeah this is a bug but it's a bug that's become part of the core design and isn't one we can't fix sadly, due to it beign an upstream issue.

If you'd like, you could have a play around with setting up a custom game loop (the source code for the one built in to the library is [here](https://github.com/Ultz/Silk.NET/blob/554af444457e52691cf87c479f9c9a5c77f1ff59/src/Windowing/Silk.NET.Windowing.Common/WindowExtensions.cs/#L44)) and mess around with how/when you call the DoEvents function, as that's the call that blocks.

Other than that there's not really much that can be done to alleviate this. Sorry!

OAguinagalde:
I see, got it. Thanks for the reply, I'm not sure I'm up to the task but I might give it try. Thanks again!

Arugin:
About resizing: 
One way that works: add two calls to drawFrame() in the resize callback. Two calls because the first after each resize will only recreate the swap chain, need a second call to get the actual render.

Perksey:
Closed in #1979 as best we can at this time - you can now render while the message loop is blocked so long as things are actually happening. So if you drag, each pixel change gives you an opportunity to redraw. But if you hold the window still, no drawing will occur. This is a limitation we are unprepared to overcome at this time.

PR #1979, opened 2024-03-07
Update and Repaint on resize.

otac0n:
# Summary of the PR
Repaint on resize/drag.
This does NOT automatically queue paint messages with a timer or the like, so the view still hangs as long as the mouse is held still.

https://github.com/dotnet/Silk.NET/issues/241

# Further Comments
I am not sure how (or if) to implement this for the SdlView.

### DIFF
--- a/src/Windowing/Silk.NET.Windowing.Common/WindowExtensions.cs
+++ b/src/Windowing/Silk.NET.Windowing.Common/WindowExtensions.cs
@@ -43,29 +43,21 @@ namespace Silk.NET.Windowing
         public static void Run(this IView view)
         {
             view.Initialize();
-
-            void OnFrame()
-            {
-                view.DoEvents();
-                DoFrame();
-            }
-
-            void DoFrame()
-            {
-                if (!view.IsClosing)
+            view.Run
+            (
+                () =>
                 {
-                    view.DoUpdate();
+                    view.DoEvents();
+                    if (!view.IsClosing)
+                    {
+                        view.DoUpdate();
+                    }
+                    if (!view.IsClosing)
+                    {
+                        view.DoRender();
+                    }
                 }
-
-                if (!view.IsClosing)
-                {
-                    view.DoRender();
-                }
-            }
-
-            view.Refresh += DoFrame;
-            view.Run(OnFrame);
-            view.Refresh -= DoFrame;
+            );
 
             view.DoEvents();
             view.Reset();
--- a/src/Windowing/Silk.NET.Windowing.Common/WindowExtensions.cs
+++ b/src/Windowing/Silk.NET.Windowing.Common/WindowExtensions.cs
@@ -43,21 +43,29 @@ namespace Silk.NET.Windowing
         public static void Run(this IView view)
         {
             view.Initialize();
-            view.Run
-            (
-                () =>
+
+            void OnFrame()
+            {
+                view.DoEvents();
+                DoFrame();
+            }
+
+            void DoFrame()
+            {
+                if (!view.IsClosing)
                 {
-                    view.DoEvents();
-                    if (!view.IsClosing)
-                    {
-                        view.DoUpdate();
-                    }
-                    if (!view.IsClosing)
-                    {
-                        view.DoRender();
-                    }
+                    view.DoUpdate();
                 }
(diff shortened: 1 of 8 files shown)


src/Windowing/Silk.NET.Windowing.Glfw/GlfwWindow.cs
```diff
@@ -658,6 +659,11 @@ protected override void RegisterCallbacks()
                 FramebufferResize?.Invoke(new(width, height));
             };
 
+            _onRefresh = (window) =>
+            {
+                Refresh?.Invoke();
```
otac0n, line 664: Another feasible option is to keep the API surface the same and call the  Update/Render methods here, perhaps with a zero update value to indicate an inter-frame update.

Perksey:
This has been added to the Working Group Meeting agenda so that we can ratify API changes. Realistically this is minimal, but there are other API changes we have to discuss that we can sweep this under. No date set yet, I have one more proposal to write before we can set one, but hopefully it shouldn't be too long.

otac0n:
https://github.com/dotnet/Silk.NET/assets/193584/a2f2f9c0-c14a-42c9-b945-003529426cdd

src/Windowing/Silk.NET.Windowing.Common/WindowExtensions.cs
```diff
@@ -43,21 +43,29 @@ public static void Invoke(this IView view, Action t)
...
                 }
-            );
+
+                if (!view.IsClosing)
+                {
+                    view.DoRender();
+                }
+            }
+
+            view.Refresh += DoFrame;
```
Perksey, line 66: I appreciate the intent, but I don't think automatically binding to events is good. If we want this to be default behaviour, then maybe we should just make it so. Wouldn't be the worst idea actually, then we wouldn't need Working Group approval at all and we can sweep the SDL vs GLFW behaviour differences under the rug as "undefined behaviour" (seeing as we only officially support SDL on Android and experimentally on iOS)

src/Windowing/Silk.NET.Windowing.Glfw/GlfwWindow.cs
Perksey: I think I'm leaning more towards this after seeing the latest changes (per the above comment)

Perksey (changes requested):
Pre-WG comments

src/Windowing/Silk.NET.Windowing.Glfw/GlfwWindow.cs
```diff
@@ -658,6 +659,18 @@ protected override void RegisterCallbacks()
...
+            {
+                if (!IsClosing)
+                {
+                    DoUpdate();
+                }
+                if (!IsClosing)
+                {
+                    DoRender();
+                }
+            };
```
Perksey, line 672: This is fine, and I'm pretty happy to just merge as is, but one possible Q which I'm not sure how much we should care about is "what if they're not using our Render/Update events?"

What we could do:
- Have a boolean field on the window e.g. `_isInDoEvents` that is set to `true` when `DoEvents` is called.
- If `DoEvents` is called and this is already `true`, `DoEvents` does not execute
- In addition to the `Run` implementation just being a `while (true)` loop by default, we could capture the `Action` passed in and use that for `_onRefresh` instead.

The reason why we need the boolean flag is because the user's action will naturally try to call `DoEvents`, which would cause infinite recursion/undefined behaviour.

So sorry I didn't think of this earlier, I really don't want to mess you around with this too much. This is technically fine so we can merge this as is if you want, but I just remembered that this is exactly the sort of scenario that `Run(Action)` was designed for!

Perksey (commented):
Just a quick one your thoughts would be appreciated on...

src/Windowing/Silk.NET.Windowing.Glfw/GlfwWindow.cs
otac0n: Can you please give me a test case for such an Update() event?  I am not 100% sure I follow and I want to make sure I cover your concern.

Perksey (approved):
Perfect, thank you!

PR #1990, opened 2024-03-20
Fix AVE in PtrToString on .NET 8 & remove unnecessary marshalling

### LINKED ISSUE #1989: Assimp GetMaterialString binding can cause AccessViolationException
# Summary
For certain overload of the Assimp.GetMaterialString, there is a code that copies managed string to global memory and passes the pointer to native method. However when it returns, it attempts to convert to string (`SilkMarshal.PtrToString`) which is useless since the resulting string is not use anywhere.

The aforementioned PtrToString (or rather `Utf8PtrToString` used for the used `LPUTF8Str` encoding) then can cause `AccessViolationException` because the way it is implemented is it scans for null-terminator in `Span<byte>(ptr, int.MaxValue)` which can use vectorization instructions and "look ahead". However when the pointer is near the edge of accessible memory, this can fail.

So technically there are two issues: unnecessary pointer to string conversion, and Access Violation risk when using `Utf8PtrToString`

# Steps to reproduce
- Platform: Desktop
- Framework Version: .NET 8
- Tested Nuget Versions: 2.13.0, 2.20.0

I wasn't able to create easy repro, but the bug was encountered after using Silk.NET.Assimp like this:
```cs
string GetMaterialName(Material* material)
{
    AssimpString materialName = new AssimpString();
    _assimp.GetMaterialString(material, Assimp.MaterialNameBase, 0, 0, ref materialName);
...
```

```
Fatal error. System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
    at System.SpanHelpers.NonPackedIndexOfValueType[[System.Byte, System.Private.CoreLib, Version=8.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[System.SpanHelpers+DontNegate`1[[System.Byte, System.Private.CoreLib, Version=8.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]], System.Private.CoreLib, Version=8.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]](Byte ByRef, Byte, Int32)
    at System.MemoryExtensions.IndexOf[[System.Byte, System.Private.CoreLib, Version=8.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]](System.Span`1<Byte>, Byte)
    at Silk.NET.Core.Native.SilkMarshal.Utf8PtrToString(IntPtr)
    at Silk.NET.Core.Native.SilkMarshal.PtrToString(IntPtr, Silk.NET.Core.Native.NativeStringEncoding)
    at Silk.NET.Assimp.Assimp.GetMaterialString(Silk.NET.Assimp.Material*, System.String, UInt32, UInt32, Silk.NET.Assimp.AssimpString ByRef)
```

# Comments
In the following image, the highlighted code shouldn't be necessary at all.
image

Perksey:
Closes #1989

### DIFF
--- a/src/Core/Silk.NET.Core/Native/SilkMarshal.cs
+++ b/src/Core/Silk.NET.Core/Native/SilkMarshal.cs
@@ -518,6 +518,10 @@ namespace Silk.NET.Core.Native
         /// <param name="ptr">The native string pointer.</param>
         /// <param name="encoding">The encoding.</param>
         /// <returns>The length of the string.</returns>
+        /// <remarks>
+        /// Note that this returns the length in characters. Namely, if a 16-bit character encoding is used, the actual
+        /// byte count will be double the return value from this function.
+        /// </remarks>
 #if NET6_0_OR_GREATER
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static unsafe nuint StringLength(
@@ -531,7 +535,7 @@ namespace Silk.NET.Core.Native
             );
 #else
         public static unsafe nuint StringLength(
-            nint ptr, 
+            nint ptr,
             NativeStringEncoding encoding = NativeStringEncoding.Ansi
         )
         {
@@ -540,7 +544,13 @@ namespace Silk.NET.Core.Native
                 return 0;
             }
             nuint ret;
-            for (ret = 0; ((byte*) ptr)![ret] != 0; ret++) { }
+            for (
+                ret = 0;
+                encoding == NativeStringEncoding.LPWStr
+                    ? ((char*)ptr)![ret] != 0
+                    : ((byte*)ptr)![ret] != 0;
+                ret++
+            ) { }
             return ret;
         }
 #endif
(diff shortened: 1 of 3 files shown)


HurricanKai (approved):
Seems reasonable. I'd recommend extensively testing a change like this, ref changes shouldn't be done lightly.
Especially test GLFW, GL, Vk.
Overall I'd expect this to be fine though.

src/Core/Silk.NET.Core/Native/SilkMarshal.cs
```diff
@@ -512,19 +512,43 @@ public static string[] MemoryToStringArray
...
-            var span = new Span<byte>((void*) ptr, int.MaxValue);
-            span = span.Slice(0, span.IndexOf(default(byte)));
-            if (span.Length == 0)
+            if (ptr == 0)
             {
-                return string.Empty;
+                return 0;
             }
+            nuint ret;
+            for (ret = 0; ((byte*) ptr)![ret] != 0; ret++) { }
```
MihaZupan, line 543: This would be broken for `NativeStringEncoding.LPWStr`, or is that never used < net6?

Perksey: Good point. This codepath specifically isn't hit today, but if we're claiming to support multiple encodings we should implement it properly. Resolved in latest commit.

PR #1991, opened 2024-03-20
Use sensible defaults for depth/stencil

### LINKED ISSUE #1863: "Tutorial 2.2 - Camera" don't work on macos
# Summary
"Tutorial 2.2 - Camera" works as expected on Windows, but it seems to have a depth detection problem on my Mac.

# Steps to reproduce
- Platform: Desktop
- Framework Version: .NET 8
- API: OpenGL
- API Version: OpenGL Version: 4.1 Metal - 88
1. Just run "Tutorial 2.2 - Camera" on Mac

# Comments
I am testing in an Apple M1 Pro
macOS: Sonoma 14.2.1

Screenshot 2023-12-27 at 12 16 49

Perksey:
Closes #1863

### DIFF
--- a/examples/CSharp/Directory.Build.props
+++ b/examples/CSharp/Directory.Build.props
@@ -1,7 +1,7 @@
 ﻿<Project>
     <PropertyGroup>
         <AssemblyName>Tutorial</AssemblyName>
-        <LangVersion>9</LangVersion>
+        <LangVersion>latest</LangVersion>
         <WarningsAsErrors>0618</WarningsAsErrors>
     </PropertyGroup>
 </Project>
--- a/src/Windowing/Silk.NET.Windowing.Common/Interfaces/IViewProperties.cs
+++ b/src/Windowing/Silk.NET.Windowing.Common/Interfaces/IViewProperties.cs
@@ -78,7 +78,8 @@ namespace Silk.NET.Windowing
         /// Preferred depth buffer bits of the window's framebuffer.
         /// </summary>
         /// <remarks>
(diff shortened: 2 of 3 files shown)


ThomasMiz:
Should we update the SDL backend too? [The current relevant code looks like this](https://github.com/dotnet/Silk.NET/blob/main/src/Windowing/Silk.NET.Windowing.Sdl/SdlView.cs):

```cs
            Sdl.GLSetAttribute(GLattr.DepthSize,
                    opts.PreferredDepthBufferBits is null || opts.PreferredDepthBufferBits == -1
                        ? 24 : opts.PreferredDepthBufferBits.Value);

            Sdl.GLSetAttribute(GLattr.StencilSize,
                    opts.PreferredStencilBufferBits is null || opts.PreferredStencilBufferBits == -1
                        ? 8 : opts.PreferredStencilBufferBits.Value);
```

So the only difference is that if one of these options is set to `null`, it uses 24/8 instead of the system default.

Everything looks good though

Perksey:
Yeah I agree, but I’m not sure that SDL has “system default” functionality. This PR was mostly to align it with SDL for the common case. I’ll double check to make sure.

Perksey:
Couldn’t find anything equivalent to GLFW’s don’t care… maybe not setting the attributes does that. I’ll give that a go.

Perksey:
@ThomasMiz Slightly adjusted the SDL logic, PTAL

ThomasMiz (approved):

PR #1992, opened 2024-03-20
Fire GLFW joystick events

### LINKED ISSUE #1862: SDL2 Doesn't Respect WindowOptions.Position Member
# Summary
It seems that SDL2 windows open at the same exact position every time. With the same exact code on GLFW I do not experience this issue.

# Steps to reproduce
- Platform: Windows
- Framework Version: .NET Core 8.0
- API: OpenGL (highest supported is 4.6)
```
<ItemGroup>
	<PackageReference Include="Silk.NET.Windowing" Version="2.20.0" />
	<PackageReference Include="Silk.NET.OpenGL" Version="2.20.0" />
	<PackageReference Include="Silk.NET.Input" Version="2.20.0" />
</ItemGroup>
```
```cs
public static void Main(string[] args)
{
	Window.PrioritizeSdl();
	
	options.Samples = config.Window.Samples;
	options.Size = new(800, 600);
	options.Position = new(1000, 400);
	options.Title = "Example";
	
	window = Window.Create(options);
	window.Load += OnLoad; // set up IInputContext, GL.GetApi(window)
	window.Render += OnRender; // glClear
	window.Update += OnUpdate; // empty
	window.Closing += OnClose; // key == Key.Escape window.Close()
	
	window.Run();
	window.Dispose();
}
```

### LINKED ISSUE #1889: Silk.NET.Input.Glfw.GlfwJoystick: Events not implemented
# Summary

When trying to use joysticks with GLFW, they do not fire events (ButtonDown, ButtonUp, AxisMoved, HatMoved), probably because there isn't any code that invokes those events. SDL, on the other hand, does: https://github.com/dotnet/Silk.NET/blob/main/src/Input/Silk.NET.Input.Sdl/SdlJoystick.cs#L55

# Steps to reproduce
- Platform: N/A
- Framework Version: N/A
- API: N/A
- API Version: N/A

1. Plug in any joystick
2. Run InputTest: https://github.com/dotnet/Silk.NET/tree/main/src/Lab/Experiments/InputTest
3. Attempt to use the joystick
4. Note that no events for the joystick are being logged to console
5. (optional) Run with `--sdl` and note that SDL does log to console

# Comments

Since I haven't written any C# before, I don't know what the right fix for this is. If you have any pointers for me, I might attempt it anyway.

Perksey:
Fixes #1889 
Closes #1862

### DIFF
--- a/src/Input/Silk.NET.Input.Glfw/GlfwJoystick.cs
+++ b/src/Input/Silk.NET.Input.Glfw/GlfwJoystick.cs
@@ -57,17 +57,29 @@ namespace Silk.NET.Input.Glfw
 
             for (var i = 0; i < btnCount; i++)
             {
-                ((Button[]) Buttons)[i] = new Button(ButtonName.Unknown, i, btn[i] == (int) InputAction.Press);
+                var thisBtn = new Button(ButtonName.Unknown, i, btn[i] == (int) InputAction.Press);
+                var shouldInvoke = ((Button[]) Buttons)[i].Pressed != thisBtn.Pressed;
+                ((Button[]) Buttons)[i] = thisBtn;
+                if (shouldInvoke)
+                {
+                    (thisBtn.Pressed ? ButtonDown : ButtonUp)?.Invoke(this, thisBtn);
+                }
             }
 
             for (var i = 0; i < axisCount; i++)
             {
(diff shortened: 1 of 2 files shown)


ThomasMiz (approved):
lgtm!

PR #2005, opened 2024-03-24
Fix VSync not being respected on window creation

### LINKED ISSUE #2000: VSync setting is not being respected when switching WindowState inside Update event.
# Summary
Switching `Window.WindowState` inside `Window.Update` event ignores `VSync` setting defaulting to false.
Switching window state inside `Window.Load` event works properly.
This issue can be omitted in a hacky way by adding `Window.VSync = Window.VSync` after switching `Window.WindowState`

# Steps to reproduce
- Platform: Desktop (Windows 11)
- Framework Version: .NET 8.0
- API: OpenGL
- API Version: OpenGL 3.3.0 Core Profile Context 24.2.1.240215

1. Initialize window using `WindowOptions` with `VSync` set to true
2. Assign Update event and inside the event change `WindowState` to Fullscreen.

This will result in fullscreen window that has vsync disabled, with `Window.VSync` still being set to true.
This issue seems to be present only when switching from any window state to fullscreen.

# Minimal code sample reproducing the issue
```cs
using Silk.NET.Maths;
using Silk.NET.Windowing;

options.Size = new Vector2D<int>(1280, 720);
options.Title = "Window Name";
options.VSync = true;

ulong frame = 0;

IWindow Window = Silk.NET.Windowing.Window.Create(options);
Window.Update += OnUpdate;

Window.Run();

void OnUpdate(double dt)
{
    frame++;
    if (frame == 100)
    {
        Window.WindowState = WindowState.Fullscreen;
        // trick below solves the issue
        //Window.VSync = Window.VSync;
    }
}
```

ThomasMiz:
# Summary of the PR
Fixes #2000.

I'm having issues compiling ATM so I couldn't test this yet, and I'm out of time to work on this for today. I'll get back to this soon enough

### DIFF
--- a/src/Windowing/Silk.NET.Windowing.Glfw/GlfwWindow.cs
+++ b/src/Windowing/Silk.NET.Windowing.Glfw/GlfwWindow.cs
@@ -115,7 +115,7 @@ namespace Silk.NET.Windowing.Glfw
             }
         }
 
-        protected override IGLContext? CoreGLContext => _glContext ??= new(_glfw, _glfwWindow, this);
+        protected override IGLContext CoreGLContext => _glContext ??= new(_glfw, _glfwWindow, this);
 
         protected override IVkSurface? CoreVkSurface
         {
@@ -433,6 +433,8 @@ namespace Silk.NET.Windowing.Glfw
             {
                 _glfw.MakeContextCurrent(_glfwWindow);
             }
+
+            CoreGLContext.SwapInterval(opts.VSync ? 1 : 0);
 
             GLFW.Glfw.ThrowExceptions();
         }


Perksey:
I think the cause may be the change in video mode i.e. when it goes full screen, the swap interval is reset. My impression was that we had logic somewhere in there already to set SwapInterval after creation (or at least I have a feeling that I wrote that at some point in the last 5 years, possibly in DoRender where we try to detect if we’ve done it already maybe)

ThomasMiz:
Yes, in [ViewImplementationBase.cs](https://github.com/dotnet/Silk.NET/blob/main/src/Windowing/Silk.NET.Windowing.Common/Internals/ViewImplementationBase.cs):

```cs
        public bool VSync
        {
            get => _optionsCache.VSync;
            set
            {
                _swapIntervalChanged = true;
                _optionsCache.VSync = value;
            }
        }
```

```cs
        public void DoRender()
        {
            [...]

            if ((delta >= _renderPeriod) || VSync)
            {
                [...]

                if (!IsContextControlDisabled && _swapIntervalChanged)
                {
                    GLContext?.SwapInterval(VSync ? 1 : 0);
                    _swapIntervalChanged = false;
                }

                [...]
            }
            
            [...]
        }
```

I was going to suggest we initialize  `_swapIntervalChanged` to true, but...
```cs
        // Ensure we keep SwapInterval up-to-date
        private bool _swapIntervalChanged = true;
```

Not sure what's going on here

Perksey:
Setting _swapIntervalChanged to true probably makes sense in CoreInitialize. I think we should also set that to true whenever we change the window state or monitor.

PR #2008, opened 2024-04-02
Parse Khronos metadata (primarily for SupportedApiProfileAttributes and GL/CL groups) 

Perksey:
This PR aims to add a lot of XML parsing logic, particularly with regards to the following items:
- **API Sets**: Which features/extensions symbols fall into. We'll use this later to make the SupportedApiProfileAttributes accurate for later consumption by the analyzer that we want to make.
- **Enum Groups**: Creating C# `enum`s from the spec where the associated Khronos header does not have C `enum`s doing this for us.

**_This PR does not change codegen yet, that will come in a subsequent PR_**


sources/Core/DSL/Constant`2.cs
```diff
@@ -0,0 +1,51 @@
...
+/// <summary>
+/// Represents a constant that optionally can be represented as an enum. This allows both constant integers and
+/// enumerated members to be used by the same parameter without a cast.
+/// </summary>
+/// <param name="Value">The underlying value.</param>
+/// <typeparam name="TUnderlying">The underlying value type.</typeparam>
+/// <typeparam name="TEnum1">
+/// The enum that can be used in place of a <typeparamref name="TUnderlying"/> constant.
+/// </typeparam>
+public readonly record struct Constant<TUnderlying, TEnum1>(TUnderlying Value)
```
Perksey, line 14: Need thoughts on what the name of this should be, as in theory we might want to use this for OpenGL handles as well e.g. `Handle<uint, Program>`. Maybe we could all this `Alt`?

sources/SilkTouch/Mods/MixKhronosData.cs
```diff
@@ -25,11 +32,78 @@ public partial class MixKhronosData(
...
+        /// </summary>
+        public required Configuration Configuration { get; init; }
+
+        /// <summary>
+        /// A mapping of enum native names to group names.
+        /// </summary>
+        /// <remarks>
+        /// This is OpenGL and OpenCL specific.
+        /// </remarks>
+        // TODO map OpenCL into these concepts
```
Perksey, line 62:
```suggestion
```

sources/SilkTouch/Mods/MixKhronosData.cs
```diff
@@ -25,11 +32,78 @@ public partial class MixKhronosData(
...
+        public Dictionary<string, HashSet<string>> EnumsToGroups { get; } = [];
+
+        /// <summary>
+        /// A mapping of group names to constant declarators.
+        /// </summary>
+        /// <remarks>
+        /// This is OpenGL and OpenCL specific.
+        /// </remarks>
+        // TODO map OpenCL into these concepts
```
Perksey, line 71:
```suggestion
```

sources/SilkTouch/Mods/MixKhronosData.cs
```diff
@@ -116,35 +196,869 @@ JsonSerializerOptions options
...
+            var ext in xml.Element("registry")?.Element("extensions")?.Elements("extension") ?? []
+        )
+        {
+            var name =
+                ext.Attribute("name")?.Value
+                ?? throw new InvalidOperationException("Extension has no name.");
+
+            // Note that we assume that if an extension is reported to be supported, then at least one of its dependency
+            // options is supported. The reason why we can use allowRequireAll is because this is codified in the
+            // SupportedApiProfileAttribute documentation - if any one of the reported implications has a +, then you
```
Perksey, line 966: This has just reminded me, don't forget to deal with the `,` with no `+` case

Perksey: Done

Perksey:
@curin This is now ready for review.

ThomasMiz:
Is there a reason why do the two "Constant" files have backticks in their names?

Perksey:
That's just my personal style of representing generics in file names (you can see this on main as well). We should decide on one as a team if it's not desirable.

curin (approved):

ThomasMiz:
> That's just my personal style of representing generics in file names (you can see this on main as well). We should decide on one as a team if it's not desirable.

I never noticed, what I do (and what I thought was the general standard) is to not include the generic in the file name

Perksey:
Yeah I noticed Curin uses an “OfT.cs” suffix, so we all have different styles here haha.

PR #2011, opened 2024-04-05
Reference native packages from managed SPIR-V packages.

### LINKED ISSUE #2007: Should Silk.NET.Shaderc depend on Silk.NET.Shaderc.Native?
I just happened to notice that [Silk.NET.Shaderc](https://www.nuget.org/packages/Silk.NET.Shaderc) does not pull in [Silk.NET.Shaderc.Native](https://www.nuget.org/packages/Silk.NET.Shaderc.Native) as a dependency.

Obviously I can easily reference the latter myself, but it seems to usually be the case that the managed Silk.NET packages pull in their native counterparts as dependencies, so I wonder if this is intentional or just an oversight?

alexrp:
Closes #2007.

### DIFF
--- a/src/SPIRV/Silk.NET.SPIRV.Cross/Silk.NET.SPIRV.Cross.csproj
+++ b/src/SPIRV/Silk.NET.SPIRV.Cross/Silk.NET.SPIRV.Cross.csproj
@@ -7,6 +7,10 @@
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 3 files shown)


Beyley (approved):
yeaƐ�▒!��C���|ڈAC*�yf]�?�LK�[�᜘�ܑ�R|�D�fM%#�(��o������k��
                                                                                                     ������&)�>��&E���tP9��x�=�86������,"ܚH��F| ��#����
���A6K�O�s����õ��S�w﷣\

PR #2017, opened 2024-04-06
Make Silk.NET buildable on Linux by skipping iOS and Mac Catalyst stuff.

alexrp:
Also update build instructions.

Closes #2014.

PR #2018, opened 2024-04-06
Don't run native binary workflows in forks.

alexrp:


Perksey (approved):
I think this looks fine, I had a look into whether we could do something in common.props and it looks like it’ll be more trouble than it’s worth

Perksey (approved):
Ignore the above message, that was meant for #2017

PR #2022, opened 2024-04-08
Fix MoltenVK native build failures due to file overwrite.

alexrp:
For #1993.

cc @Perksey @Beyley


alexrp:
Note: CI failure is unrelated and already happening on #1993.

Perksey:
Forgoing CI as it's going to a feature branch. Would page Beyley but I don't see them having any objections.

PR #2030, opened 2024-04-12
A proposal for Bindings Owners to allow more bindings into Silk.NET

Perksey:
Just read the proposal, I promise this isn't scope creep.


Perksey:
This is ready for a pre-WG review.

Beyley (commented):
in my initial glance all is pona tawa mi

alexrp (approved):

aquagoose (approved):

curin (approved):

documentation/proposals/Proposal - Project Governance, Bindings Expansion and Ownership.md
```diff
@@ -0,0 +1,142 @@
...
+
+A "core binding" **shall** be defined as follows:
+1. The API being bound to **must** be two or more of the following as determined by the maintainers team:
+    1. An API existing in the multimedia and/or compute space, or aiding in the creation of applications making use of such APIs.
+    2. Used as a dependency by a Working Group approved HLU, Silk.NET core functionality, or a Silk.NET core binding.
+    3. An API developed by a multi-vendor standards team/body, or an API with notable industry presence otherwise implemented by multiple vendors and/or developed in collaboration with those vendors. 
+2. The C API being bound to **must** have open-source headers that do not preclude the automated transformation to C# bindings without undue licensing provisions inflicted on users of such transformations.
+
+All Khronos and DirectX APIs meet both requirements 1.1 and 1.3. OpenAL also meets this requirement based on its historical development. GLFW and SDL both meet these rqeuirements today given that the Windowing HLU requires both of them in 2.X, their continued status as "core binding" is dependent on what the final implementation of the Windowing 3.0 proposal [5] uses.
```
ThomasMiz, line 103:
```suggestion
All Khronos and DirectX APIs meet both requirements 1.1 and 1.3. OpenAL also meets this requirement based on its historical development. GLFW and SDL both meet these requirements today given that the Windowing HLU requires both of them in 2.X, their continued status as "core binding" is dependent on what the final implementation of the Windowing 3.0 proposal [5] uses. 
```

ThomasMiz (approved):

documentation/proposals/Proposal - Project Governance, Bindings Expansion and Ownership.md
```diff
@@ -0,0 +1,142 @@
...
+
+1. The C API being bound to **must** have open-source headers that do not preclude the automated transformation to C# bindings without undue licensing provisions inflicted on users of such transformations. 
+2. The C API being bound to **must** have at least one open-source implementation that is usable for the vast majority of use cases of that binding.
+    - The Silk.NET team has in the past allowed implementations with copyleft restrictions (namely with regards to static linking a la LGPL) that don't apply on desktop but would apply on mobile.
+3. The C API **must** be a recognised industry standard multimedia or compute API, aid in the creation of applications making use of such APIs, and/or used to implement a Silk.NET High Level Utility.
+    - The "recognised industry standard multimedia or compute API" covers all Khronos APIs, OpenAL, and the 2.X portfolio of DirectX bindings.
+    - The "aid in the creation of applications making use of such APIs" covers Assimp and Shaderc.
+    - The "used to implement a Silk.NET High Level Utility" covers GLFW and SDL.
```
joskuijpers, line 44: I understand why you write this, but seeing that Silk SDL bindings are the best ones available, is it there for the HLU or primarily for Standard Multimedia API?

Perksey: Basically, we will maintain our SDL bindings to the best of our ability so long as we're using them in our windowing/input HLU APIs. This does not preclude the bindings going beyond what is needed for the HLUs e.g. binding to the rendering APIs for instance, but it sets a tone for how we approach their maintenance.

This is no different to how we approach the SDL bindings in 2.X today.

joskuijpers: So if the HLU, which some might not use (preferring their own), switches to direct Mac/windows/X bindings, Silk SDL could be dropped according to the new proposal.

Or rather, it would have to become an Additional Binding.

According to https://www.nuget.org/profiles/Silk.NET, the most popular bindings are OGL/Vulkan, then LWGL, and then SDL. The HLU is actually less popular than SDL itself. Do I read that correctly?

Perksey: I think you're correct that we should likely make it so core bindings must remain core for the entirety of the major version's lifecycle.

Perksey: I've encompassed this in the proposal.

Perksey:
- Approved subject to these comments.
- Version suffixes should be used for betas, specifically for bindings that are not quite stable but still meet the requirements. Those suffixes apply only to those additional binding nuget packages.

PR #2115, opened 2024-04-15
This is the April 2024 Update (v2.21.0)

### LINKED ISSUE #1500: Keyboard events stops working after few Window Create/Dispose calls.
# Summary
It seems that in a scenario where we create and dispose a Window several times, at some point the keyboard event trigger stop working.
I pointed this out when transferring several OpenGL examples into single project, where individual examples runs from the console menu.
I'm not entirely sure if this is a bug in Windowing/Input, or maybe I'm doing something wrong ?

# Steps to reproduce
- Platform: Desktop
- Framework Version: .NET 7
- API: N/A

Below is an extracted code that shows the problem (console app).

- When I use the KeyDown event to close the window, after few calls this stops working.
- But when checking IsKeyPressed, closing works every time.

I noticed too, that when calling the ImGui Demo that way, keyboard events stop being passed to the controller, while mouse events continue to work.

# Example

```cs
using Silk.NET.Windowing;
using Silk.NET.Input;

while (true)
{
    Console.WriteLine("Press any key to open window");
    Console.ReadKey(true);
    OpenWindow_WithEvent(); // <-- there is a problem
    // OpenWindow_WithoutEvent(); // <-- this is fine
    while (Console.KeyAvailable) Console.ReadKey(true);
    Console.WriteLine($"Window closed\n");
}

// Escape KeyDown event stops working after few calls.
static void OpenWindow_WithEvent()
{
    var window = Window.Create(WindowOptions.Default);
    window.Initialize();
    var input = window.CreateInput();
    Console.WriteLine("Window opened, press Escape key to close");
    input.Keyboards[0].KeyDown += (IKeyboard a1, Key a2, int a3)
       => { Console.WriteLine(a2); if (a2 == Key.Escape) window.Close(); };
    window.Run();
    input.Dispose();
    window.Dispose();
}

// Escape check by IsKeyPressed works every time.
static void OpenWindow_WithoutEvent()
{
    var window = Window.Create(WindowOptions.Default);
    window.Initialize();
    var input = window.CreateInput();
    Console.WriteLine("Window opened, press Escape key to close");
    window.Run(() => {
        if (input.Keyboards[0].IsKeyPressed(Key.Escape)) window.Close();
        window.DoEvents();
        window.SwapBuffers();
    });
    input.Dispose();
    window.Dispose();
}
```

### LINKED ISSUE #2000: VSync setting is not being respected when switching WindowState inside Update event.

### LINKED ISSUE #2010: Extensions do not work on full trimming (audit Activator usage)
Easy issue. May need some linking attributes littered some places. Reported in Discord.

### LINKED ISSUE #2026: Input related callbacks in GLFW cause an AccessViolationException
# Summary
When setting any input related callbacks in GLFW, PollEvents causes an AccessViolationException

# Steps to reproduce
- Platform: Desktop
- Framework Version: .NET 4.7.2
- API: OpenGL
- API Version: Legacy/OpenGL 1.1 (I mostly use Silk.NET to port legacy games)
1. Properly initialize a GLFW context and a window
2. Use either SetKeyCallback, SetMouseButtonCallback or SetCursorPosCallback (probably others too)
3. Call PollEvents

# Comments
Oddly enough, SetWindowSizeCallback and other non-input related callbacks do not cause an access violation

```
// Leaving on this callback is fine
glfw.SetWindowSizeCallback(window, (WindowHandle* window, int width, int height) => 
{
	// Code is executed
	// ...
});
// But uncommenting this line, causes an AccessViolationException when polling the events
glfw.SetKeyCallback(window, (WindowHandle * window, Keys key, 
    int scanCode, InputAction action, KeyModifiers mods) =>
{
	// Code is not executed, game crashes
	// ...
});
```

Perksey:
These are the final bugfixes and patch notes for 2.21.

Fixes #2026 
Fixes #2000 
Closes #2005 
Fixes #2010 
Fixes #1500


generator.json
```diff
@@ -77,6 +77,44 @@
...
+                "SecondaryColorPointerEXT": ["pointer"],
+                "SecondaryColorPointerListIBM": ["pointer"],
+                "VertexAttribPointer": ["pointer"],
+                "VertexAttribPointerARB": ["pointer"],
+                "VertexAttribPointerNV": ["pointer"],
+                "VertexAttribIPointer": ["pointer"],
+                "VertexAttribIPointerEXT": ["pointer"],
+                "TangentPointerEXT": ["pointer"],
+                "BinormalPointerEXT": ["pointer"]
+            },
```
Perksey, line 117: Add to the end of Section 2.10 (Vertex Array Objects):

    "The following commands:

        void VertexArrayVertexOffsetEXT(uint vaobj, uint buffer,
                                        int size, enum type, sizei stride,
                                        intptr offset);
        void VertexArrayColorOffsetEXT(uint vaobj, uint buffer,
                                       int size, enum type, sizei stride,
                                       intptr offset);
        void VertexArrayEdgeFlagOffsetEXT(uint vaobj, uint buffer,
                                          sizei stride, intptr offset);
        void VertexArrayIndexOffsetEXT(uint vaobj, uint buffer,
                                       enum type, sizei stride,
                                       intptr offset);
        void VertexArrayNormalOffsetEXT(uint vaobj, uint buffer,
                                        enum type, sizei stride,
                                        intptr offset);
        void VertexArrayTexCoordOffsetEXT(uint vaobj, uint buffer,
                                          int size, enum type, sizei stride,
                                          intptr offset);
        void VertexArrayMultiTexCoordOffsetEXT(uint vaobj, uint buffer,
                                               enum texunit,
                                               int size, enum type, sizei stride,
                                               intptr offset);
        void VertexArrayFogCoordOffsetEXT(uint vaobj, uint buffer,
                                          enum type, sizei stride,
                                          intptr offset);
        void VertexArraySecondaryColorOffsetEXT(uint vaobj, uint buffer,
                                                int size, enum type,
                                                sizei stride, intptr offset);
        void VertexArrayVertexAttribOffsetEXT(uint vaobj, uint buffer,
                                              uint index, int size,
                                              enum type, boolean normalized,
                                              sizei stride, intptr offset);
        void VertexArrayVertexAttribIOffsetEXT(uint vaobj, uint buffer,
                                               uint index, int size,
                                               enum type, sizei stride,
                                               intptr offset);

    operate identically to the corresponding command where "VertexArray"
    is deleted from the name and "Offset" is changed to "Pointer" (and
    extension suffixes are dropped or updated appropriately) except,
    1) rather than updating the currently in use vertex array object,
    these "VertexArray" commands update the vertex array object named
    by the initial vaobj parameter, and 2) the corresponding command
    operates as if the ARRAY_BUFFER buffer binding is bound to the
    buffer named by the buffer parameter.  The remaining parameters
    following the initial vaobj parameter for the listed "VertexArray"
    commands match the parameters for the corresponding non-"VertexArray"
    vertex array specification command and are interpreted as they
    are for the non-"VertexArray" vertex array specification command.
    The one exception is that final intptr offset parameter of the
    "VertexArray" command is cast to the final pointer parameter of the
    non-"VertexArray" command (intptr is guaranteed to be large enough
    to store any pointer value).

Perksey:
I think I've got all the low-hanging fruits for the next update, will finalise the patch notes once the natives work has concluded.

Perksey:
This is ready for review.

Beyley (approved):
j

build/props/common.props
```diff
@@ -45,7 +45,7 @@
 
             After this update:
             - Apple Silicon support is complete except for missing ANGLE binaries.
-            - Arm Linux support is complete except for missing DXVK, ANGLE, Vkd3d, and Vulkan Loader (missing on 32-bit only) binaries.
+            - Arm Linux support is complete except for missing DXVK, ANGLE, and Vulkan Loader (missing on 32-bit only) binaries.
```
alexrp, line 48: Note that Vulkan Loader doesn't actually fully support Arm32. So it's not just a shortcoming on Silk's end.

build/props/common.props
```diff
@@ -17,9 +17,9 @@
...
             - Added a distinction between "unset" (-1) and "default" (null) for Windowing depth/stencil buffer bits, defaulting to 24/8 on all platforms unless overridden.
-            - Added support for 64-bit Arm Linux for Assimp, GLFW (+ Windowing &amp; Input), OpenAL Soft, SDL, Vulkan Loader, SwiftShader, and WGPU. (thanks @alexrp)
+            - Added support for 64-bit Arm Linux for Assimp, GLFW (+ Windowing &amp; Input), OpenAL Soft, SDL, Vkd3d, Vulkan Loader, SwiftShader, and WGPU. (thanks @alexrp)
             - Added support for Apple Silicon Macs for OpenAL Soft, Vulkan Loader, and SwiftShader. (thanks @alexrp)
-            - Added support for 32-bit Arm Linux for Assimp, GLFW, OpenAL Soft, SDL, Shaderc, SPIR-V Cross, SPIR-V Reflect, SwiftShader, and WGPU. (thanks @alexrp)
+            - Added support for 32-bit Arm Linux for Assimp, GLFW, OpenAL Soft, SDL, Shaderc, SPIR-V Cross, SPIR-V Reflect, Vkd3d, SwiftShader, and WGPU. (thanks @alexrp)
             - Added support for 64-bit Windows on Arm for Assimp, OpenAL Soft, VulkanLoader, and SwiftShader. (thanks @alexrp)
             - Added native library support for glibc versions as old as 2.17 on Linux platforms.
```
alexrp, line 24: The situation is a bit more nuanced:

* SPIR-V Cross, SPIR-V Reflect, and Shaderc are compiled for glibc 2.17 (built with Zig w/o linking to system libraries).
* SDL is compiled for glibc 2.27 (built on Ubuntu 18.04).
* DXVK is compiled for whatever glibc version Arch Linux is currently on (so 2.39 at the moment, AFAICT).
    * This is actually pretty bad now that I think about it...
* Everything else is compiled for glibc 2.35 (built on Ubuntu 22.04).

Perksey: Thanks!
```suggestion
            - Added limited native library support (SPIR-V Tools and Shaderc) for glibc versions as old as 2.17 on Linux platforms.
```
We've never advertised glibc versions for the other libs so these will effectively be unchanged compared to the previous release.

build/props/common.props
```diff
@@ -45,7 +45,7 @@
 
             After this update:
             - Apple Silicon support is complete except for missing ANGLE binaries.
-            - Arm Linux support is complete except for missing DXVK, ANGLE, Vkd3d, and Vulkan Loader (missing on 32-bit only) binaries.
+            - Arm Linux support is complete except for missing DXVK, ANGLE, and Vulkan Loader (missing on 32-bit only) binaries.
             - 64-bit Windows on Arm support is complete except for missing DXVK and ANGLE binaries.
             - 32-bit Windows on Arm remains unsupported except for DirectStorage.
```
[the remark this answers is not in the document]
Perksey: We still support UWP by virtue of using .NET Standard 2.0 which supports Arm32.

build/props/common.props
Perksey: Yep, was careful with the wording here i.e. saying "missing binaries" instead of "unsupported"

PR #2156, opened 2024-04-24
GlfwNativeWindow: Allow wayland window ptr to be read when GLFW is compiled with both X11 and Wayland

Beyley:
# Summary of the PR
Currently if wayland is used we say its X11 with no window or display ptr


Perksey (approved):
ok

uwx:
Nitpick: Methinks you should actually check the error code for [GLFW_PLATFORM_UNAVAILABLE](https://www.glfw.org/docs/latest/group__errors.html#ga3608c6c29ab7a72f3bf019f4c3a2563d) instead of just relying on the only possible errors being a sign that X11 is not to be used.

PR #2161, opened 2024-04-26
Fix build on macOS runner which is apparently arm now

Perksey:


Perksey:
@dotnet/silk-dotnet anyone can review this one

Perksey:
The new job is failing because we weren't building the whole sln before, fixed on my branch :/ recommend merge with this failing for now

ThomasMiz (approved):

PR #2180, opened 2024-05-14
Update Assimp to 5.4.1 and build with Draco support.

### LINKED ISSUE #2025: Build Assimp with Draco support
Draco is a form of texture compression for formats like glTF.

To do this, we'd like to avoid having to ship an extra shared library in the `.Native` NuGet package, and instead link Draco statically into the Assimp shared library. I've submitted an upstream PR to enable us to do this: https://github.com/assimp/assimp/pull/5535

alexrp:
Closes #2025.


alexrp:
For some reason, on Windows only, Assimp ends up linking dynamically to `draco.dll`. I'll need to figure out why that is.

alexrp:
Closing this until I can get around to investigating why that linking issue happens on Windows (and seemingly only in CI 😔).

PR #2199, opened 2024-05-29
[2.X] Add a helper for custom IO in Assimp

Perksey:
# Summary of the PR
Adds a CustomFileIO class which allows a Stream factory to be used in the context of a FileIO implementation in Assimp.

https://discord.com/channels/521092042781229087/607634593201520651/1245146529468121138

# Further Comments


Perksey:
This is ready for review.

src/Assimp/Silk.NET.Assimp.Tests/GlobalUsings.cs
```diff
@@ -0,0 +1 @@
+global using Xunit;
```
alexrp, line 1: Just noting that if you would like to avoid having files such as this one, you can put this in the project file:

```xml
    <ItemGroup>
        <Using Include="Xunit" />
    </ItemGroup>
```

Perksey: Good point, Rider generated this as part of the template. Will fix.

src/Assimp/Silk.NET.Assimp/CustomFileIO.cs
```diff
@@ -0,0 +1,198 @@
...
+            ? throw new InvalidOperationException("Invalid UserData for File.")
+            : (nuint) s.Position;
+    
+    [UnmanagedCallersOnly(CallConvs = new[] { typeof(CallConvCdecl) })]
+    private static void FileFlush(File* file)
+    {
+        if (GCHandle.FromIntPtr((nint) file->UserData).Target is not Stream s)
+        {
+            throw new InvalidOperationException("Invalid UserData for File.");
+        }
```
Beyley, line 178: This function can probably be wrapped into a `Stream s = GetStreamFromUserData(file->UserData);`, which would de-duplicate a lot of code here

Beyley (changes requested):

Beyley (approved):

PR #2211, opened 2024-06-03
Build SDL2 for Android in CI and productise iOS support

### LINKED ISSUE #2210: Android Bluetooth permission question
I am assuming pull #1372 is in version 2.20.0 and 2.21.0 (the two latest versions of Silk) and contained the SDL changes which defaults to not enabling checking the bluetooth permission.
According to SDL the bluetooth permission check is ignored by default on Android (the hint SDL_HINT_JOYSTICK_HIDAPI_STEAM)
https://github.com/libsdl-org/SDL/commit/66058bbbd5083acfb0ff88c49faac76d725a711f

However, when I run the android app (just the Silk Android sample project in the repo) I am noticing the debug log shows
> [hidapi] Initializing Bluetooth

which seems to imply the hint is being set in Silk?
Is there a way to forcibly unset this hint value?

Perksey:
Closes #2210


Perksey:
OH NO THEY'RE ALL BUILDING

Perksey:
@Beyley this is ready for review, will test tomorrow maybe

build/nuke/Native/SilkDroid.cs
```diff
@@ -36,6 +36,12 @@ string AndroidHome
                 return AndroidHomeValue;
             }
 
+            if ((Environment.GetEnvironmentVariable("ANDROID_HOME") ?? Environment.GetEnvironmentVariable("ANDROID_SDK_ROOT")) is {} sdk)
```
Beyley, line 39: for reasons i cannot describe this is the worst thing ive ever seen but i cant think of a better way of writing this line of code in C#

src/Input/Silk.NET.Input/Silk.NET.Input.csproj
```diff
@@ -1,14 +1,25 @@
...
+    <ItemGroup Condition="'$(TargetFramework)' != 'net6.0-android' And '$(TargetFramework)' != 'net6.0-ios' And '$(TargetFramework)' != 'net6.0-maccatalyst' And '$(TargetFramework)' != 'net6.0'">
+        <ProjectReference Include="..\Silk.NET.Input.Glfw\Silk.NET.Input.Glfw.csproj" />
+        <ProjectReference Include="..\Silk.NET.Input.Sdl\Silk.NET.Input.Sdl.csproj" />
+    </ItemGroup>
+
+    <ItemGroup Condition="'$(TargetFramework)' == 'net6.0-android' Or '$(TargetFramework)' == 'net6.0-ios' Or '$(TargetFramework)' == 'net6.0-maccatalyst'">
+        <ProjectReference Include="..\Silk.NET.Input.Sdl\Silk.NET.Input.Sdl.csproj" />
+    </ItemGroup>
+    
+    <ItemGroup Condition="'$(TargetFramework)' == 'net6.0'">
```
Beyley, line 21: won't this mean SDL input wont be pulled on net6?

Beyley (commented):
one small thing mostly

src/Input/Silk.NET.Input/Silk.NET.Input.csproj
Perksey: Correct. This is as per Windowing as well. Before .NET 6 we can’t tell where you’re deploying, so we pull in both, whereas after we can tell if you’re using a platform we prefer SDL for from the TFM so we only pull in GLFW if we don’t know you’re using a mobile platform.

Beyley (approved):

Beyley (approved):

PR #2223, opened 2024-06-11
Update SDL2 binaries

silkdotnet:


Perksey (approved):

tweggen:
@Perksey 

I believe I might be able to fix this, I did a similar thing on a "practice branch" in my private fork, that is including a more recent SDL library, adapting the java files inside SilkDroid (e.g. to fix that Android API 34 requirement of RECEIVERS_EXPORTED as also described in #2210), adapting the files that describe what should be and what should be omitted from the public API.
I comitted that back into my Silk.NET fork, to have my github actions build me an updated set of nuget packages pushed to my private github package stream.
So I'd be happy to give it a try and contribute my knowledge back to the 2.x branch.
I would be perfectly fine with creating a pull request for my work, however, I believe, this one already is pretty similar.
An app containing the build result also already has passed google's checks for the playstore.

Perksey:
This is great news! Yeah we'd happily take this as a contribution. The main thing left to do on my PR was to make it add the Org.Libsdl public APIs to the public API file and retest. If you've figured a way to do that automatically then I'll happily incorporate that.

This is mainly because we try to ensure our binaries are traceable to CI so that there's less likelihood of tampering by an individual contributor.

PR #2278, opened 2024-09-06
[3.0] Windowing Rewrite

Perksey:
# Summary of the PR
A work-in-progress rewrite and design for windowing in 3.0.

https://discord.com/channels/521092042781229087/587346162802229298/1281715720186626079
https://github.com/dotnet/Silk.NET/blob/main/documentation/proposals/Proposal%20-%20Windowing%203.0.md

# Further Comments
None


sources/Windowing/Windowing/Implementations/SDL3/SdlSurfaceComponents.OpenGL.cs
```diff
@@ -0,0 +1,431 @@
...
+            }
+
+            // We need to create a dummy window if we haven't already (InitializePlatform does this, but it's only made
+            // available to the first Surface).
+            var props = CreateDummyWindowProps();
+            Handle = Sdl.CreateWindowWithProperties(props);
+            Sdl.DestroyProperties(props);
+            if (Handle == nullptr)
+            {
+                Sdl.ThrowError();
```
Perksey, line 234: This needs some thought

sources/Windowing/Windowing/Implementations/SDL3/SdlSurfaceComponents.OpenGL.cs
```diff
@@ -0,0 +1,431 @@
...
+    }
+
+    private void InitializeOpenGL(uint props)
+    {
+        if (Profile == OpenGLContextProfile.None)
+        {
+            return;
+        }
+
+        AddWindowCreateFlags(props, Sdl.WindowOpengl);
```
Perksey, line 420: What does this mean for the dummy window? Will IsSupported work as expected?

Perksey (commented):
Self review comments that need more thought

Perksey:
@curin Ready for review. Obviously see above for things that I know I want to look at. Will fix before merge.

Perksey:
Will update to SDL 3.2 after merge

sources/Windowing/Windowing/Implementations/SDL3/SdlSurfaceComponents.OpenGL.cs
Perksey: The dummy window has now changed to be retrieved through GetDummyWindow, specifying the flags that the dummy window should have and shouldn't have. If these criteria are already met, the Handle property is returned as-is. Otherwise, the dummy window is torn down if it exists and (re)created with the proper flags.

sources/Windowing/Windowing/Implementations/SDL3/SdlSurfaceComponents.OpenGL.cs
Perksey: This is now just a `return false` as a dummy window failing to be created with the OpenGL flags indicates non-support.

Perksey:
Last self-review comments have now been addressed.

sources/Core/Core/Abstractions/IGLContext.cs
```diff
@@ -5,34 +5,29 @@ namespace Silk.NET.Core;
...
 /// Represents an OpenGL context.
 /// </summary>
-public interface IGLContext : INativeContext, IDisposable
+public interface IGLContext : INativeContext
 {
     /// <summary>
     /// Whether the context is current on this thread.
     /// </summary>
-    bool IsCurrent { get; }
+    bool IsCurrent { get; set; }
```
curin, line 13: so can we make this context current on this thread?

curin: actually better question given that there was a makecurrent previously. What happens if I try and set this to false?

sources/Core/Core/Abstractions/INativeWindow.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    /// <summary>
+    /// Attempts to obtain native platform information with type <typeparamref name="TPlatformInfo" />.
+    /// </summary>
+    /// <param name="info">
+    /// The platform-specific information, or <c>default</c> if the platform-specific information is not available for
+    /// this platform.
+    /// </param>
+    /// <returns>True if <paramref name="info" /> contains the platform-specific information, false otherwise.</returns>
+    bool TryGetPlatformInfo<TPlatformInfo>([NotNullWhen(true)] out TPlatformInfo? info);
```
curin, line 21: do we want to have some sort of type constraint here to avoid passing in non-platform into structs here?

sources/Core/Core/Abstractions/IGLContext.cs
Perksey: IsCurrent = true is MakeCurrent. The context will be made current on the calling thread.
IsCurrent = false is ClearCurrent (or Clear, don't remember what this was called). The context will be cleared from the calling thread.

curin (approved):

PR #2284, opened 2024-09-22
ImGui memory leak fix

rbertels:
# Summary of the PR
The `ImGuiController::UpdateImGuiImput`  function creates a `MouseState` but never disposes it. 
This causes the `MemoryPool` to keep renting new buffers in the constructor, resulting in a small memory leak.


rbertels:
As per the contribution guide tagging @Perksey for a review.

I'm not a frequent OS contributor so if I skipped a step or did something incorrectly please let me know!

Perksey (approved):
Great spot, thank you!

PR #2245, opened 2024-07-11
Adding SyntaxContext

curin:
# Summary of the PR
Adds the SyntaxContext object to better interact with the whole syntax tree. Reworks the mods system to use this instead of the GeneratedSyntax object, which is instead used as part of the SyntaxContext construction.

PR #2234, opened 2024-06-26
Adding OpaqueCOM structs

curin:
# Summary of the PR
This PR will include the following:

-Opaque COM struct mod
 + Modifies COM interface structs to act as ComPtrs behind the scenes

PR #2207, opened 2024-05-31
[3.0] SDL Bindings

Perksey:
This is the initial set of SDL3 bindings. They will likely be modified in further PRs.

## Dependencies

None.

## Dependents

- #2296 
    - #2288 
        - #2234 
        - #2278

PR #2288, opened 2024-10-02
[3.0] Refactor SilkTouch to Use Roslyn Workspaces + more

Perksey:
# Summary of the PR
Refactors SilkTouch to make it fit for the future.

Links go here.

# Further Comments

## ClangScraper is now a mod in its own right

It was too special and I wanted to ensure any major refactor made the core of SilkTouch nothing but a mod execution and documents pipeline, rather than being magical and tied to ClangSharp intrinsically.

## Use Roslyn Workspaces

This is so we can make use of the `Renamer` which uses the semantic model to determine usages of the symbol we're renaming, rather than having to work all that context out ourselves. This is _much_ slower, but more correct and fixes issues encountered in #2234 wherein the renames didn't propagate through `thing->a->b` for instance

## Repository structure

As discussed in team talk.

## Dependencies

- #2296 
    - #2207 

## Dependents

- #2234 
- #2278 

## Supersedes

- Closes #2245


Perksey:
This has been rebased and to the best of my knowledge and belief it's ready for review @curin. It's quite hard to review though, I'd focus on the SilkTouch changes. Most of the bindings changes are fiddling with usings or member ordering, which really I would like to somehow make deterministic so it doesn't pollute PRs like it has here.

curin (approved):

PR #2290, opened 2024-10-05
Add CI builds (MacOS only) for ANGLE

ramezgerges:
# Summary of the PR
This PR adjusts the `Angle.cs` script to optimize disk space usage and also build fat binaries for MacOS (to work on both arm64 and x64 macs). For now, only MacOS builds are enabled.

# Related issues, Discord discussions, or proposals
Links go here.
https://github.com/dotnet/Silk.NET/issues/2024
# Further Comments

PR #2293, opened 2024-10-05
Use architecture-agnostic RID for ANGLE OSX builds

Perksey:
Follow on from #2290 given that we're using universal binaries.

### DIFF
--- a/build/nuke/Native/Angle.cs
+++ b/build/nuke/Native/Angle.cs
@@ -101,25 +101,21 @@ partial class Build {
                         // create universal mac binaries
                         var arm64NativeRuntimes = runtimes / "osx-arm64" / "native";
                         var x64NativeRuntimes = runtimes / "osx-x64" / "native";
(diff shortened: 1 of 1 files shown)


build/nuke/Native/Angle.cs
```diff
@@ -101,25 +101,21 @@ partial class Build {
                         // create universal mac binaries
                         var arm64NativeRuntimes = runtimes / "osx-arm64" / "native";
```
ramezgerges, line 102: We can also drop these variables I guess (and the associated EnsureCleanDirectory calls)

Perksey: Yeah... to be honest this was literally just because I couldn't be bothered to delete the files using GitHub UI haha. Will remove now.

ramezgerges (approved):

Perksey:
Just going to merge this given 2.X has no maintainers to approve atm.

PR #2306, opened 2024-10-22
Update SDL2 binaries

silkdotnet:


Perksey:
Opting to stick with what was approved for #1438

PR #2315, opened 2024-10-25
Use Apple M2 Pro in CI

Perksey:
Title says all. One line change.


Perksey:
I'm not sure this has been rolled out yet, should be some point in November. Moving to Draft.

Issue #2240, opened 2024-07-02
Silk.Net 2.21 window is stuck after minimizing


Arugin:
# Summary
Window is stuck right after minimizing. Everything works as expected in 2.20

# Steps to reproduce
- Platform: Windows
- Framework Version: .NET 7
- API: Vulkan

1. Get the sample https://github.com/dotnet/Silk.NET/tree/main/src/Lab/Experiments/ImGuiVulkan
2. Update to SIlk.Net 2.21
3. Run the sample
4. Minimize the window
5. Now it stuck and won't expand back

# Comments
If you set the version of Silk.Net and all other libraries to 2.20 everything works as expected.

Probably releated to recent changes of resize?

qian-o:
I have encountered this problem before, where I got stuck in an infinite loop while refactoring the swap chain. This may be related to thread scheduling.

Arugin:
I am not sure about thread scheduling, this issue is reproduced on a simple vulkan imgui sample which is single threaded.

Perksey:
Is there a minimal repro for this?

Arugin:
Yes, as I mentioned in the issue description, it reproduces on Imgui Vulkan sample from this repo: https://github.com/dotnet/Silk.NET/tree/main/src/Lab/Experiments/ImGuiVulkan

Arugin:
@Perksey hey, did you reproduce this issue? I can prepare reproducable project, if this will help.

Perksey:
My Windows PC isn't built atm, I can give it a go soon.

Perksey:
Reproduced, will investigate now.

Perksey:
I've identified this to be specific to the way our application is written (namely, calling `DoEvents` within an event callback).

Will investigate why we originally added this safeguard.

Perksey:
Ok looking at this, this logic was added for the `Refresh` callback (specifically, it calls the game loop which will call DoEvents which will call the game loop on refresh etc etc)

I'm not quite sure how I want to tackle this. I do want to make it so we can't infinitely recurse, but at the same time it's causing issues and really is just a safeguard. I think I'll just reduce this to just the refresh callback specifically (as that's the most likely cause of recursion, anything else and the user should deal with the consequences of their actions, if any).

Personally I'd love to make this sort of recursion an exception, but this issue is evidence that we have actual code out in the wild using this pattern and therefore that would be irresponsible.

References:
[...]

PR #2184, opened 2024-05-22
This is the Silk.NET Mobile Update (v2.22.0)

Beyley:
# Summary of the PR
the j

PR #2322, opened 2024-10-31
Regenerate bindings as of 31/10/2024

silkdotnet:


Perksey:
Will wait to see what CI says. If build passes, I'll merge into #2184 and review the whole PR (which is mostly Beyley's work tbh) there as that's the main PR. I might even turn that into a "this is" PR for the release as I don't see us doing anything further beyond this. _Maybe_ #2240

Issue #2298, opened 2024-10-09
Override ToString() method in Vulkan object structs


tonisimakov99:
Override ToString() method in Vulkan object structs
```
[NativeName("Name", "VkSemaphore")]
public struct Semaphore
{
    [NativeName("Type", "")]
    [NativeName("Type.Name", "")]
    [NativeName("Name", "")]
    public ulong Handle;

    public Semaphore(ulong? handle = null)
    {
        this = default(Semaphore);
        if (handle.HasValue)
        {
            Handle = handle.Value;
        }
    }

    public override string ToString()
    {
        return Handle.ToString();;
    }
}
```
it's useful for debug
![Image](https://github.com/user-attachments/assets/f1cf0dcd-fa14-4fdf-912f-639f5507d36e)

PR #2323, opened 2024-10-31
#2298 override ToString() method for vulkan objects

tonisimakov99:
# Summary of the PR
Adds an override of the ToString() method to structures to output the Handle field

[Issue 2298](https://github.com/dotnet/Silk.NET/issues/2298)

# Further Comments

### DIFF
--- a/src/Core/Silk.NET.BuildTools/Bind/StructWriter.cs
+++ b/src/Core/Silk.NET.BuildTools/Bind/StructWriter.cs
@@ -503,6 +503,16 @@ namespace Silk.NET.BuildTools.Bind
                 sw.WriteLine();
             }
 
+            var handleField = @struct.Fields.FirstOrDefault(t => t.Name == "Handle" && t.Type.Name == "ulong");
+            if (handleField != null)
+            {
+                sw.WriteLine("       public override string ToString()");
+                sw.WriteLine("       {");
+                sw.WriteLine("           return string.Format(\"0x{0:X}\", Handle);");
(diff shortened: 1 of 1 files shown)


Perksey (changes requested):
Hey, thanks so much for contributing to Silk.NET! And more notably, for trying to get stuck in to BuildTools - it's a rather complicated beast, and I always welcome anyone wanting to brave a change in there.

I actually swept this up around the same time as you submitted this PR in #2184, and I think I'll have that supersede this PR if that's okay with you? :) Feel free to see what I implemented over there re the comment below

src/Core/Silk.NET.BuildTools/Bind/StructWriter.cs
```diff
@@ -503,6 +503,16 @@ public static void WriteStruct
                 sw.WriteLine();
             }
 
+            var handleField = @struct.Fields.FirstOrDefault(t => t.Name == "Handle" && t.Type.Name == "ulong");
```
Perksey, line 506: Good spot, but this isn't _quite_ the right place to put it as this will generate for every structure (including those that don't have a `Handle` field). What I decided to do in my PR is handle this in the `VulkanReader` i.e. right at the source of the generation of those `Handle` fields. But something at this layer would've worked to with a check for a structure being handle-like! Many ways to skin a cat :)

Perksey (changes requested):
Didn't meant to press submit!

PR #2352, opened 2024-11-11
Dispose Glfw/Sdl When Checking if Applicable

Aragas:
# Summary of the PR
Calling `GLFW.Glfw.GetApi();` when Glfw is applicable without disposing it leaks the native module. So we make sure `Dispose()` is called, which triggers freeing of the native library and reduce the 'counter'.

# Further Comments  
This was discovered when I was trying to delete the native library (`glfw3.dll`) on program exit. The delete failed because the program was still referencing the module.  
The codebase uses the lazy `GlfwProvider.GLFW` where it makes sense. When needed, the lazy implementation will have the `Dispose()` called, thus freeing `glfw3`. But because of the applicable check, the module will never truly be freed without calling `GlfwProvider.GLFW.Value._ctx.Dispose();` manually at least once after a Window creation

### DIFF
--- a/src/Windowing/Silk.NET.Windowing.Glfw/GlfwPlatform.cs
+++ b/src/Windowing/Silk.NET.Windowing.Glfw/GlfwPlatform.cs
@@ -23,9 +23,10 @@ namespace Silk.NET.Windowing.Glfw
         (
             () =>
             {
+                GLFW.Glfw? api = null;
                 try
                 {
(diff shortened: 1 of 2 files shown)


Perksey:
Good spot! Would it be possible to make this change in SdlPlatform as well?

Aragas:
> Good spot! Would it be possible to make this change in SdlPlatform as well?

Good point, done

Aragas:
@dotnet-policy-service agree

Perksey (approved):
Thanks!

PR #2369, opened 2024-11-28
Workaround for UWP native build

bhatnagarsid:
Creating a workaround for UWP native build based on issue reported some years ago in [this issue](https://github.com/microsoft/dotnet/issues/807) and similar pull request addressing this issue in https://github.com/SharpGenTools/SharpGenTools/pull/210

When building Blank UWP application with Silk.NET pulled from nuget on x64 release build, seeing the following error:
```
Error MCG0037: MCG0037:InvalidCSharpIdentifierName Struct 'Silk.NET.Core.Bool32' in assembly 'Assembly(Name=Silk.NET.Core, Version=2.22.0.0, Culture=neutral, PublicKeyToken=null, Location=<>\obj\x64\Release\ilc\in\Silk.NET.Core.dll)' has a field with name '<Value>k__BackingField' that is invalid. This could be because the name is obfuscated or the field is auto-implemented by the compiler. Please make sure the field name follows C# identifier conventions.
```
Signature follows what is being reported in [this issue](https://github.com/microsoft/dotnet/issues/807)

This fix adds an explicit backing field for Bool32 class

### DIFF
--- a/src/Core/Silk.NET.Core/Miscellaneous/Bool32.cs
+++ b/src/Core/Silk.NET.Core/Miscellaneous/Bool32.cs
@@ -10,10 +10,12 @@ namespace Silk.NET.Core
     /// </summary>
     public readonly struct Bool32 : IEquatable<Bool32>, IEquatable<uint>, IEquatable<bool>
     {
+        private readonly uint _value;
+
         /// <summary>
         /// Gets the 32-bit value for this boolean.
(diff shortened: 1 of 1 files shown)


Perksey (approved):
Thanks! 👍I suspect this won't be the only issue you'll hit so please keep the PRs coming! :)

PR #2371, opened 2024-11-29
Bump Microsoft.Extensions.DependencyModel

### LINKED ISSUE #2370: Build warning due to vulnerable transitive System.Text.Json dependency
When using `Silk.NET.Core` as a dependency and compiling a project with the .NET 9 SDK, warnings regarding a vulnerability in the `System.Text.Json` package are shown:

```
> Project.csproj: Warning NU1903 : Package 'System.Text.Json' 8.0.0 has a known high severity vulnerability, https://github.com/advisories/GHSA-8g4q-xg66-9fp4
```
This is very annoying when multiple projects are built as the warning will be shown for each where `Silk.NET.Core` is a (transitive) dependency.
The cause for this seems to be that `Microsoft.Extensions.DependencyModel` (8.0.0), which is referenced from `Silk.NET.Core`, depends on `System.Text.Json ≥ 8.0.0`. As per the security advisory, the first patched version is `8.0.5`, so bumping `Microsoft.Extensions.DependencyModel` to at least `8.0.2` should fix this warning.

Morilli:
This is mainly to bump the transitive System.Text.Json dependency from 8.0.0 to 9.0.0, fixing https://github.com/advisories/GHSA-hh2w-p6rv-4g7w.

Closes #2370. As described in that issue, the System.Text.Json dependency is shipped to all consumers of the `Silk.NET.Core` package and will result in build warnings when building with the .NET 9 SDK.

### DIFF
--- a/src/Core/Silk.NET.Core/Silk.NET.Core.csproj
+++ b/src/Core/Silk.NET.Core/Silk.NET.Core.csproj
@@ -10,7 +10,7 @@
     <ItemGroup>
         <PackageReference Include="Microsoft.CSharp" Version="4.7.0" />
         <PackageReference Include="Microsoft.DotNet.PlatformAbstractions" Version="3.1.6" />
-        <PackageReference Include="Microsoft.Extensions.DependencyModel" Version="8.0.0" />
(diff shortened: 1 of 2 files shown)


Perksey (approved):
LGTM if build passes, thanks :)

PR #2389, opened 2024-12-20
[3.0] Streamline native builds (and build native libs for 3.0)

Perksey:
Adds SDL native builds and a much more streamlined native workflow.

More info in the for-contributors doc.


Perksey:
Removed that from the PR description as the Build/Test jobs, which are required to merge, don't run on the bot commits.

Perksey:
I don't know how to change it so it doesn't do that either, ideally we'd have `paths-filter` use the last dotnet-bot commit as the base for the compare (or obviously the target branch as it is today if we haven't done that yet) as it'd lessen the problem by making it so the last commit isn't always a bot commit. We'd have to fork the action, and I don't feel like doing that right now.

curin (approved):

PR #2403, opened 2024-12-30
Create building-from-source.md

joskuijpers:
# Summary of the PR
Adds information for contributors on how to build Silk from source.

# In Progress during my Silk adventures

PR #2405, opened 2024-12-30
[3.0] Contributor experience improvements

Perksey:
- Adds a basic for-contributors getting started README for building the repo, detailing what you need to get started. Closes #2403 which also does the same.
- Adds a "best effort" attempt to resolve native binaries when used as a `ProjectReference`.
- Adds the ability to disable building for Android/iOS by adding properties to Directory.Build.props which are then used when setting `TargetFrameworks`


Perksey:
The PR check is triggering because I've modified the native project to not include an Android build if disabled. This does not warrant a rebuild.

Perksey:
@curin Ready for review.

curin (approved):

Perksey:
@joskuijpers hopefully this helps? PTAL

joskuijpers:
Good changes! Playground now runs without SDL DLL issues, and the contributor document was very useful too.

PR #2409, opened 2025-01-09
ImGui DirectX11 Implementation

marcb152:
# Summary of the PR

This PR introduces initial support for ImGui with DirectX11 in Silk.NET. It includes:

- C# rewriting of [imgui_impl_dx11.cpp](https://github.com/ocornut/imgui/blob/master/backends/imgui_impl_dx11.cpp) for Silk.NET.
- ImGui controller very similar to the existing [OpenGL implementation](https://github.com/dotnet/Silk.NET/blob/main/src/OpenGL/Extensions/Silk.NET.OpenGL.Extensions.ImGui/ImGuiController.cs) as well as the [Lab Vulkan implementation](https://github.com/dotnet/Silk.NET/blob/main/src/Lab/Experiments/ImGuiVulkan/Lib/ImGuiController.cs) (actually I copied most, if not all, the code from the OpenGL controller).
- Minimal sample following the [existing OpenGL ImGui demo](https://github.com/dotnet/Silk.NET/blob/main/examples/CSharp/OpenGL%20Demos/ImGui/Program.cs) to demonstrate ImGui usage with DirectX11 (not started yet).

- [Discussed in the Discord server](https://discord.com/channels/521092042781229087/607634593201520651/1324401007563771995)

# Current progress

- [x] Working prototype in my own project [see](https://github.com/marcb152/DX11-Silk-Learning/blob/main/DX11_Silk.NET_Learning/ImGui_DX11_Impl/ImGui_Impl_DX11.cs)
- [x] Create directory & project under `/src/Lab/Experiments/ImGui_DX11/`
- [x] Create `ImGui_Impl_DX11.cs` (C++ port) and the associated `ImGuiDX11Controller.cs`
    - Not sure if I should keep those 2 files separated though? I am sticking to the original C++ file as DX11 code is a bit verbose and the controller would reach +1k lines very fast.
- [x] Create another directory and a sample minimal project to demonstrate
- [ ] Solve TODOs and clean up the code

# Further Comments

The changes should not introduce any breaking API or behavioral changes: they will extend the functionality of Silk.NET.

## Tested on

- Silk.NET v2.22.0
- ImGui.NET v1.91.6.1

## Remarks

This is actually my first PR to such a large open-source repo. Any help, constructive criticism, remarks, suggestions, improvements and feedback are greatly appreciated!


marcb152:
@dotnet-policy-service agree

marcb152:
I have pushed my prototype code all in one. My apologize for the long files!

As a recap:

- `ImGuiDX11Controller.cs` is mostly a copy paste from the existing OpenGL controller implementation, I replaced OpenGL calls by calls to `ImGui_Impl_DX11.cs` equivalent functions. Also assigned DirectX `hwnd` handle to ImGui (taken from [imgui_impl_win32.cpp](https://github.com/ocornut/imgui/blob/master/backends/imgui_impl_win32.cpp)). Not a lot of changes there.
- `ImGui_Impl_DX11.cs` is the full port of the original C++ and header (linked in the `README.txt`)
- `ImGuiFontConfig.cs` imported from the OpenGL impl, no changes at all.
- I created the `.csproj` file linking the necessary dependencies inside Silk.NET repo **⚠️never compiled yet!⚠️**

Perksey:
No worries. Is this ready for review? (it's still marked as draft :P)

marcb152:
Not yet, I've set aside my compilation issues and improved the code to support .NET 6.0 features (instead of the .NET 9.0 I created my prototype on).
So, the code needs at least C# 12 to compile, I'll push that ASAP (probably beginning of next week) then it'll be ready for review!

I've also made a minimal sample that works which I'll push soon afterwards.

marcb152:
Hey, I think this PR is now ready for review (I provided a sample as well).

There are still work to do (tell me if you need comments/self-explanatory code anywhere, this won't bother me). Most TODOs are out of my knowledge, there's one to check out as it can cause issues later:
https://github.com/dotnet/Silk.NET/pull/2409/files#diff-8c4147963edd0bf7166adef79bb3a37e316e17b1d610f2591590f81e6acd4ab9R593
It's the part where we revert back the state of the whole pipeline, I still don't understand how the bindings work with pointers to retrieve/restore arrays of viewports.

And sorry for the long wait also.

jjb-pro:
There appears to be a rendering bug in the DX11 backend.

`ImGui_ImplDX11_RenderDrawData` (ImGui_Impl_DX11.cs) currently does not bind `drawCmdPtr.TextureId` before issuing draw calls because the following code is commented out:

```csharp
// ID3D11ShaderResourceView* texture_srv = (ID3D11ShaderResourceView*)pcmd.TextureId;
// pd3dDeviceContext.PSSetShaderResources(0, 1, &texture_srv);
```

As a result, every ImGui draw command samples whatever SRV was already bound to slot 0. This breaks font rendering as soon as another texture is bound beforehand. It results in corrupted text rendering (rectangles instead of glyphs): 
grafik

The OpenGL backend performs the equivalent operation per draw command:

```csharp
_gl.BindTexture(GLEnum.Texture2D, (uint)cmdPtr.TextureId);
```

If rebinding per draw command caused issues previously, then the font texture binding should be moved before the draw loop instead of after it:

```csharp
pd3dDeviceContext.PSSetShaderResources(0u, 1u, ref pFontTextureView);
```

Currently, it is executed after all draw calls have already completed, and shortly afterward the previous state is restored via:

```csharp
pd3dDeviceContext.PSSetShaderResources(0, 1, in old.PSShaderResource);
```
So the `pFontTextureView` binding is effectively never used for any ImGui draw call if another SRV was already bound before entering `ImGui_ImplDX11_RenderDrawData`.

Perksey (changes requested):
Thanks so much for your contribution! My only comment is can we put the demo project in the lab as well? Usually we like the lab to be pretty self-contained (as it's all experimental stuff), and if we get to the point where it's explicitly production-ready, then we'll add an example in the obvious directory.

Alternatively, if you'd like this to go to production (i.e. not be contained in the lab), please do shout and I'll review accordingly :) the first change towards this would be creating a Silk.NET.Direct3D11.Extensions.ImGui project and moving this code there, and then making sure the public-facing API and implementation is up to the standards we'd expect.

But I'm happy with just merging this into the lab for posterity if you're happy to make the change to the demo project :) thank you again for your contribution

PR #2415, opened 2025-01-24
Add initial axis input proposal draft

domportera:
# Summary of the PR
Adds the axis-based input proposal's first draft

Several I've lost track of and I am sorry to report I am feeling particularly lazy after spewing so many words from my fingertips in a single afternoon

# Further Comments
Any and all feedback ahead of our discussion this sunday is welcome - even if just for clarity of writing, and especially re: the contents of the proposal

### DIFF
--- a/documentation/proposals/Proposal - Axis Input Devices.md
+++ b/documentation/proposals/Proposal - Axis Input Devices.md
@@ -1,0 +1,624 @@
+# Summary
+
+"Everything is an axis" - the idea that every input device can be abstracted into a series of axes (floating point values from 0-1), and an API for defining and working with such axes in a generic way.
+
+# Contributors
+
+- Dom Portera (@domportera)
+- Dylan Perks (@Perksey)
+
+# Current Status
+
+- [x] Proposed
+- [ ] Discussed with Community
+- [ ] Approved
+- [ ] Implemented
+
+# Dependencies
+This proposal assumes some knowledge of the [Multi-Backend Input proposal](./Proposal - Multi-Backend Input.md) and the previous version of this proposal.
+
+# Background
+
+This proposal is born of a conversation regarding the [Multi-Backend input proposal](./Proposal - Multi-Backend
+Input.md), where the topic of "Every input is an axis" was brought up. This proposal is an attempt to flesh out that idea into a proper API and lay the foundation for how this could enable more advanced features in the future.
+
+# Goals
+
+The hopes is to put a small level of burden on backend implementations, thoughtfully implementing this API, we will:
+
+- Make working with inputs more modular and consistent from both a high and low level
+- Allow for both higher-level and lower-level common code and functionality
+    - such as shared logic handling buttons, keys, accelerometers, gyroscopes, touchpads, etc
+- Allow for creative uses of backend input systems
+- Allow for the creation of simple or "free" input accessibility features
+- Provide device-specific input descriptions without the need for the user to know what device or backend is in use
+- Provide a foundation for future device-level key/button mapping
+- Provide a foundation for to enable the creation of straightforward or automatic virtual device creation systems
+- Enable the creation of extremely modular input systems, including the creation of virtual input devices, whereby an implementation could exist that allows for a straightforward, re-mappable implementation of device emulation
+    - to be further fleshed out in a future proposal.
+- Provide an API foundation for in-depth input accessibility features (related to the above-mentioned modularity)
+
+# Note on Requirements
+
+This proposal can be thought of in two primary areas: the API surface, and the implementation requirements. Wherever it was possible, the API surface was designed to enforce certain requirements, however there are quite a few requirements that must be followed that are not enforced by the API itself. Comments have been added to the API where these requirements are necessary and further detail is provided where needed. Feedback on clarity is welcome.
+
+# Core API
+
+The root of the API is in the interface `IAxisDevice`. This interface is essentially a host for collections of axes paired with definitions of these axes and how they relate to one another. In order to be used, each `IAxisDevice` must be validated, similar to Vulkan's validation layers, through the use of the `AxisDeviceRegistry` class. 
+
+It is important to note that the value of an input axis a single 32-bit floating point value. is primarily expected between 0-1.0. There are some axis types, or rather `AxisTrait`s, that are not straightforwardly normalized, and those respective axes can be labeled as such (more on that later).
+
+## IAxisDevice
+
+```csharp
+/// <summary>
+/// This represents a single input device, physical or virtual, that is comprised of a collection of axes (or floats)
+/// The IReadOnlyList<float> interface is used to provide the normalized values of each axis, if it is indeed normalized (most should be). Otherwise, it would provide the raw, unnormalized value.
+/// </summary>
+public interface IAxisDevice : IReadOnlyList<float>, IInputDevice
+{
+    /// <summary>
+    /// The main entry-point for useful information about common axis groups - e.g. joysticks, trackpads, buttons, gyros, etc.
+    /// Turns single axes/floats into more broadly useable information
+    /// Must be provided in order of <see cref="AxisGroup.Index"/>
+    /// </summary>
+    IReadOnlyList<AxisGroup> Groups { get; }
+
+    /// <summary>
+    /// The description of each axis in the hardware device
+    /// This information is used primarily to allow for programmatically defined axis groups, axis devices, runtime validation,
+    /// and custom device mappings and handling.
+    /// Must be provided in order of <see cref="AxisDescription.Index"/>
+    /// </summary>
+    IReadOnlyList<AxisDescription> Axes { get; }
+
+    /// <summary>
+    /// Provides raw values for each axis as defined by the device's native API
+    /// Must be provided in order of <see cref="AxisDescription.Index"/>
+    /// </summary>
+    IReadOnlyList<float> RawValues { get; }
+
+    /// <summary>
+    /// A human-readable description of the device, intended for end-user display
+    /// </summary>
+    public string? Description { get; }
+
+    /// <summary>
+    /// A human-readable description of the device's status, intended for end-user display
+    /// In the case of device malfunctions, especially those that are composed of multiple devices,
+    /// this can be used to display any issues preventing the device from functioning properly
+    /// </summary>
+    public string? Status { get; }
+    
+    /// <summary>
+    /// False if this is a single real-world device, true if this is a device based on one or more physical devices,
+    /// or is entirely virtual (e.g. a pointer in a VR environment, a virtual gamepad, etc)
+    ///
+    /// Note: defining "virtual" input devices should not *necessarily* come with
+    /// the burden of a full-blown IAxisDevice implementation. Future proposal may include a request to move this to IInputDevice, which would help make the IInputDevice <-> IAxisDevice APIs more consistent with each other.
+    /// </summary>
+    bool IsVirtual { get; }
+}
+```
+
+## IAxisInputHandler
+
+This is how a developer could subscribe directly to the axis changes of an `IAxisDevice`. It mimicks the `IGamepadInputHandler` interface from the [Multi-Backend Input proposal](./Proposal - Multi-Backend Input.md).
+
+```csharp
+
+public interface IAxisInputHandler
+{
+    public void OnAxisChanged(IAxisDevice device, AxisDescription description, float value);
+}
+```
+
+## Axis Descriptions
+
+There are two primary structures that give meaning to the axes of an `IAxisDevice`: `AxisDescription` and `AxisGroup`. The former defines the axes themselves, and the latter defines how these axes relate to one another. Note that both of these structures have an `Index` field, which must match the index of the `IAxisDevice`'s collection that it is defined in.
+
+### Struct Definition
+
+```csharp
+/// <summary>
+/// This describes a single axis on a device
+/// </summary>
+/// <param name="Index">the index of the axis in the device</param>
+/// <param name="Traits">the traints of the axis, described above</param>
+/// <param name="Name">A human-readable name for the axis</param>
+/// <param name="RawValueBounds">The bounds of the raw value - a value of 'default' indicates it is typical (0, 1), or
+/// in the case that the provided traits are <see cref="AxisTrait.RawValueOnly"/>, this must be default and indicates no known bounds.</param>
+public readonly record struct AxisDescription(
+    int Index,
+    AxisTrait Traits,
+    string Name,
+    Vector2 RawValueBounds = default)
+{
+    public static implicit operator int(AxisDescription description) => description.Index;
+}
+```
+
+### Axis Traits
+These are the primary flags through which an axis's use is defined. Its definition includes several constraints meant to both make defining axes less cumbersome and enforce logically consistent behaviour where possible. This enum is one of the primary places where validation needs to occur at runtime, and some flags place implementation requirements on the implementer. Wherever possible, we should seek to provide default implementations of expected behaviours regarding these constraints.
+
+```csharp
+
+/// <summary>
+/// This is how individual axes are defined
+/// Pay attention to the flags which require specific implementation details related to their RawValues and normalization
+/// </summary>
+[Flags]
+public enum AxisTrait
+{
+    Unknown = 0,
+    Analog = 1 << 0,
+    Binary = 1 << 1,
+
+    // an analog axis that explicitly does not return to center or zero like a joystick or trigger
+    // things like touch-pads, mice, or world-space positions
+    Point = 1 << 8 | Analog | HasRawValue,
+
+    // ---- Physical orientation ----
+    Orientation = 1 << 2 | Analog,
+    Rotation = 1 << 3 | Orientation,
+
+    // this is specified because single axes of euler angle components can be useful by themselves,
+    // whereas anything shorter than 3 or 4 axes is not for quaternion rotation components
+    /// <summary>
+    /// Indicates that this axis is a component of a euler angle (do we want this to be in degrees or radians?)
+    /// If this flag is set, the axis is expected to be normalized between 0-360 degrees, with wrapping applied
+    /// It is required you do NOT set <see cref="AxisDescription.RawValueBounds"/>, as it will be assumed to be (0, 360/2pi).
+    /// </summary>
+    EulerAngleComponent = 1 << 4 | Rotation | HasRawValue,
+
+    /// <summary>
+    /// Acceleration - in meters per second <br/>
+    /// If this is set, it is expected that the axis is NOT provided any raw value constraints,
+    /// as it is not expected to be normalized
+    /// </summary>
+    Acceleration = 1 << 6 | Orientation | RawValueOnly,
+
+    /// <summary>
+    /// Position - raw values are in meters if world-space, in pixels if screen-space, etc <br/>
+    /// Normalized values are required if the bounds are known via the device's API.
+    /// If they are unknown, this must be paired with <see cref="RawValueOnly"/>
+    /// </summary>
+    Position = 1 << 7 | Orientation | Point,
+    // -----------------------------
+
+    /// <summary>
+    /// Indicates that the raw value provided for this axis is not the same as the normalized value. <br/>
+    /// If this flag is signaled on an axis, it is highly recommended that the bounds are provided as well, even if they are infinite
+    /// This can be useful for things like euler angle components, where single axes 0-360 degrees (or 0-2pi radians? do we want to standardize which?) can be useful by themselves
+    /// </summary>
+    HasRawValue = 1 << 24 | Analog,
+
+    /// <summary>
+    /// Indicates that there is no one-size-fits-all way to normalize this axis, and must be handled in a case-by-case basis <br/>
+    /// Examples of this would be: <br/>
+    /// - Accelerometers, as the relevant magnitude differs depending on the application <br/>
+    /// - VR controller position without a bounding area <br/>
+    /// - Quaternion rotation values, as these can be expected to be negative <br/>
+    /// <br/>
+    /// If this is set, it is expected that the axis is NOT provided any raw value constraints
+    /// </summary>
+    RawValueOnly = 1 << 25 | HasRawValue,
+
+    /// <summary>
+    /// Things like a mouse movement, trackpad, etc
+    /// </summary>
+    Delta = 1 << 26 | HasRawValue,
+
+    /// <summary>
+    /// Indicates that there is no one-size-fits-all way to normalize this axis, and must be handled in a case-by-case basis <br/>
+    /// Examples of this would be: <br/>
+    /// - Mouse position <br/>
+    /// - Trackpad without an API for "absolute" position
+    /// </summary>
+    DeltaOnly = 1 << 27 | Delta | RawValueOnly,
+
+    /// <summary>
+    /// Used for axes that are on the left side of a device from the user's perspective.
+    /// Useful for left/right handedness accommodation and various symmetrical devices. <br/><br/>
+    /// For example: a left stick on a gamepad, a left trigger, a left shoulder button, the DPad on a gamepad, etc
+    /// </summary>
+    LeftSide = 1 << 30,
+
+    /// <summary>
+    /// Used for axes that are on the right side of a device from the user's perspective.
+    /// Useful for left/right handedness accommodation and various symmetrical devices. <br/><br/>
+    /// For example: a right stick on a gamepad, a right trigger, a right shoulder button, the face buttons on a gamepad, etc
+    /// </summary>
+    RightSide = 1 << 31,
+}
+```
+
+### Example Definition List
+
+The following is a demonstration of how this `AxisDescription` list could be populated by an input backend:
+
+```csharp
+public class DualsenseDevice : IAxisDevice
+{
+    IReadOnlyList<AxisDescription>
+    private static readonly IReadOnlyList<AxisDescription> Axes =
+    [
+        // left analog stick
+        new(0, AxisTrait.Analog | AxisTrait.LeftSide, "Left Joystick X-"),
+        new(1, AxisTrait.Analog | AxisTrait.LeftSide, "Left Joystick X+"),
+        new(2, AxisTrait.Analog | AxisTrait.LeftSide, "Left Joystick Y-"),
+        new(3, AxisTrait.Analog | AxisTrait.LeftSide, "Left Joystick Y+"),
+        new(4, AxisTrait.Binary | AxisTrait.LeftSide, "Left Joystick Press"),
+        new(5, AxisTrait.Binary | AxisTrait.LeftSide, "Left Joystick Touch"),
+
+
+        // right analog stick
+        new(6, AxisTrait.Analog | AxisTrait.RightSide, "Right Joystick X-"),
+        new(7, AxisTrait.Analog | AxisTrait.RightSide, "Right Joystick X+"),
+        new(8, AxisTrait.Analog | AxisTrait.RightSide, "Right Joystick Y-"),
+        new(9, AxisTrait.Analog | AxisTrait.RightSide, "Right Joystick Y+"),
+        new(10, AxisTrait.Binary | AxisTrait.RightSide, "Right Joystick Press"),
+        new(11, AxisTrait.Binary | AxisTrait.RightSide, "Right Joystick Touch"),
+
+        // d-pad
+        new(12, AxisTrait.Binary | AxisTrait.LeftSide, "D-Pad Left"),
+        new(13, AxisTrait.Binary | AxisTrait.LeftSide, "D-Pad Right"),
+        new(14, AxisTrait.Binary | AxisTrait.LeftSide, "D-Pad Down"),
+        new(15, AxisTrait.Binary | AxisTrait.LeftSide, "D-Pad Up"),
+
+        // face buttons
+        new(16, AxisTrait.Binary | AxisTrait.RightSide, "Square"),
+        new(17, AxisTrait.Binary | AxisTrait.RightSide, "Circle"),
+        new(18, AxisTrait.Binary | AxisTrait.RightSide, "Cross"),
+        new(19, AxisTrait.Binary | AxisTrait.RightSide, "Triangle"),
+
+        // touchpad
+        new(20, AxisTrait.Binary, "Touchpad Press"),
+
+        // touchpad touch points - up to 4 touch points
+        new(21, AxisTrait.Position, "Touchpad X1"),
+        new(22, AxisTrait.Position, "Touchpad Y1"),
+        new(23, AxisTrait.Position, "Touchpad X2"),
+        new(24, AxisTrait.Position, "Touchpad Y2"),
+        new(25, AxisTrait.Position, "Touchpad X3"),
+        new(26, AxisTrait.Position, "Touchpad Y3"),
+        new(27, AxisTrait.Position, "Touchpad X4"),
+        new(28, AxisTrait.Position, "Touchpad Y4"),
+
+        // select/start/home
+        new(29, AxisTrait.Binary | AxisTrait.LeftSide, "Share"),
+        new(30, AxisTrait.Binary | AxisTrait.RightSide, "Start"),
+        new(31, AxisTrait.Binary, "Home"),
+
+        // shoulder buttons
+        new(32, AxisTrait.Binary | AxisTrait.LeftSide, "L1"),
+        new(33, AxisTrait.Binary | AxisTrait.RightSide, "R1"),
+        new(34, AxisTrait.Analog | AxisTrait.LeftSide, "L2"),
+        new(35, AxisTrait.Analog | AxisTrait.RightSide, "R2"),
+
+        // gyro
+        new(36, AxisTrait.EulerAngleComponent, "Gyro X"),
+        new(37, AxisTrait.EulerAngleComponent, "Gyro Y"),
+        new(38, AxisTrait.EulerAngleComponent, "Gyro Z"),
+
+        // accelerometer
+        new(39, AxisTrait.Acceleration, "Accelerometer X"),
+        new(40, AxisTrait.Acceleration, "Accelerometer Y"),
+        new(41, AxisTrait.Acceleration, "Accelerometer Z"),
+
+        // battery level
+        new(42, AxisTrait.Analog, "Battery Level"),
+
+        // microphone input level
+        new(43, AxisTrait.Analog, "Microphone Level"),
+    ];
+}
+ ```
+
+As additional food for thought, a Sony Dualshock 2 controller whose face buttons are famously (though infrequently used as) analog, might contain face buttons with the following definition:
+
+```csharp
+_ = new AxisDescription(0, AxisTrait.Binary | AxisTrait.Analog | AxisTrait.RightSide, "Square");
+```
+
+## Axis Groups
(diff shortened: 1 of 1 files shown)


domportera:
@dotnet-policy-service agree

documentation/proposals/Proposal - Axis Input Devices.md
```diff
@@ -0,0 +1,624 @@
...
+These are the primary flags through which an axis's use is defined. Its definition includes several constraints meant to both make defining axes less cumbersome and enforce logically consistent behaviour where possible. This enum is one of the primary places where validation needs to occur at runtime, and some flags place implementation requirements on the implementer. Wherever possible, we should seek to provide default implementations of expected behaviours regarding these constraints.
+
+```csharp
+
+/// <summary>
+/// This is how individual axes are defined
+/// Pay attention to the flags which require specific implementation details related to their RawValues and normalization
+/// </summary>
+[Flags]
+public enum AxisTrait
```
Perksey, line 151:
```suggestion
public enum AxisTrait : ulong
```
Future proofing

documentation/proposals/Proposal - Axis Input Devices.md
```diff
@@ -0,0 +1,624 @@
...
+    // an analog axis that explicitly does not return to center or zero like a joystick or trigger
+    // things like touch-pads, mice, or world-space positions
+    Point = 1 << 8 | Analog | HasRawValue,
+
+    // ---- Physical orientation ----
+    Orientation = 1 << 2 | Analog,
+    Rotation = 1 << 3 | Orientation,
+
+    // this is specified because single axes of euler angle components can be useful by themselves,
+    // whereas anything shorter than 3 or 4 axes is not for quaternion rotation components
```
Perksey, line 166: Is the ordering of `AxisDescription.Index` important for grouping multiple rotation components into a quaternion for instance? If so, should be specified in the group requirements. Is azimuth/altitude considered here or does this have to be converted before being reported as an axis? And other similar questions - basically how is the user of the axis device intended to group and interpret rotational components into a transformation they can apply to a 3D model for instance.

documentation/proposals/Proposal - Axis Input Devices.md
```diff
@@ -0,0 +1,624 @@
...
+    /// as it is not expected to be normalized
+    /// </summary>
+    Acceleration = 1 << 6 | Orientation | RawValueOnly,
+
+    /// <summary>
+    /// Position - raw values are in meters if world-space, in pixels if screen-space, etc <br/>
+    /// Normalized values are required if the bounds are known via the device's API.
+    /// If they are unknown, this must be paired with <see cref="RawValueOnly"/>
+    /// </summary>
+    Position = 1 << 7 | Orientation | Point,
```
Perksey, line 186: Again are we assuming that the order of the axes within the group matters so we can interpret these axes as X/Y/Z etc.

documentation/proposals/Proposal - Axis Input Devices.md
```diff
@@ -0,0 +1,624 @@
...
+
+    /// <summary>
+    /// Things like a mouse movement, trackpad, etc
+    /// </summary>
+    Delta = 1 << 26 | HasRawValue,
+
+    /// <summary>
+    /// Indicates that there is no one-size-fits-all way to normalize this axis, and must be handled in a case-by-case basis <br/>
+    /// Examples of this would be: <br/>
+    /// - Mouse position <br/>
```
Perksey, line 215: We probably need to look more into how we map an axis device into a pointer device for instance, I'd imagine a lot of questions I may have will naturally occur in trying to define this. (trying to think of the pointer targets and the different possible bounds)

documentation/proposals/Proposal - Axis Input Devices.md
```diff
@@ -0,0 +1,624 @@
...
+/// <param name="Purpose">The intended purpose of this group</param>
+/// <param name="Axes">The indices of the axes that make up this group</param>
+/// <param name="TwinIndex">The AxisGroup index of the "twin" of this group - e.g. if this is the left thumbstick, the twin would be the right thumbstick</param>
+/// <param name="Name">A human-readable name for the group</param>
+/// <remarks>Can names be inferred at runtime from the AxisDescription and AxisGroup? Should they be nullable/optional
+/// in the definition and then populated at runtime if they're missing?</remarks>
+public readonly record struct AxisGroup(
+    int Index,
+    AxisGroupType Purpose,
+    IReadOnlyList<int> Axes,
```
Perksey, line 343: Should we just do FirstAxisIndex & AxisCount rather than allocating an additional list here? Is there ever a case where a groups axes will not be contiguous?

documentation/proposals/Proposal - Axis Input Devices.md
```diff
@@ -0,0 +1,624 @@
...
+    /// made of 6 physical axes (-X, +X, -Y, +Y, -Z, +Z)
+    /// If constructed with 7 axes, the 7th axis is the "pressure" axis for a pointer
+    /// </summary>
+    Position3D = 1u << 8,
+
+    // should this be 6 and 8 axes, similar to how a 2d joystick is defined by 4 axes?
+    RotationEuler = 1u << 9, // gyroscope, gyroscope + magnetomete, VR peripheral rotation - requires 3 axes, XYZ order
+
+    RotationQuaternion =
+        1u << 10, // gyroscope, gyroscope + magnetomete, VR peripheral rotation - requires 4 axes, XYZW order
```
Perksey, line 431: Aha you've answered my questions! Will keep them posted to provide insight into the interpretation of this doc but will mark them as resolved.

documentation/proposals/Proposal - Axis Input Devices.md
```diff
@@ -0,0 +1,624 @@
...
+
+
+
+```csharp
+/// <summary>
+/// This is how one defines the purpose of groups of multiple axes in a device
+/// It has specifications that must be followed to be considered valid
+/// </summary>
+[Flags]
+public enum AxisGroupType : uint
```
Perksey, line 364:
```suggestion
public enum AxisGroupType : ulong

```

documentation/proposals/Proposal - Axis Input Devices.md
```diff
@@ -0,0 +1,624 @@
...
+
+could instead be defined as:
+```csharp
+new(3, AxisGroupType.DPad | AxisGroupType.LeftHanded, [14, 15, 16, 17], 6, "D-Pad"),
+new(4, AxisGroupType.FourFaceButtons | AxisGroupType.LeftHanded, [14, 15, 16, 17], 6, "D-Pad"),
+```
+As previously stated, the latter grants you the benefit of having a defined order of primacy for the axes in the D-Pad group, which can be useful for certain implementations.
+
+
+Though no such API is defined in this proposal, I strongly encourage the development of extension methods to aid in determining the "ranking" of `AxisGroup`s for any given axis, as well as determining the "default" set of axis groups that make up a given device. (Request for feedback: would the latter require an additional property in `IAxisDevice`?)
```
Perksey, line 488:
```suggestion
Though no such API is defined in this proposal, the Silk.NET team reserves the right to expose extension methods to aid in determining the "ranking" of `AxisGroup`s for any given axis, as well as determining the "default" set of axis groups that make up a given device. (Request for feedback: would the latter require an additional property in `IAxisDevice`?)

```

documentation/proposals/Proposal - Axis Input Devices.md
```diff
@@ -0,0 +1,624 @@
...
+}
+```
+
+## The Handlers
+
+```csharp
+/// <summary>
+/// An interface for handling deadzones in different ways
+/// </summary>
+public interface IDeadzoneHandler
```
Perksey, line 564: Sceptical about calling this `IDeadzoneHandler` given that in our input API "handlers" usually refer to our actor-like model for handling input events.

Maybe `IDeadzoneDeterminant`, or maybe call this `IDeadzone` and rename `Deadzone` to `DeadzoneBounds`? Or something else?

documentation/proposals/Proposal - Axis Input Devices.md
```diff
@@ -0,0 +1,624 @@
...
+    /// </summary>
+    public IReadOnlyList<IAxisDevice> PhysicalDevices { get; }
+
+    // A future composability proposal can include additional properties to retrieve all available axis groups, irrespective
+    // of the source devices
+    
+    // there may also be a case for making the above 3 collection types (All, Virtual, Physical) a model that can be followed by 
+    // the standard input context, which would be the primary benefit to adding `IsVirtual` to `IInputDevice` - to be discussed in a future proposal
+}
+```
```
Perksey, line 624: I don't see why we need this when we have `InputContext` which could be changed to implicitly validate when it encounters an `IAxisDevice` in a backends, and connections/disconnections are handled in the obvious way as already specified by the base proposal. If there is a desire to register/deregister individual devices then there should be an implementation of `IInputBackend` that accommodates this.

Perksey (changes requested):
I **_love_** this proposal!

documentation/proposals/Proposal - Axis Input Devices.md
Perksey: Also I'm not too sure what you mean in this request for feedback

documentation/proposals/Proposal - Axis Input Devices.md
```diff
@@ -0,0 +1,624 @@
...
+    public string? Status { get; }
+    
+    /// <summary>
+    /// False if this is a single real-world device, true if this is a device based on one or more physical devices,
+    /// or is entirely virtual (e.g. a pointer in a VR environment, a virtual gamepad, etc)
+    ///
+    /// Note: defining "virtual" input devices should not *necessarily* come with
+    /// the burden of a full-blown IAxisDevice implementation. Future proposal may include a request to move this to IInputDevice, which would help make the IInputDevice <-> IAxisDevice APIs more consistent with each other.
+    /// </summary>
+    bool IsVirtual { get; }
```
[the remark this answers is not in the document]
domportera: I'd love to have these in IInputDevice, but I fear that DIMming them could lead to them being "hidden" when Description and Status in particular seem pretty important to define, with very little involved in implementing them. ok maybe not "important", but having some implementations not have it by accident sort of diminishes the usefulness of the feature as a whole. IsVirtual I think makes perfect sense to be DIMmed though.

documentation/proposals/Proposal - Axis Input Devices.md
```diff
@@ -0,0 +1,624 @@
...
+
+## IAxisInputHandler
+
+This is how a developer could subscribe directly to the axis changes of an `IAxisDevice`. It mimicks the `IGamepadInputHandler` interface from the [Multi-Backend Input proposal](./Proposal - Multi-Backend Input.md).
+
+```csharp
+
+public interface IAxisInputHandler
+{
+    public void OnAxisChanged(IAxisDevice device, AxisDescription description, float value);
```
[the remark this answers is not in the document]
domportera: i don't feel strongly one way or another. AxisDescription is a relatively small struct, and idk what the cache-related consequences are.

note: I used to reflexively `in` structs > 8 bytes until you told me there was cache related consequences so if it sounds good to you it's def good to me

documentation/proposals/Proposal - Axis Input Devices.md
```diff
@@ -0,0 +1,624 @@
...
+
+/// <summary>
+/// This is how individual axes are defined
+/// Pay attention to the flags which require specific implementation details related to their RawValues and normalization
+/// </summary>
+[Flags]
+public enum AxisTrait
+{
+    Unknown = 0,
+    Analog = 1 << 0,
```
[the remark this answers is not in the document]
domportera: yes i will do that shortly

documentation/proposals/Proposal - Axis Input Devices.md
```diff
@@ -0,0 +1,624 @@
...
+
+### Example Definition List
+
+The following is a demonstration of how this `AxisDescription` list could be populated by an input backend:
+
+```csharp
+public class DualsenseDevice : IAxisDevice
+{
+    IReadOnlyList<AxisDescription>
+    private static readonly IReadOnlyList<AxisDescription> Axes =
```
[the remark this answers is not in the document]
domportera: I haven't yet defined any sort of API for axis output, only input. my thoughts were outputs could be a non-breaking extension to this API, but to be honest I haven't thought at length about it. i have no problem re-visiting this to properly complete the axis list though

documentation/proposals/Proposal - Axis Input Devices.md
domportera: yes I think there is ample opportunity for axes to be non-contiguous. I believe most implementations would be able to be contiguous, but some would be awkward if not impossible. I think WASD would be the most common case where they would likely be non-contiguous, though aside from keyboard inputs it becomes more of a niche issue. But i'd like to avoid closing any doors if i can. 

though I also don't like the need for a collection, i couldn't think of a better way that's still straight-forward

documentation/proposals/Proposal - Axis Input Devices.md
domportera: ultimately i have no issue with this whatsoever,  but I think there is plenty of room as-is, and it might be good to discourage over-populating this enum. that being said, we all know this is going to need additions at *some* point in the future, so I have no strong feelings.

documentation/proposals/Proposal - Axis Input Devices.md
domportera: I would much prefer that - just wanted to silo things so as not to disturb The Great Base Proposal. Though, if we were to do that, I would also request the addition of the `VirtualDevices` and `Physical Devices` collections to the `InputContext`

documentation/proposals/Proposal - Axis Input Devices.md
domportera: the naming was an attempt to keep the API consistent with how people will be used to handling input events, though if this doesn't actually follow the model of the others then I'm totally down to change the name. Though not crazy about those names in particular bc i think it's less clear

we can shop it with The Gang? ¯\_(ツ)_/¯

documentation/proposals/Proposal - Axis Input Devices.md
domportera: fuck it i'm shipping the ulong bc i got some evil ideas for the future

documentation/proposals/Proposal - Axis Input Devices.md
domportera: I think all you need to worry about is the fact that this API will provide you with a normalized 1d, 2d, 3d position value from the input device, as well as any relevant rotation information. projecting that into any different bounds should be fairly straightforward from there as a mapping function if a user wants to hone in on a subset of a device surface, no?

documentation/proposals/Proposal - Axis Input Devices.md
```diff
@@ -0,0 +1,624 @@
...
+/// </summary>
+[Flags]
+public enum AxisTrait
+{
+    Unknown = 0,
+    Analog = 1 << 0,
+    Binary = 1 << 1,
+
+    // an analog axis that explicitly does not return to center or zero like a joystick or trigger
+    // things like touch-pads, mice, or world-space positions
```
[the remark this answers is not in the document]
domportera: in the case of a touchpad, I would expect the `IAxisDevice` to have as many axes as fingers it can support, with the axis value being the effective touch pressure.

If, in the future, certain (more exotic) use cases make this unwieldy, I think there's room for an extension for "Dynamic Axes"

documentation/proposals/Proposal - Axis Input Devices.md
domportera: just a note - azimuth would likely be a Rotation axis group that could be separated from the "gyroscopic" rotation if need be,  and altitude would likely be the Z component of a Position3D axis group, or a single Position1D axis group

documentation/proposals/Proposal - Axis Input Devices.md
Perksey: Is there really a case where just looping and testing whether IsVirtual is set isn’t sufficient? We can provide utility properties like these but I’d imagine that’d be all it’s doing under-the-hood given that I don’t think we’d want to allocate extra for something that probably isn’t used in hot paths (and if they are to be used in hot paths, then I’d encourage those users to handle the connection event and maintain their own lists)

documentation/proposals/Proposal - Axis Input Devices.md
Perksey: Will discuss in meeting.

documentation/proposals/Proposal - Axis Input Devices.md
Perksey: Ok

documentation/proposals/Proposal - Axis Input Devices.md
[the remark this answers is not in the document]
Perksey: This could probably be eased by adding an event that occurs when the axes change i.e. so axis devices can add and remove axes at will.

documentation/proposals/Proposal - Axis Input Devices.md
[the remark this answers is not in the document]
Perksey: I do think it would be a shame to accept the proposal not having a path to implementing the base proposal in terms of this proposal, as that is ultimately one of the main goals.

A lot of the benefits we get from this axis model still exist without that mind e.g. it makes developing an input action model a lot easier, but we should have a clear strategy for getting to the future we want. And if we accept this proposal without everything, we open up opportunities to have fragmented implementations given that we'd have to care about backwards compatibility if it does get implemented without the other proposals being approved.

documentation/proposals/Proposal - Axis Input Devices.md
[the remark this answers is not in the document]
Perksey: Most of the time where you're going to use the whole struct, the struct isn't large, and there's no enregistration concerns (e.g. vectors) for fields highly likely to be accessed, then yeah by-value is probably what you want. I don't think every possible implementation of this will use every single field, and I'm under the impression that most of the times it's only the index that matters after some initialisation that the higher-level wrappers do to map the axes.

But the main thing here specifically is I really want the`IAxisDevice` collections to be a one-and-done initialisation, with everything else being defined in terms of indexes or refs into those immutable collections.

documentation/proposals/Proposal - Axis Input Devices.md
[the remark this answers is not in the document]
domportera: i love the idea of having axis events in this way, i was just under the impression that you weren't a fan of events within device implementations for object lifetime reasons. would be pretty straightforward to plug into this proposal if these dynamic axes are populated and accessed in a separate collection ,otherwise axis groups would need to adapt in realtime, and it would be more straightforward to deal with the special behavior of these things from an input bindings perspective

documentation/proposals/Proposal - Axis Input Devices.md
domportera: I think it is sufficient and isnt too much to ask, though i can see the case where *only* virtual or *only* physical devices will be preferred, and indexing could be hampered without dedicated collections

documentation/proposals/Proposal - Axis Input Devices.md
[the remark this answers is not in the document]
Perksey: Oh yeah by events I mean handler methods using the actor-style model a la the base proposal because of the lifetime considerations.

I don't want to separate them out, but you're right that we need a way to convey delta updates. Perhaps we can add a fixed flag, and make it illegal for a fixed group to come after a dynamic group i.e. all fixed ones need to come first to prevent the indices being invalidated, same with the axes themselves. And then the handler methods can be axis added/axis removed. Any changes to fixed axes will need to be a disconnect and reconnect (which can be emulated by the input backend if it wants) as this is assumed to be a slow path.

documentation/proposals/Proposal - Axis Input Devices.md
Perksey: Then the user can make those collections in that wrapper :)

I don't want to force everyone into overhead for those small use cases, and generally want to try and avoid having too many points-of-entry here as this could be more confusing.

documentation/proposals/Proposal - Axis Input Devices.md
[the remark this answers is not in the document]
domportera: lemme stew a bit. i just dont want to paint ourselves into a corner with device outputs (since some may not be very simple - e.g. streaming outputs that arent abstractable to a floating point value)

documentation/proposals/Proposal - Axis Input Devices.md
domportera: that is Valid and Agreeable!

documentation/proposals/Proposal - Axis Input Devices.md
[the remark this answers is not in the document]
domportera: yeah re: indexing immutable collections we have the same goal (barring "dynamic axes")

i will gladly mark this as `in`!

documentation/proposals/Proposal - Axis Input Devices.md
```diff
@@ -0,0 +1,611 @@
...
+# Core API
+
+The root of the API is in the interface `IAxisDevice`. This interface is essentially a host for collections of axes paired with definitions of these axes and how they relate to one another. 
+
+It is important to note that the value of an input axis a single 32-bit floating point value. is primarily expected between 0-1.0. There are some axis types, or rather `AxisTrait`s, that are not straightforwardly normalized, and those respective axes can be labeled as such (more on that later).
+
+## Device Validation
+Due to the complex set of constraints detailed below, it is important that a validation layer exists to ensure that higher-level code can trust the device is following important conventions and constraints. As a result, each `IAxisDevice` must be validated, similar to Vulkan's validation layers, through the current `IInputContext` if such a context desires to make reliable use of this API. The Silk.NET team shall make their validation implementation easily callable by custom `IInputContext` implementations due to its inevitable complexity.
+
+The signature of Silk's validation method should resemble the following:
```
Perksey, line 85: Just some nits regarding:
- `InputContext` no longer being an interface
- some of the intended uses of the API

Apologies that this isn't a `suggestion`, can't have code blocks within code blocks.

```suggestion
Due to the complex set of constraints detailed below, it is important that a validation layer exists to ensure that higher-level code can trust the device is following important conventions and constraints. As a result, each `IAxisDevice` must be validated, similar to Vulkan's validation layers, before `InputContext` can use the device to create a wrapper/implementation of the higher-level interfaces defined in Multi-Backend Input (or otherwise make use of the raw axes). This validation is done by `InputContext` if used, however if the user is using their own device aggregator type then it is important that the validation implementation is easily callable by those implementations due to its inevitable complexity.

This validation shall be exposed as follows:
```

documentation/proposals/Proposal - Axis Input Devices.md
```diff
@@ -0,0 +1,611 @@
...
+
+```csharp
+public static bool IsAxisDeviceValid(IAxisDevice device, [NotNullWhen(false)] out string? error);
```
Perksey, line 88:
```suggestion
namespace Silk.NET.Input;

public partial interface IAxisDevice
{
    static sealed void Validate(IAxisDevice device);
    static sealed bool TryValidate(IAxisDevice device);
}
```
We can bikeshed these methods if we want, I just don't like the opaque error string as you can't really programmatically rely on its value given it could be anything. It's an exceptional circumstance that a device is invalid, hence this should be given in exceptions, unless for some reason the user is tolerant of invalid device in which case we expose the try-style method (but no further info as again this could be anything...)

domportera: I'm personally happy with an exception - just didnt know just how core you were willing to take this (e.g. it MUST be correct or it SHOULD be correct but some implementations might lag behind etc etc)

documentation/proposals/Proposal - Axis Input Devices.md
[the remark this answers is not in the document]
domportera: I've added relevant flags for this, however I'm not sure what the "event" in this situation would look like from the `IAxisDevice` perspective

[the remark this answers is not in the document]
Perksey: Would be adding something to IAxisInputHandler similar to the connection/disconnection events in the base proposal. Like a HandleAxesChanged with a list of axis indices, same with groups. Or something like that.

Also come to think of it all the IAxisInputHandler methods should use record structs similar to the base proposal and Enhanced Input Events that came before it (the rationale being it’s easier to add more data to these events in a non-breaking way). This supersedes my `in` comment (though I guess we could use a ref readonly struct containing a ref readonly if we wanted, but that’s probably a bit excessive)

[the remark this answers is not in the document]
domportera: makes sense! In that case does `IAxisDevice` also need state properties pertaining to the arguments of those methods? 

Also as a side note - do you think that anything related to multitouch surfaces should probably still have at least one "fixed" axis for touch, with additional ones being "dynamic"? for the sake of input mapping logic and what have you across different pointer types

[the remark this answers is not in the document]
domportera: in the case of removals/additions, others would need to be shifted to maintain persistence (for backends that track that). This would probably necessitate reporting a queue of changes to the lists, though id love to hear a better idea

documentation/proposals/Proposal - Axis Input Devices.md
Perksey: It’s more the transitioning between bounded vs unbounded for which we have the IPointerTargets. Basically this API is just observation of current state only (which is all you need most of the time) and there’s no way to set anything for a device.

documentation/proposals/Proposal - Axis Input Devices.md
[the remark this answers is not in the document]
Perksey: I guess we can probably make the index change implicit. Means more work for the consumer to deal with that but oh well.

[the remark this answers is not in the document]
domportera: ok my latest commit creates the `AxisState` struct, which has an `IsActive` bool that defaults to `true`. The idea being that dynamically populated axes stay allocated in the device's axis list, though they can be replaced in-place with no effect on the indices of other dynamic axes

documentation/proposals/Proposal - Axis Input Devices.md
```diff
@@ -0,0 +1,642 @@
...
+
+This is how a developer could subscribe directly to the axis changes of an `IAxisDevice`. It mimicks the `IGamepadInputHandler` interface from the [Multi-Backend Input proposal](./Proposal - Multi-Backend Input.md).
+
+```csharp
+
+/// <summary>
+/// Represents the state of a single axis
+/// </summary>
+/// <param name="IsActive">True if the axis is currently in use / actively updated (required to be true if axis is not <see cref="AxisTrait.Dynamic"/>)
+public readonly record struct AxisState(IAxisDevice Device, AxisDescription Description, float Value, bool IsActive = true);
```
Perksey, line 138: Should `IsActive` be in `AxisDescription`? Or more specifically, should this be queryable outside of a handler method call?

documentation/proposals/Proposal - Axis Input Devices.md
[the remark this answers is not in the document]
domportera: I added a draft of an idea to the end of the proposal

documentation/proposals/Proposal - Axis Input Devices.md
```diff
@@ -0,0 +1,720 @@
...
+    /// if the implementation utilizes <see cref = "OutputAxisTrait.RawDataOnly"/>.
+    ///<br/><br/>
+    /// It is more or less expected that this will have few, if any, implementations in the early stages of this API.
+    /// Its use is reserved for advanced use-cases or higher-level implementations where the caller knows the exact
+    /// data type required the output - for example, setting a specific bit pattern for a specific LED, driving
+    /// embedded displays, built-in speakers, etc.
+    /// </summary>
+    /// <param name="index">The index of the <see cref="OutputDescription"/> in <see cref="Outputs"/></param>
+    /// <param name="values">A pointer to the value(s) to be set</param>
+    public void SetOutput<T>(int index, T* values, int count = 1) where T : unmanaged => throw new NotImplementedException();
```
Perksey, line 682: Can we change this to be ReadOnlySpan<T>, perhaps with a DIM that implicitly creates a 1-element span from a single `ref readonly` parameter?

documentation/proposals/Proposal - Axis Input Devices.md
```diff
@@ -0,0 +1,720 @@
...
+
+/// <summary>
+/// The definition of an output of an <see cref="IAxisDevice"/>
+/// </summary>
+/// <param name="Index">The index of this output in <see cref="IAxisDevice.Outputs"/></param>
+/// <param name="AssociatedAxisIndex">An associated axis index if one exists - for example, force feedback for a
+/// specific trigger, an LED for a specific button, etc.</param>
+/// <param name="Name">An optional name for the output</param>
+/// <param name="CustomDataType">Valid (and required) only for outputs marked as <see cref="OutputAxisTrait.RawDataOnly"/></param>
+public readonly record struct OutputDescription(int Index, OutputAxisTrait Traits, int? AssociatedAxisIndex = null, string? Name = null, Type? CustomDataType = null);
```
Perksey, line 719: `AssociatedAxisIndex` should probably be `AssociatedAxes` and be a nullable `IReadOnlyList<int>` for consistency with the other structs. `CustomDataType` I don't think should be included in the description, and we should instead make this inferable from the traits.

documentation/proposals/Proposal - Axis Input Devices.md
```diff
@@ -0,0 +1,720 @@
...
+    ///<br/><br/>
+    /// It is more or less expected that this will have few, if any, implementations in the early stages of this API.
+    /// Its use is reserved for advanced use-cases or higher-level implementations where the caller knows the exact
+    /// data type required the output - for example, setting a specific bit pattern for a specific LED, driving
+    /// embedded displays, built-in speakers, etc.
+    /// </summary>
+    /// <param name="index">The index of the <see cref="OutputDescription"/> in <see cref="Outputs"/></param>
+    /// <param name="values">A pointer to the value(s) to be set</param>
+    public void SetOutput<T>(int index, T* values, int count = 1) where T : unmanaged => throw new NotImplementedException();
+}
```
Perksey, line 683: Come to think of it, we should probably add a `TryGetAxisInfo<T>` and a `TryGetOutputInfo<T>` so that we can have extensions in the future.

documentation/proposals/Proposal - Axis Input Devices.md
```diff
@@ -0,0 +1,720 @@
...
+{
+    None = 0,
+    LED = 1u << 0,
+    Haptic = 1u << 1,
+    
+    LEDBrightness = 1 << 4 | LED,
+    LEDRed = 1u << 5 | LED,
+    LEDGreen = 1u << 6 | LED,
+    LEDBlue = 1u << 7 | LED,
+    LEDWhite = 1u << 8 | LED,
```
Perksey, line 695: Kind of feels like we need output groups as well. Which I guess makes sense from a consistency standpoint.

domportera: it does feel that way huh. could get rid of all these different sub-types of LED in favor of an LED OutputAxisGroupTypeNamePending

documentation/proposals/Proposal - Axis Input Devices.md
domportera: wouldnt that only apply to "OutputAxisGroup"s? this is more like `TwinIndex` in `AxisDescription`

Perksey: Yes now that we have decided that axis groups are going to exist. These comments did not have knowledge of eachother seemingly 😅

domportera: > `CustomDataType` I don't think should be included in the description, and we should instead make this inferable from the traits

I also don't like it, but it was just off-the-dome in the last few minutes. all in all im not sure how to handle the typing for this sort of thing in a graceful way atm

documentation/proposals/Proposal - Axis Input Devices.md
domportera: this could be a nicer way to handle the typing - kind of enforces the idea that axis devices should strive to use similar output types where possible. I wonder how this would extend to the income "OutputGroup"

domportera: that being said, i think that sort of api needs to be more thoroughly thought out than the next 30 minutes would allow us

Perksey: I think I'd be happy defining these two methods and just having no acceptable types for T atm i.e. Silk.NET reserves the right to add device-type-specific information to aid in axis identification, mapping, configuration, etc. Use of these APIs shouldn't be encouraged as the happy path, but it's sometimes useful to get more info than our intentionally over-generalised API provides.

documentation/proposals/Proposal - Axis Input Devices.md
domportera: funny i actually changed it from a span to a pointer when writing this - happy to change it back to appease my paranoid ass

documentation/proposals/Proposal - Axis Input Devices.md
Perksey: Basically I think we should just have a very similar API surface for outputs as we do axes - we worked hard on making that generalised abstraction and I like the trait system that you came up with e.g. if it's an LED and it has 3 axes you can assume they're RGB etc

documentation/proposals/Proposal - Axis Input Devices.md
Perksey: Would recommend replacing with my other comment i.e. the try get info

documentation/proposals/Proposal - Axis Input Devices.md
domportera: I like the idea of putting it in `AxisDescription`, though I'd really like to avoid the scenario where a "fixed" axis is marked as "IsActive = false" post-validation. currently theres no defense against this anyway, but it could create ideas.

though it could be a nice way to say "hey my X button is broken, can we not?" so in that case im actually gonna update this to be in the AxisDefinition and allow IsActive (w/ potential rename) to be used universally. we're probably never going to do that, but ¯\_(ツ)_/¯

documentation/proposals/Proposal - Axis Input Devices.md
Perksey: I think if we're splitting this up into outputs vs output groups then we should likely have a single output be a single float or a single T (with the T path being try-based to enable extension detection, again we should reserve the right to define these extensions as we see fit given that they're not the happy path, and that not everything is cleanly representable as a float.

Perksey:
- We need to discuss this further another time (primarily because of the last minute additions like outputs that naturally need more thought), but we reviewed the majority and were happy with most of what we saw, with the following comments...
- `AsAvailable` isn't needed as C# has `with`
- Use radians instead of degrees for EulerAngleComponents
- Use sequential bits for sanity, or otherwise use a more defined bit usage scheme for the bitmasks ("enum sandwich")
- Use pitch, yaw, roll instead of ypw
- Some consideration has to be done for the diamond layouts and region-specific mappings of the buttons therein.

Perksey (approved):
Merging as WIP. We'll come back to this proposal once we've had a chance to play around with the abstractions proposed - this proposal not being proposed at this time needn't preclude us having private implementations to work out the details which we can then later publicise - the way I see it, this is still the proposal process fulfilling its purpose as even if that isn't Dom, this proposal serves as enough of an idea that others can take over later.

PR #2419, opened 2025-02-03
Fix aar duplication for Android

Perksey:
Android applications currently can't build due to type duplication issues. It turns out we've been shipping the same AAR twice and haven't noticed until now.


Perksey:
Confirmed working https://discord.com/channels/521092042781229087/607634593201520651/1335953176901324842

Beyley (approved):

PR #2423, opened 2025-02-11
[3.0] Update to SDL 3.2.4

Perksey:
Does what it says on the tin. Updates native libs and bindings.


Perksey:
@curin Ready for review. As you can see from the commit history this has in fact been built, I just haven't fixed the bug where the PR mandatory checks don't run on commits that were generated from a check.

curin (approved):

Perksey:
@dotnet-policy-service rerun

PR #2427, opened 2025-02-15
[3.0] Input Rewrite

Perksey:
Implements the now-approved proposal.


sources/Input/Input/Implementations/SDL3/SdlInputBackend.Devices.cs
```diff
@@ -0,0 +1,110 @@
...
+        _ = UnregisterDevice(device.Id);
+
+        // device IDs may have changed when a device was removed, so we need to refresh them
+        RefreshDeviceIds(devices);
+        return true;
+    }
+
+    private bool UnregisterDevice(nint uniqueId)
+    {
+#if DEBUG
```
Perksey: The code on both sides of these `#if`s are exactly the same once compiled by virtue of the `InputLog` calls being `Conditional`.

PR #2433, opened 2025-03-03
[3.0] Update SDL to 3.2.6

joskuijpers:
I ran the update.sh script to update the commit ref and version.txt file.

3.2.6 contains a whole host of ABI compatible fixes: https://github.com/libsdl-org/SDL/releases/tag/release-3.2.6

I'd guess this also needs updates to the binaries. Is there an action doing that?

/build-native sdl


joskuijpers:
Done. but it is unclear to me which binaries to take for the simulators. The iOS and tvOS build both have a simulator binary, and so do the iosSimulator and tvossimulator builds.

Exanite:
This is superseded by https://github.com/dotnet/Silk.NET/pull/2570, which has been merged.
I'll also look into improving the native build workflows.

Please let me know if you have any questions or concerns.

PR #2439, opened 2025-04-04
Use UTF-8 for OpenAL string marshalling

Morilli:
`Marshal.PtrToStringAnsi` uses the system codepage which is problematic especially on windows. OpenAL returns UTF-8 strings which should be parsed as UTF-8.

Using the following test code:
```csharp
var api = ALContext.GetApi();
api.TryGetExtension<EnumerateAll>(null, out var enumerateAll);
var deviceEnumerable = enumerateAll.GetStringList(GetEnumerateAllContextStringList.AllDevicesSpecifier);
foreach (var deviceName in deviceEnumerable)
{
    Console.WriteLine(deviceName);
}
```

with a non-utf-8 codepage and a device named `test ドットネット`, I get:

before:
`OpenAL Soft on test ãƒ%ãƒƒãƒ^ãƒ?ãƒƒãƒ^ (Realtek High Definition Audio)`

after (with proper console encoding configured):
`OpenAL Soft on test ドットネット (Realtek High Definition Audio)`

### DIFF
--- a/src/OpenAL/Extensions/Silk.NET.OpenAL.Extensions.Creative/EnumerateAll.cs
+++ b/src/OpenAL/Extensions/Silk.NET.OpenAL.Extensions.Creative/EnumerateAll.cs
@@ -3,7 +3,8 @@
 
 using System;
 using System.Collections.Generic;
-using System.Runtime.InteropServices;
+using System.Linq;
+using System.Text;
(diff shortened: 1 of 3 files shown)


Perksey (approved):
Thank you so much!

PR #2457, opened 2025-05-23
[3.0] Add Vulkan bindings for Silk 3.0

Exanite:
# Current progress

(Will update this as I work)

Bindings now compile and work!

See bottom of PR for current progress since I'm now actively tracking todos here.

# Summary of the PR

This PR primarily adds Vulkan bindings, but also many smaller fixes/changes to the overall SilkTouch codebase. See the rest of the PR for more information.

Initial Discord discussion: https://discord.com/channels/521092042781229087/587346162802229298/1375224188129906778

Discord thread: https://discord.com/channels/521092042781229087/1376331581198827520/1376623544875880589

# Tasks done / todos

Before undrafting PR

- [x] Self code review and make sure all important changes/decisions are documented in the PR summary.

---

Before merging PR

- [x] Consider committing the Windows "version" of the Vulkan bindings. Generating the bindings on different platforms leads to different results. In general, which platform should we use to generate each of the bindings?
   - Decided not to since the bindings generation for Vulkan is pretty much consistent now. Only the NativeTypeName attributes on enums differ. 

---

Future PRs

- [x] Change naming convention to not capitalize acronyms
   - See: https://discord.com/channels/521092042781229087/1376331581198827520/1395097379329282089 
- [x] The `Vk` in `PFNVkDebugUtilsMessengerCallbackEXT` is not trimmed properly. Same with `BufferTHandle`.
   - Will be fixed when we switch to the new prettification approach (save native name early on and replace native name at end)
   - The `BufferTHandle` case was fixed in this PR, ~~but reverted~~ since the new approach will fix it in a more general way.
   - [x] Revert "hacky" THandle fix.
- [x] Consider renaming handle types as `AccelerationStructureHandleKHR`, with the vendor suffix at the very end instead of being before `Handle`.
- [x] Fix `StdVideoEncodeH264SliceHeaderFlags` property names. They currently aren't getting prettified.
- [x] Apply BoolType transformation to VkBool32 in structs. They currently are only handled for functions.
- [x] Structs don't have constructors unlike Silk 2. I don't know if this was an intentional change or if this functionality simply hasn't been properly implemented yet.
   - Conclusion: The constructors were for setting default field values, but we can set default field values directly now.
   - [ ] Add default field values to structs where it makes sense
- [ ] Ensure attributes are correct
   - [ ] SupportedApiProfiles
      - [ ] Properly resolve API profiles for Flags/FlagBits types
   - [x] ~~NativeTypeName~~ NativeName (Storing the native name is more useful than the type name for bindings consumers)
- [ ] Reimplement the struct chaining API
- [x] Update Curin's branch to use my new LocationTransformation code. The new renamer should be replaced with this. See the NameUtils.cs and Renamer.cs files on Curin's branch. Renamer.cs should be removed. NameUtils.cs should have the methods updated to use my LocationTransformation code instead (this should show up as a merge conflict).

### Done

This list should hopefully provide some context on decisions made during this PR. If anything is unclear, please feel free to ask me here or in Discord.

This list is up-to-date with the newest tasks at the top.

Note that I changed LocationTransformation and TransformHandles over the course of the PR.

LocationTransformation is my renaming replacement that allows for multiple modifications to be applied to all references of a symbol without needing to search for references multiple times. As it does not use FindReferencesAsync, it also is much faster.

TransformHandles originally was rewritten to have empty handle structs be added in ExtractNestedTyping, but reverted that and TransformHandles mostly behaves as it originally did, just with a new symbol-based implementation. The original change was so that types like `Buffer_T` had the `_T` suffix removed during prettification before the name has the `-Handle` suffix added during TransformHandles. Because this has been reverted, the current output is `BufferTHandle`, but that will be fixed with the new prettification approach (save native name early on and replace native name at end).

Edit: I removed the _T suffix using a hacky 1 line fix. I'll remove this in my next PR after the prettification change is made.

Renamer

- [x] Update my LocationTransformer code to work with changes made in curin's PR: https://github.com/dotnet/Silk.NET/pull/2406/commits/0cc49b84d4c84a43c12816e93ae861a22570347c#diff-4d93e56d6d07a37ac3577a7fa2dfd0f0649e9ac07912045809c723351c9be56c
   - Curin is now aware of this code and is working with it
   - Note: This code will be removed before this PR is merged. This is because we decided to use a different renaming approach.
- [x]  Update the Vulkan job config to match other jobs
   - [x] ExtractNestedTyping and TransformHandles
   - [x] PrettifyNames mod order
- [x] Update `TransformHandles` to account for new prettification approach (save native name early on and replace native name at end)
- [x] Copy over the renamer changes from Curin's Microsoft branch.
- [x] Rewrite TransformHandles to work with the new renamer.

Account for changes from OpenAL branch

- [x] Merge in changes from OpenAL branch
- [x] Use `InterceptNativeFunctions` instead of `TransformVulkan`
- [x] Update Vk interfaces to expose CurrentDevice, CurrentInstance, etc

ClangSharp

- [x] Update to .NET 10 and update ClangSharp so my fixes are pulled in.
- [x] The two ClangSharp tasks are completed on the ClangSharp side. I just need to update ClangSharp and remove the enum type cast workaround.
- [x] Fix generation of bitshifts
   - Eg: `(((uint)(0)) << 29U) | (((uint)(1)) << 22U) | (((uint)(0)) << 12U) | ((uint)(0))`
   - I'm fixing this in ClangSharp: https://github.com/dotnet/ClangSharp/issues/611
- [x] Fix missing type casts between bitfield backing type and enums
   - I have an issue open for this in ClangSharp: https://github.com/dotnet/ClangSharp/issues/618
   - I currently am using a hack to fix this behavior. The issue above describes this fix. 
- [x] Fix `PFNVkDebugUtilsMessengerCallbackEXT`'s second parameter being an uint instead of DebugUtilsMessageTypeFlagsEXT
   - Probably the same issue with the other Flags/FlagBits types.

Library loading

- [x] Fix `Unable to load DLL 'vulkan'...` on Windows
   - Vulkan has different shared library names on different platforms. How do we handle this?
   - See: https://github.com/dotnet/Silk.NET/blob/main/generator.json#L529-L536
   - Fix was to use `LoaderInterface.RegisterAlternativeName` in a static ctor in the `Vk` class.

VTables

Note: This section of tasks has been replaced using InterceptNativeFunctions mod from the OpenAL branch.

- [x] Implement Vulkan vtable function pointer management
   - ~~Not exactly sure what I need to do here yet. I'll do some research.~~
   - [x] Code generation for instance/device members
   - [x] Implement custom function pointer loading
      - [x] Clone method
      - [x] Create method

---

Bitfield structs

- [x] Consider transforming bitfield structs into enums where possible
   - Eg: `StdVideoEncodeH264SliceHeaderFlags`
   - Decision: We're keeping these as bitfield structs.

---

Bools

- [x] `SamplerCreateInfo.UnnormalizedCoordinates`, `SamplerCreateInfo.CompareEnable`, etc uses `uint` instead of a `bool`-like type.
   - [x] (Discuss) Is this related to the MaybeBool type? I haven't looked at the rest of Silk 3 yet. If so, then this might be something that can be fixed separately from the PR. 
   - The SDL bindings already use MaybeBool, so the functionality already exists. I'll look into how to use it for Vulkan.
   - I implemented this by adding a remapping in the rsp files (`VkBool32=MaybeBool<uint>`).

---

Deprecated aliases

- [x] (Discuss) `VK_STRUCTURE_TYPE_SURFACE_CAPABILITIES2_EXT` (alias of `VK_STRUCTURE_TYPE_SURFACE_CAPABILITIES_2_EXT`), `VK_KHR_MAINTENANCE1_SPEC_VERSION`, etc are deprecated aliases that cause compilation issues due to name conflicts. These are currently manually excluded.
   - These are marked with the `deprecated="aliased"` attribute in the XML spec 
   - [x] Automate the list of type exclusions
   - Decision: I'm going to implement this because these names tend to cause conflicts during name prettification and because there is always an obvious alternative available. This can lead to more upstream breakages, but fixing these errors are trivial. Removing these also reduces confusion about which one to use.

---

MaxEnum members

- [x] MaxEnum enum member removal also needs to consider `INVALID` members (Vulkan Video)
   - This is trickier because we can't just indiscriminately remove the `INVALID` members when the name contains the string
- [x] MaxEnum member removal skips over enums not read from the XML
   - This includes both the Vulkan Video enums (there is a XML spec, but we don't read it) and the enums that are empty
 in the Vulkan XML spec (the XML spec is out of date/out of sync) 
   - This also impacts the adding of `[Flags]` attributes.

I ended up making it so that the `TransformEnums` mod handles `MaxEnum` removal instead of `MixKhronosData`. This is because the max enum value pattern applies to C as a whole, rather than just Vulkan or Khronos.

- [x] (Discuss) MixKhronosData ignores empty enums/groups from the XML. Would it be better to generate empty enums for these instead of completely ignoring them?
   - This is due to the group data only being initialized if there is at least one group member (search for `data.Groups[group] = data.Groups.TryGetValue(group, out var groupInfo)` in MixKhronosData and read the surrounding code for more context)
   - Because this seems like an implementation detail / oversight to me, I want to make sure that this is intentional.
   - Decided to fix this since it affects the correctness of other transformations.

I also made it so that MixKhronosData does not skip over empty enum groups, which fixes the generation of `[Flags]` attributes and leads to a few more enums being generated.

---

Trimming

- [x] `ComponentSwizzle.ComponentSwizzleIdentity`, etc is not trimmed properly. The enum name isn't getting trimmed here for some reason.
- [x] `ShaderStageFlags.ShaderStageVertexBit`, etc is not trimmed properly
   - ~~Guess: ShaderStageFlags is not a prefix of ShaderStageVertexBit~~ Incorrect. This was the same issue as above.
- [x] Fix `PipelineCompilerControlFlagsAMD.FlagBitsMaxEnumAMD`, etc trimming
   - (Discuss) Didn't realize it when I first wrote this, but these are the `MaxEnum` members. We don't want these do we?
   - Note to self: ~~Fix trimming first, then remove these members.~~ The `MaxEnum` members are the only ones that are named like this in the headers so we can just remove them. This isn't a trimming issue.
- [x] Removing the `MaxEnum` members breaks `IndirectStateFlagsNV.FlagFrontfaceBitNV` due to no prefixes being shared by member names
   - [x] Exhaustive list: `IndirectStateFlagsNV`, `SemaphoreImportFlags`, `SemaphoreWaitFlags`, `SurfaceCounterFlagsEXT`

---

Linux

- [x] Remove Linux specific Clang workaround in `eng/silktouch/vulkan/settings.rsp`
   - I needed to add `/usr/lib/clang/20/include` as an include directory. `20` is the Clang version. 
- [x] Discuss longterm solution for handling the Clang include directory. I know there is `UnixStdIncludeResolver`
   - I added support for finding the `/usr/lib/clang/*/include` folder to UnixStdIncludeResolver. It finds all matching folder patterns and takes the latest version. If my current implementation is okay, then we can consider this task done.
- [x] Add a `remap-stdint.rsp` file specific to Vulkan bindings generation to fix https://github.com/dotnet/ClangSharp/issues/574

---

Handling of Flags enums.

- [x] (Fixed) `DebugUtilsMessageSeverityFlagsEXT` is not marked with the `[Flags]` attribute.
   - [x] (Discuss) Looks like none of the Flags enums are marked with `[Flags]`. Does this functionality exist in Silk 3 yet?
      - Doesn't seem like it
   - Implemented by checking `KnownBitmask` in `MixKhronosData.Rewriter`.
- [x] (Discuss) `DependencyFlags.None` does not exist, but does in Silk 2. None does not exist in the Vulkan headers nor XML spec, so I'm assuming None was added by Silk itself.
   - [x] Also `ShaderStageFlags.None` 
   -  ~~I attempted to fix this, but led to some unwanted outputs, such as names not being trimmed properly (does the trimmer check to see if all enum members share a prefix?). Also, knowing when to add in the `None = 0` member requires evaluating the value of the existing enum members. This is somewhat difficult because they look like `unchecked((ulong)0UL)`, which isn't trivially comparable to 0.~~
   - ~~This might fit better in a TransformFlags mod so we can have this behavior in all Silk bindings.~~
   - ~~It looks like we can use `IFieldSymbol.ConstantValue` for this~~
   - Implemented by using `semanticModel.GetConstantValue` in a new `TransformEnums` mod.

---

Incorrect typing for Flags/FlagBits enums

Original comment: ~~These two issues are Flags/FlagBits specific issues. This disconnect causes ClangSharp to just output ulongs/uints. This is likely something we should fix on our side. We can fix this by looking at the NativeTypeName attribute.~~

Original comment: ~~Some are trivial, but some NativeTypeNames look like `const VkPipelineStageFlags *`, such as in SubmitInfo.~~

- [x] `MemoryBarrier2.SrcStageMask` does not use the `AccessFlags2` enum and uses ulongs instead.
- [x] `DependencyInfo.DependencyFlags` does not use the `DependencyFlags` enum and uses uint instead.
   - [x] Also `ImageSubresourceRange.AspectMask`
   - [x] Also `CommandPoolCreateInfo.Flags`
   - [x] Also `FenceCreateInfo.Flags`
   - [x] Also `ClearAttachment.AspectMask`
   - ...etc

This was fixed by parsing the type name from the NativeTypeName attributes and replacing the field/parameter type with the parsed type name.

---

Renaming of FlagBits to Flags.

Original comment: ~~I recently found code that handles the FlagBits -> Flags mapping in MixKhronosData. I think the issue is that this code doesn't fully work yet so I'm looking into seeing how that code works and fixing that code.~~

- [x] Remove code used to generate ClangSharp enum type remappings in `sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs`.
- [x] (Not implemented as originally planned) ~~Automate the generation of ClangSharp enum type remappings~~
   -  Actual implementation: Don't remap using ClangSharp. Use MixKhronosData.Rewriter to replace `FlagBits` with `Flags` in identifier names globally.

---

Missing constants

 - [x] (Fixed) Not sure where `VK_LOD_CLAMP_NONE` and other constants went.
    - Was due to the fake "API Constants" enum defined in the XML spec. Because of this, the constants were getting removed by MixKhronosData.

---

Generated function class should be named `Vk`

- [x] Consider renaming the Vulkan class to Vk to match Silk 2
   - I decided to go ahead and rename it to Vk 

---

Document handling

- [x] ~~Change MSBuildModContext to only remove project documents that are part of the output directory. This is to ensure that the global usings files are not removed.~~
   - See https://discord.com/channels/521092042781229087/1376331581198827520/1381026216466845716
   - I reverted this when I merged in the OpenAL changes since this was causing errors and is not strictly needed. The original reason for this was to reduce the number of errors returned by `compilation.GetDiagnostics` in TransformHandles.
- [x] Change ModCSharpSyntaxRewriter to exclude global usings when injecting using statements.
   - This is in response to the MSBuildModContext change.
   - See https://discord.com/channels/521092042781229087/1376331581198827520/1391155465328529491

---

TransformHandles and LocationTransformation

- [x] Add missing handle type discovery and empty struct generation for missing handle types to ExtractNestedTyping
- [x] ~~Rewrite TransformHandles to only transform existing empty structs. It will no longer generate new empty structs.~~
   - Outdated: This change has been reverted since we are going with a different prettification approach. This change was originally to handle the `_T` suffix on Vulkan's handle types.
- [x] Add the `Silk.NET.SilkTouch.Mods.LocationTransformation` namespace. This contains code that allows a set of transformations to be applied to the locations returned by ~~SymbolFinder.FindReferencesAsync~~. This is currently used for renaming and for pointer dimension reduction (eg: replacing `Type*` with `Type`).
   - LocationTransformation now does not use FindReferencesAsync for performance reasons and instead gathers and checks symbols manually. This is a continuation of the non-FindReferencesAsync renamer I implemented on Curin's branch which saves multiple hours per run.


Exanite:
Putting this here for reference.

These are the logs related to the generation of enums during the MixKhronosData mod after the latest commit.

The enums specified in these errors either are:
- Not enums (API Constants)
- Are part of a non-core Vulkan header
    - VkFullScreenExclusiveEXT is part of vulkan_win32.h
    - VkDisplacementMicromapFormatNV is part of vulkan_beta.h
- Or don't exist at all in the headers in the Vulkan-Docs nor Vulkan-Headers repos

```
fail: Silk.NET.SilkTouch.Mods.MixKhronosData[0] Enum "API Constants" has no base type. Please add TypeMap entry to the configuration. This enum group will be skipped.
fail: Silk.NET.SilkTouch.Mods.MixKhronosData[0] Enum "VkFullScreenExclusiveEXT" has no base type. Please add TypeMap entry to the configuration. This enum group will be skipped.
fail: Silk.NET.SilkTouch.Mods.MixKhronosData[0] Enum "VkFaultLevel" has no base type. Please add TypeMap entry to the configuration. This enum group will be skipped.
fail: Silk.NET.SilkTouch.Mods.MixKhronosData[0] Enum "VkFaultType" has no base type. Please add TypeMap entry to the configuration. This enum group will be skipped.
fail: Silk.NET.SilkTouch.Mods.MixKhronosData[0] Enum "VkFaultQueryBehavior" has no base type. Please add TypeMap entry to the configuration. This enum group will be skipped.
fail: Silk.NET.SilkTouch.Mods.MixKhronosData[0] Enum "VkPipelineMatchControl" has no base type. Please add TypeMap entry to the configuration. This enum group will be skipped.
fail: Silk.NET.SilkTouch.Mods.MixKhronosData[0] Enum "VkSciSyncClientTypeNV" has no base type. Please add TypeMap entry to the configuration. This enum group will be skipped.
fail: Silk.NET.SilkTouch.Mods.MixKhronosData[0] Enum "VkSciSyncPrimitiveTypeNV" has no base type. Please add TypeMap entry to the configuration. This enum group will be skipped.
fail: Silk.NET.SilkTouch.Mods.MixKhronosData[0] Enum "VkPipelineCacheValidationVersion" has no base type. Please add TypeMap entry to the configuration. This enum group will be skipped.
fail: Silk.NET.SilkTouch.Mods.MixKhronosData[0] Enum "VkDisplacementMicromapFormatNV" has no base type. Please add TypeMap entry to the configuration. This enum group will be skipped.
```

sources/Core/Core/Silk.NET.Core.csproj
```diff
@@ -22,13 +22,11 @@
   <!-- Public API Tracking -->
   <PropertyGroup>
     <SilkTrackPublicAPI>true</SilkTrackPublicAPI>
+    <TreatWarningsAsErrors>false</TreatWarningsAsErrors>
```
Exanite: Probably re-enable TreatWarningsAsErrors before merging

Exanite:
Going to wait to self code review this until the OpenAL bindings are merged. Currently the diffs are combined making it hard to see what I personally changed.

sources/SilkTouch/SilkTouch/Mods/ExtractNestedTyping.cs
```diff
@@ -1,48 +1,37 @@
...
-/// </description>
-/// </item>
-/// <item><description>Fixed buffers and anonymous structures contained within structures.</description></item>
+/// </description></item>
+/// <item><description>
+/// Extracting fixed buffers and anonymous structures contained within structures into separate types.
+/// </description></item>
 /// </list>
 /// </summary>
 public partial class ExtractNestedTyping(ILogger<ExtractNestedTyping> logger) : Mod
```
Exanite, line 37: I used to have code in ExtractNestedTyping to identify handle types and to generate empty handle structs, but that got removed once I reverted most of my TransformHandles changes. What remains is just the cleanup and comments I left when working with ExtractNestedTyping.

sources/Maths/Maths/Silk.NET.Maths.csproj
```diff
@@ -1,14 +1,15 @@
...
-        <TargetFramework>net8.0</TargetFramework>
+        <TargetFramework>net10.0</TargetFramework>
         <Nullable>enable</Nullable>
         <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
         <GenerateDocumentationFile>true</GenerateDocumentationFile>
         <DefineConstants>SSE;BTEC_INTRINSICS;MATHF;POH;AdvSIMD;$(DefineConstants)</DefineConstants>
         <EnableNETAnalyzers>true</EnableNETAnalyzers>
         <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
         <SilkDSLExempt>true</SilkDSLExempt>
+        <TreatWarningsAsErrors>false</TreatWarningsAsErrors>
```
Exanite: Might want to revert this. Search for similar changes as well.

sources/Core/Core/PublicAPI/net10.0/PublicAPI.Shipped.txt
```diff
@@ -0,0 +1 @@
+#nullable enable
```
Exanite, line 1: This needs updating

sources/Core/Core/PublicAPI/net10.0/PublicAPI.Unshipped.txt
```diff
@@ -0,0 +1 @@
+#nullable enable
```
Exanite, line 1: This needs updating

eng/silktouch/vulkan/header.txt
```diff
@@ -0,0 +1,5 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+// Ported from the Vulkan headers and corresponding dependencies.
+// Original source is Copyright 2015-2023 The Khronos Group Inc. Licensed under the MIT license.
```
Exanite: Is this header correct? For reference: https://github.com/KhronosGroup/Vulkan-Headers/blob/main/LICENSE.md

sources/OpenAL/Soft.Native/runtimes/win-x64/native/soft_oal.dll
Exanite, line 1: Not sure why these OpenAL binaries are showing up in the diffs, I should have accepted the ones from the main branch. Consider reverting to main's version.

sources/OpenGL/OpenGL/Enums/FogParameter.gen.cs
```diff
@@ -2,6 +2,8 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // Ported from the OpenGL headers and corresponding dependencies.
 // Original source is Copyright 2013-2020 The Khronos Group Inc. Licensed under the MIT license.
+using System;
```
Exanite, line 5: The committed bindings (not just OpenGL) in general are not _exactly_ up to date, at least when I regenerated them on Linux, but the changes are very minor and are mainly using statement/platform differences. I don't know which platform each binding should be generated on, so I didn't commit the changes.

sources/OpenGL/OpenGL/Enums/GLEnum.gen.cs
```diff
@@ -899,10 +899,10 @@ public enum GLEnum : uint
     CompressedRed = unchecked((uint)0x8225),
     CompressedRg = unchecked((uint)0x8226),
     ContextFlagForwardCompatibleBit = unchecked((uint)0x00000001),
-    RGBA32F = unchecked((uint)0x8814),
-    RGB32F = unchecked((uint)0x8815),
-    RGBA16F = unchecked((uint)0x881A),
-    RGB16F = unchecked((uint)0x881B),
+    Rgba32F = unchecked((uint)0x8814),
```
Exanite, line 902: Will address in a future PR by capitalizing all acronyms. Don't remember why these changed though.

sources/OpenGL/OpenGL/Enums/SyncObjectMask.gen.cs
```diff
@@ -2,14 +2,17 @@
...
 
 namespace Silk.NET.OpenGL;
 
 [Transformed]
+[Flags]
 public enum SyncObjectMask : uint
 {
-    Bit = unchecked((uint)0x00000001),
-    BitApple = unchecked((uint)0x00000001),
+    FlushCommandsBit = unchecked((uint)0x00000001),
```
Exanite, line 16: I added a prefix override for this so this doesn't get shorted to just Bit.

sources/SilkTouch/SilkTouch/Mods/AddVTables.cs
```diff
@@ -919,7 +919,6 @@ out var callConv
                             )
                             : null
                     )
-                    .WithAttributeLists(List<AttributeListSyntax>())
```
Exanite, line 922: This was clearing the attributes on some of the generated VTable methods. Which attributes, uhh, I don't remember, but I think it was the NativeFunction attributes.

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
```diff
@@ -295,143 +337,122 @@ .. currentConfig.NonStandardExtensionNomenclature
...
+                    WithGuids = rsp.GeneratorConfiguration.WithGuids,
+                    WithLibraryPaths = rsp.GeneratorConfiguration.WithLibraryPaths,
+                    WithManualImports = rsp.GeneratorConfiguration.WithManualImports,
+                    WithNamespaces = rsp.GeneratorConfiguration.WithNamespaces,
+                    WithSetLastErrors = rsp.GeneratorConfiguration.WithSetLastErrors,
+                    WithSuppressGCTransitions = rsp.GeneratorConfiguration.WithSuppressGCTransitions,
+                    WithTransparentStructs = rsp.GeneratorConfiguration.WithTransparentStructs,
+                    WithTypes = rsp.GeneratorConfiguration.WithTypes,
+                    WithUsings = rsp.GeneratorConfiguration.WithUsings,
+                    WithPackings = rsp.GeneratorConfiguration.WithPackings,
```
Exanite: Do we want to wrap this with some sort of wrapper record class so we can just use `with` instead of copying everything manually?

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
```diff
@@ -1740,16 +1783,23 @@ public bool TryGetSymbolMetadata(
...
+        }
+
+        if (!job.SupportedApiProfiles.TryGetValue(nativeName, out var mdList))
+        {
+            metadata = null;
+            return false;
+        }
+
+        metadata = mdList;
+        return true;
```
Exanite, line 1793: Same behavior as before, but I rewrote it when debugging since the original was hard to debug and understand.

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
```diff
@@ -1975,6 +2300,19 @@ var group in (groupName is null ? Enumerable.Empty<string>() : [groupName])
...
+            {
+                data.Groups[groupName] = new EnumGroup(
+                    groupName,
+                    null,
+                    [],
+                    isBitmask,
+                    VendorFromString(groupName, vendors),
+                    enumNamespace
+                );
+            }
```
Exanite, line 2291: If I recall, this broke some of the tests since it leads to more enum groups being output.

sources/SilkTouch/SilkTouch/Mods/TransformEnums.cs
```diff
@@ -0,0 +1,221 @@
...
+
+    /// <summary>
+    /// Represents a filter used to match enum members.
+    /// </summary>
+    public record EnumMemberFilterConfiguration
+    {
+        /// <summary>
+        /// The enum type name must match this regex.
+        /// </summary>
+        public string TypeName { get; init; } = ".*";
```
Exanite, line 89: Should I rename this and similar properties to `TypeNameRegex` to clarify that regex is expected?

Exanite (commented):
Self code review complete. My comments are a mix of minor todos, comments that provide context, and questions.

eng/silktouch/vulkan/header.txt
Perksey: It's dual-licensed under MIT and Apache, so it's technically correct, notwithstanding the end date of course.

eng/silktouch/vulkan/remap.rsp
```diff
@@ -0,0 +1,2 @@
+@../remap-stdint.rsp
+VkBool32=MaybeBool<uint>
```
Perksey: Ideally this should remap to `uint` and then we add `VkBool32` to `BoolTypes` for `TransformFunctions`

sources/OpenGL/OpenGL/Enums/FogParameter.gen.cs
Perksey: The bindings shouldn't care what platform they're generated on in general, and if this is the case then this is a failing to `remap` those platform-dependent native types to something sensible. The only exception is of course the Windows SDK bindings.

sources/SilkTouch/SilkTouch/Clang/UnixStdIncludeResolver.cs
```diff
@@ -18,7 +16,30 @@ public IEnumerable<string> GetStandardIncludes() =>
     /// <inheritdoc cref="GetStandardIncludes" />
     protected virtual IEnumerable<string> CoreGetStandardIncludes()
     {
+        // Add standard include directories
         yield return "-I/usr/include";
         yield return "-I/usr/local/include";
+
+        // Attempt to find and add clang include folder
```
Perksey, line 23: Should we take account of the clang version we're using? (as made available on `clang.getVersion` iirc)

sources/SilkTouch/SilkTouch/Mods/TransformEnums.cs
Perksey: I'm not particularly bothered, if the user didn't know this is a regex they'd treat it as an exact match filter which I believe would still work just fine. Eventually we'll get documentation generating (and we need to write user guide docs for SilkTouch at some point anyway)

sources/SilkTouch/SilkTouch/Mods/TransformEnums.cs
```diff
@@ -0,0 +1,221 @@
...
+
+    private class EnumMemberFilter
+    {
+        public Regex TypeName { get; }
+        public Regex MemberName { get; }
+        public long? MemberValue { get; }
+
+        public EnumMemberFilter(EnumMemberFilterConfiguration configuration)
+        {
+            TypeName = new Regex(configuration.TypeName);
```
Perksey: Probably want to use RegexOptions.Compiled on these given they'll be hit frequently

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
```diff
@@ -295,143 +337,122 @@ .. currentConfig.NonStandardExtensionNomenclature
...
         }
+
+        // Add mappings from Flags storage types to FlagBits enums types
+        // We want Flags as the output, but in order to generate everything correctly,
+        // we need to first map Flags to FlagBits, then rename FlagBits back to Flags (this is done in the rewriters).
+        // Without this, ClangSharp will output the Flags types as the integral storage type instead of the enum
+        foreach (var typeElement in xml.Elements("registry").Elements("types").Elements("type"))
+        {
+            var typedef = typeElement.Element("type")?.Value;
+            if (typedef != "VkFlags" && typedef != "VkFlags64")
```
Perksey: Not sure this should be hardcoded. This should probably be configurable with these as sensible defaults.

eng/silktouch/vulkan/header.txt
Exanite: Alright. Let me know if you want any changes.

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
Perksey: Would probably make sense!

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
```diff
@@ -636,7 +657,6 @@ .. profileVariations.TryGetValue(variant, out var v) ? v : [],
                         _listSeparators,
                         StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries
                     ) ?? [];
-            Debug.Assert((depends.Length > 0) == explicitDependencies);
```
Perksey, line 639: Why has this been removed?

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
```diff
@@ -1221,6 +1241,46 @@ public void Trim(NameTrimmerContext context)
             );
         }
 
+        // NameTrimmer trims member names by looking for a common prefix and removing it
```
Perksey, line 1236: How come this logic was moved up?

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
```diff
@@ -1813,33 +1997,171 @@ private class Rewriter(JobData job) : CSharpSyntaxRewriter(true)
             }
 
             var nativeName = node.AttributeLists.GetNativeTypeName();
-            if (nativeName is null || !nativeName.StartsWith("#define "))
+            switch (nativeName)
```
Perksey: This particular bit of code should probably be a common function

sources/Vulkan/Vulkan/Handles/BufferTHandle.gen.cs
```diff
@@ -0,0 +1,32 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// Ported from the Vulkan headers and corresponding dependencies.
+// Original source is Copyright 2015-2023 The Khronos Group Inc. Licensed under the MIT license.
+using System;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+namespace Silk.NET.Vulkan;
+
+public readonly unsafe partial struct BufferTHandle
```
Perksey: > Because this has been reverted, the current output is BufferTHandle, but that will be fixed with the new renamer approach (save native name early on and replace native name at end

This is in the "done" section, but the committed bindings still have the T suffix. Is this going to be fixed in a future PR?

sources/OpenGL/OpenGL/Enums/FogParameter.gen.cs
Perksey: This comment also covers:
>  Consider committing the Windows "version" of the Vulkan bindings. Generating the bindings on different platforms leads to different results. In general, which platform should we use to generate each of the bindings?

sources/Vulkan/Vulkan/Vulkan/CommandPoolCreateFlags.gen.cs
```diff
@@ -0,0 +1,95 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// Ported from the Vulkan headers and corresponding dependencies.
+// Original source is Copyright 2015-2023 The Khronos Group Inc. Licensed under the MIT license.
+using System.Runtime.CompilerServices;
+
```
Perksey, line 6: Why do these files have a Vk suffix in their filename?

Perksey (commented):
Looks pretty solid.

sources/SilkTouch/SilkTouch/Clang/UnixStdIncludeResolver.cs
Exanite: I'm leaning towards no. Other than needing to parse the string, my local include folder is from v21 while the string specifies v20. I also only have one include folder on Arch Linux. If we do use `getClangVersion`, it'll have to be a hint.

image

image

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
Exanite: I agree. A set of flags types would presumably work better. I should check how the other Khronos APIs work though.

sources/Vulkan/Vulkan/Handles/BufferTHandle.gen.cs
Exanite: Ah, I knew I put the wrong term somewhere. I meant the new prettification approach, which is not implemented. I plan to do this in my next PR.

sources/Vulkan/Vulkan/Vulkan/CommandPoolCreateFlags.gen.cs
Exanite: I think it's because the renaming is partially done by MixKhronosData so prettification does not rename the file later, but I have to check. I'll fix this before the PR is merged.

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
Exanite: This was removed in this commit:
image

sources/OpenGL/OpenGL/Enums/FogParameter.gen.cs
Exanite: This (very old) commit shows the differences between the two platforms: https://github.com/dotnet/Silk.NET/pull/2457/commits/20ef4637fdf5910ede171aa70676bfe2f4cd059b

The main differences are that:
- Enums use uint on Linux. Nothing (so int) on Windows. I believe ClangSharp's own unit tests check for this difference so I'm guessing this is as expected. We could just replace uint enums with int if we don't care about the slight behavioral differences though.
- Certain usages of ulong are replaced with nuint in Linux. (I haven't looked into why this is the case)

Let me know if you have any ideas on how to resolve these. Otherwise, let me know if I should investigate this further.

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
Exanite: It was rewinding the next block of code.
I don't remember what the incorrect result looked like, but it probably is that the incorrect results had extra vendor suffixes.

Edit: Reordering makes no difference anymore. I'm currently looking into this further.
Edit^2: Looks like my commit message was pretty literal. This doesn't affect any current bindings.

Commit in question:
image

"Next block of code", for reference:
image

eng/silktouch/vulkan/remap.rsp
Exanite: This doesn't seem to work for structs. Is TransformFunctions supposed to transform structs as well?
image

This also changes the generated methods in Vk.gen.cs to have a `uint` and `MaybeBool<uint>` overload, which I assume is the desired result.

Perksey: Yeah that's intentional. As for the structs, we haven't done any work looking into field/property transformations which is why it's like this today - this is a future work item across all bindings.

sources/OpenGL/OpenGL/Enums/FogParameter.gen.cs
Perksey: For the former, yes we should manually type map to one consistent type.

For the latter, these often indicate cases that should use CLong/CULong, likely requiring a type map to do so.

Perksey: Although I wasn't aware of any cases in Vulkan where CLong/CULong is needed, as they tend to be pretty bulletproof here.

Perksey: Okay yeah looking at those commits we should typemap these ourselves as they have consistent sizes in the spec, but because the header maps them into whatever type happens to be the correct size it has differing binding behaviour across platforms.

eng/silktouch/vulkan/remap.rsp
Exanite: Got it, that makes sense.
I can commit the uint version if that's desired.

Perksey: Would be appreciated 🙏 thanks

We can talk more in the discord about how to approach structs after this PR

Exanite: Alright, done and added to my future PR task tracker above.
I ended up removing the remapping entirely since VkBool32 is already typedef'd:
typedef uint32_t VkBool32;

Going to mark as resolved.

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
Exanite: Alright, done:
image

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
Perksey: Ah ok, thanks.

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
Perksey: Thanks.

sources/SilkTouch/SilkTouch/Clang/UnixStdIncludeResolver.cs
Perksey: Fair enough.

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
Exanite: Done! Here's the new syntax:
image

I also updated the rest of the codebase to use it (ChangeNamespace and IInputResolver).
It makes it much clearer what actually is being done.

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
Exanite: I decided to replace `ModUtils.GetNativeElementTypeName` with `TryParseNativeTypeName`, which handles both the original functionality of parsing inputs that look like `const NAME **` and the new functionality of parsing inputs that look like `#define NAME VALUE`.

The method's implementation is not exhaustive and will silently drop any unsupported formats, such as `const uint32_t[8]`. I considered logging or throwing an exception, but logging is noisy and exceptions don't work well for intentionally unimplemented cases that are commonly hit.

My intention with this method is that we implement cases as we need it and consolidate any NativeTypeName parsing to here.

sources/Vulkan/Vulkan/Vulkan/CommandPoolCreateFlags.gen.cs
Exanite: Fixed! I rename the document file paths and names in MixKhronosData.

One thing I'm noticing though is that we just search and replace the entire file path in PrettifyNames. 
The file path is absolute as well.

It'll probably be bad if someone attempts to run the SilkTouch generator in a folder structure with folder names happening to **contain** (not just match) a replaced name. Since these names can be any name that exists in a C header, these collisions are fairly possible.

We should probably add an extension method that handles converting the path to be relative to the project path before searching and replacing to reduce the chance of collisions. Alternatively, we can substring the path based on the project path.

Perksey: Good point, agree with adding an extension method in a future PR.

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
Perksey: Sounds good.

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
Perksey: Awesome, thanks!

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
Perksey: Thanks!

eng/silktouch/vulkan/header.txt
Perksey: Probably wouldn't hurt to update the date I guess if you've got more work to do anyway, otherwise I'm not too bothered.

Exanite: I changed it to 2015-2025. Should be good now. I was basing it off the years in the LICENSE.md before, but it looks like the Vulkan headers and xml say 2015-2025.

sources/OpenGL/OpenGL/Enums/FogParameter.gen.cs
Exanite: Ah looks like the ulong/nuint case is already fixed with Vulkan. I was just seeing it in the other bindings since they don't use the same remappings.
image

Could be worth making these remappings a common practice in the Silk repo for cross platform consistency (we'd just manually import the shared remappings in each remap.rsp file). I'm not sure about having these as a global default though.

Exanite: I'll add a CoerceBackingType option to TransformEnums to handle the other case. I'll have it be an enum and only implemented the PreferredSigned case for now though.

Exanite: Alright, done, but note that I only changed this behavior for the Vulkan bindings. We probably want to do so for all Silk repo bindings at some point.

Only difference between the Linux and Windows outputs now is the NativeTypeName attribute, which unfortunately still leads to noisy diffs, but at least the generated functionality is equivalent now.

I can strip NativeTypeName for the coerced enums, but I'm tempted to just strip the NativeTypeName from all bindings in general (eg: StripNativeTypeName as a final mod in the mod order; separate PR though). We're adding a NativeName attribute for the prettifier anyway and that provides most of the information (although in a different place since native type names and member names aren't the same thing).

sources/Vulkan/Vulkan/Vulkan/CommandPoolCreateFlags.gen.cs
Exanite: Extension method added as `ModUtils.ReplaceNameAndPath` :D

eng/build/Build.cs
```diff
@@ -1,6 +1,7 @@
 using System.IO;
 using Nuke.Common;
 using Nuke.Common.IO;
+using Nuke.Common.ProjectModel;
```
Exanite, line 4: This was needed to fix a compile error after updating packages.

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
```diff
@@ -1221,6 +1233,63 @@ public void Trim(NameTrimmerContext context)
...
+                    prevList.Add(prev);
+                }
+
+                context.Names[original] = (prev[(prev.IndexOf('_') + 1)..], prevList);
+            }
+        }
+
+        // Trim _T from the end of names
+        // This is targeted towards Vulkan handle type names, which end in _T
+        if (context.Container is null)
```
Exanite, line 1278: I moved this up since _T is always at the end of names, so it should be processed first.

Exanite (commented):
Self review done. Marking PR as ready for review :D

sources/Vulkan/Vulkan/Enums/AccessFlags2.gen.cs
```diff
@@ -10,7 +10,7 @@ namespace Silk.NET.Vulkan;
 
 [Transformed]
 [Flags]
-public enum AccessFlags2 : ulong
+public enum AccessFlags2 : long
```
Perksey: As alluded to in Discord, 2.X is wrong here - `ulong` is correct. We may need a `PreferUnsigned`

sources/Vulkan/Vulkan/Handles/AccelerationStructureKHRHandle.gen.cs
```diff
@@ -8,34 +8,34 @@
...
-        IndirectExecutionSetEXTTHandle left,
-        IndirectExecutionSetEXTTHandle right
+        AccelerationStructureKHRHandle left,
+        AccelerationStructureKHRHandle right
     ) => left.Equals(right);
 
     public static bool operator !=(
-        IndirectExecutionSetEXTTHandle left,
-        IndirectExecutionSetEXTTHandle right
+        AccelerationStructureKHRHandle left,
```
Perksey, line 28: Future discussion, not to throw a spanner in the works, but should we consider `AccelerationStructureHandleKHR`? I have absolutely no idea what abstraction we need to make it happen lol, but was just wondering from a readability/consistency perspective.

Perksey (commented):
Almost there! Just one proper comment really

sources/Vulkan/Vulkan/Enums/AccessFlags2.gen.cs
Exanite: Yeah, `PreferUnsigned` would handle the enums better:
image

Looks like both the 32-bit and 64-bit flags enums use unsigned ints.

I'll make the change.

sources/Vulkan/Vulkan/Handles/AccelerationStructureKHRHandle.gen.cs
Exanite: Hmm, `AccelerationStructureHandleKHR` is easier to read, but it further deviates from the native name.

Making this change requires having 3 separate mods cooperate (MixKhronosData, TransformHandles, and PrettifyNames), and it requires their name changes to be applied in an order that's potentially different from the defined mod order.

`AccelerationStructureHandleKHR` is more "silky" though, so I won't throw it out just yet.

sources/Vulkan/Vulkan/Enums/AccessFlags2.gen.cs
Exanite: PreferUnsigned is now implemented!

sources/Vulkan/Vulkan/Handles/AccelerationStructureKHRHandle.gen.cs
Exanite: I added this to my future PRs list as a possible consideration.

Perksey: Awesome, thanks! Yeah I know this isn't a simple ask and the implementation will be hell, we can discuss more another time.

Perksey (approved):

PR #2466, opened 2025-06-28
Add bindings for MediaFoundation

Ethereal77:
# Summary of the PR

Adds support for [MediaFoundation](https://learn.microsoft.com/es-es/windows/win32/medfound/microsoft-media-foundation-sdk) from the Windows SDK, which according to MSDN:

> ...enables the development of applications and components for using digital media on Windows Vista and later.

As discussed in Discord, this is part of the ongoing effort to port the [Stride Game Engine](https://github.com/Stride3d/stride) away from SharpDX and towards Silk.NET.

### DIFF
--- a/build/include/silk.net.mediafoundation.h
+++ b/build/include/silk.net.mediafoundation.h
@@ -1,0 +1,2 @@
+#include <Windows.h>
+#include <mfmediaengine.h>
--- a/generator.json
+++ b/generator.json
@@ -2778,7 +2778,72 @@
             "namespace": "Silk.NET.XInput",
             "extensionsNamespace": "Silk.NET.XInput.Extensions",
             "typeMaps": [
-                {},
+                {
+                    "$include.dxTypemap": "build/dx_typemap.json",
+                    "$include.commonTypeMap": "build/csharp_typemap.json"
+                }
+            ]
+        },
+        {
+            "profileName": "MediaFoundation",
+            "sources": [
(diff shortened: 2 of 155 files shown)


Perksey:
> It can't found the namespace Maths for some of the files. I suspect in those that need Box2D<int>, that appears to be one of the types it can't resolve.

Interesting, might need a ProjectReference.

> Wrt to the other types, I need to verify if I can import only some of them without bringing all the types defined in their respective headers. 

To be honest, all of those seem like MediaFoundation APIs, so provided that those aren't in a non-MediaFoundation header (or otherwise pull in non-MediaFoundation APIs) I'd say just bring them all in.

Perksey:
Btw feel free to ping in our Discord, GitHub tends to have a lot of noise (didn't realise this had outstanding questions!)

Perksey:
Is this still being worked on?

Ethereal77:
I came to a halt while working on this. I wanted to reproduce what Tanner has done in TerraFX for MediaFoundation, but it seems more difficult than I anticipated.

I talked about the problems I encountered in the Silk's Discord. Basically, MediaFoundation has a lot of types, and those types come from many headers, and are grouped so I'd need to specify a different "--with-library" for some of those groups.

I didn't know how to write the `generator.json` in an elegant way without making an explosion of entries. Also, when I asked, no one seemed to know enough about the generation process to help me with these problems 😢.

PR #2503, opened 2025-11-24
[3.0] Implement name-related improvements identified in previous Vulkan PR

Exanite:
# Summary of the PR

This PR focuses on implementing the naming-related improvements identified in #2457.
Depending on the scope of the tasks, some of these tasks may be pushed to another PR.

This PR also contains the changes of multiple other PRs.
See each individual PR for more information.

Included PRs:
- https://github.com/Exanite/Silk.NET/pull/19
- https://github.com/Exanite/Silk.NET/pull/21
- https://github.com/Exanite/Silk.NET/pull/23
- https://github.com/Exanite/Silk.NET/pull/24
- https://github.com/Exanite/Silk.NET/pull/25
- https://github.com/Exanite/Silk.NET/pull/28
- https://github.com/Exanite/Silk.NET/pull/29

Important changes:
- Handle types are now named as `-HandleEXT` instead of `-EXTHandle`.
   - Eg: `AccelerationStructureHandleKHR`
- Function pointer delegate types are now named as `-DelegateEXT` instead of `EXTDelegate`.
   - Eg: `pfn struct BufferCallbackSOFT` and `pfn delegate BufferCallbackDelegateSOFT`
- Functions with data type suffixes following a number such as `alGetSourcei64vDirectSOFT` are now trimmed and prettified as `GetSourcei64vDirectSOFT` instead of `GetSourcei64VDirectSOFT`. Note the capitalization of `v`. This is because affixes are added after prettification.
- Vendor suffixes are now kept strictly uppercased.
- OpenGL vendor suffix trimming is now generalized and can be configured and used for any Khronos API.
   - Notably, `TrimEnumMemberImpliedVendors` trims enum member vendor suffixes if it matches the enum type vendor suffix.
- Struct properties are now prettified. They were previously ignored.
- NativeName is a new attribute that saves the name output by ClangSharp. This for the most part isn't used by the generator, but can be useful for the end user.
- NameAffix is a new attribute that allows prefixes/suffixes to be declared upfront and reordered. The idea is that mods add the attribute to identifiers and PrettifyNames applies the affixes during trimming and prettification.
- A large portion of the trimming/PrettifyNames code has been rewritten.
   - NameTrimmer has been kept mostly the same.
   - `INameTrimmer.Order` is used to order the trimmers instead of the version.
   - PrettifyNames.Visitor's functionality is mostly the same, but also gathers name affix data.
   - MixKhronosData.Trim is almost all replaced by the name affix system. This is implemented in MixKhronosData.RewriterPhase3 (yes, there are 3 rewriters, but this keeps things simpler without adding much time cost).
   - Most usages of tuples have been replaced with record structs.
   - Most nullable collections (eg: `List<string>?`) have been made non-nullable. Most of these get allocated anyway and it makes the code more maintainable. If there is any performance issue (mainly concerning the Microsoft job), I will take the time to optimize it.
   - Prettification (as in calling the Prettify extension method) is now implemented as another trimmer. Prettification now happens before name conflict resolution and before name affixes are added.
- Handle structs now have constructors since they were previously not possible to construct (field is readonly).
- The StripAttributes mod can be used to remove attributes from the final bindings.
   - Currently the following are removed: `NativeTypeName`, `NameAffix`, `Transformed`

Previous Vulkan PR (initial bindings generation): #2457
Discord thread: https://discord.com/channels/521092042781229087/1376331581198827520/1442651368207941643

This PR also contains the changes of multiple other PRs.
See each individual PR for more information.

Included PRs:
- https://github.com/Exanite/Silk.NET/pull/19
- https://github.com/Exanite/Silk.NET/pull/21
- https://github.com/Exanite/Silk.NET/pull/23
- https://github.com/Exanite/Silk.NET/pull/24
- https://github.com/Exanite/Silk.NET/pull/25
- https://github.com/Exanite/Silk.NET/pull/28
- https://github.com/Exanite/Silk.NET/pull/29

# Further Comments

## Tasks not part of this PR

These are the tasks from my previous PR. These have been sorted and trimmed down to ensure that this PR remains focused. These may be added to this PR.

If you want to see the original, unmodified set of tasks, please see #2457.

Please see this link for an up to date list: https://github.com/orgs/dotnet/projects/64/views/16?pane=issue&itemId=168550672

These will also be part of a future PR.

- [ ] Apply BoolType transformation to VkBool32 in structs. They currently are only handled for functions.
- [ ] Add default field values to structs where it makes sense (eg: `SType`)
- [ ] Ensure SupportedApiProfiles attributes are correct
   - [ ] Properly resolve API profiles for Flags/FlagBits types
   - [ ] Prefer name from `NativeName` for resolving API profiles

Future PRs

- [ ] Update Curin's branch to use my new LocationTransformation code. The new renamer should be replaced with this. See the NameUtils.cs and Renamer.cs files on Curin's branch. Renamer.cs should be removed. NameUtils.cs should have the methods updated to use my LocationTransformation code instead (this should show up as a merge conflict).
   - [x] PR opened and ready for review/merge
   - [x] PR merged
   - [ ] Old renamer code removed
- [ ] Consider opening a ClangSharp PR to fix issue with `SDL_MAX_SINT64` generation on Linux
   - More context: https://discord.com/channels/521092042781229087/1376331581198827520/1443401810822828113
   - `public const nint SDL_MAX_SINT64 = unchecked(0x7FFFFFFFFFFFFFFF);` does not compile
- [ ] Reimplement the struct chaining API
- [ ] Fix issues when running multiple jobs
   - [ ] Figure out why the headers are getting mixed up when running multiple jobs at the same time.
       - Seems to be related to `TransformFunctions`.
   - [ ] Figure out why the SDL handle structs are getting named as `Silk.NET.SDL.ConditionHandle.gen.cs` instead of `ConditionHandle.gen.cs` when running multiple jobs at the same time.
- [ ] Maybe rework how using statements are added. Not sure why, but using statements are randomly added/removed between Windows/Linux. This causes noisy diffs.
- [ ] Consider splitting `ExtractNestedTyping` and `TransformHandles` into a new set of `Extract-` mods.
    - This this mainly for maintainability for `ExtractNestedTyping`, but for `TransformHandles` it is useful for `AddApiProfiles` to be executed after all types are extracted, but not yet transformed.
    - Also consider making `AddApiProfiles` work off of `NativeName` attributes.
- [ ] Optimize/refactor name overrides in PrettifyNames
- [ ] Consider reworking data type trimming regexes.
   - This was discussed more here: https://github.com/dotnet/Silk.NET/pull/2503#issuecomment-3622907776
   - The summary is that the current set of 2 regexes doesn't elegantly handle the case where the ending of a word should not be trimmed, but the word itself can have data type suffixes.
   - For example: `alSourceRewindv` and `alSourceRewind`. The `d` in `Rewind` can be mistaken as a data type suffix, so we add it to the `EndingsNotToTrim` regex. However, the `dv` in `Rewindv` also has the same problem.
   - We likely need something more akin to `WordsNotToTrimInto` than `EndingsNotToTrim`.
- [ ] Optimize `NativeName` attributes.
   -  This should involve renaming `StripAttributes` to `CleanupAttributes` and adding an option in its config to optimize `NativeName` attributes.
   - Perksey seems to really want this so considering prioritizing this: https://github.com/dotnet/Silk.NET/pull/2503#discussion_r2662921224
- [ ] Consider replacing `LenientUnderscore` and updating codebase to use `NameSplitter`
- [ ] Consider implementing a `IsProbablyABitmask` function based on [BuildTool's IsProbablyABitmask](https://github.com/dotnet/Silk.NET/blob/c87fc664cc5d15abd754997b0a9ad138470554c4/src/Core/Silk.NET.BuildTools/Common/Utilities.cs#L627) for identifying flags enums
   - Original comment: https://github.com/dotnet/Silk.NET/pull/2503#discussion_r2662888928
- [ ] Improve handling of nested struct names
	- Eg: `PerformanceCounterDescriptionArmName` should be `PerformanceCounterDescriptionARMName`
	- Conversation here: https://discord.com/channels/521092042781229087/587346162802229298/1460017726813900975
- [ ] Address `INameTrimmer.Order/Version` issue (we don't want magic numbers) and consider removing `INameTrimmer` entirely.

Done during later PRs:

- [x] Prevent `OES` being singularizing into `O` in `ArrayParameterTransformer.TransformArrayParameterRewriter`.
   - Eg: `glFeedbackBufferxOES` 
   - Discussed solution is to expose the name affix API so that `TransformArrayParameterRewriter` can singularize the base name instead of the full name.
- [x] Change native name of `GLEnum` to `GLenum`. Same for `ALCenum` and `ALenum`.

## Completed Todos

Newest groups at the top. Order within a group is chronological.

Note that I also wrote a summary at the top. The summary is more comprehensive, but this section can be useful getting additional context. Also see my commit descriptions if more context is needed.

Prepare to merge.

- [x] Finish self code review
- [x] Generate bindings on Windows for consistency
- [x] Temporarily remove generated files from PR during code review
   - This needs to be done since there are too many generated files and they hide some of the manually changed files. 
   - [x] Revert the commit after code review: https://github.com/Exanite/Silk.NET/pull/19
- [x] Review the generated bindings as a second code review
- [x] Consider updating OpenAL submodule and regenerating
- [x] Merge name affix config PR into this PR
- [x] Fix broken test cases (I know there is at least one)

Acronym threshold change and prettify improvements

- [x] Acronym related name improvements discussed here: https://discord.com/channels/521092042781229087/587346162802229298/1446979314330501172
   - [x] Re-read discussion and organize into tasks.
   - [x] Change default acronym threshold to 2 for all bindings
   - [x] Change default acronym threshold to 2 to be the default if not specified in the PrettifyNames config.
   - [x] Handle `Mat4X4` case. The "X" should be lowercased.
   - [x] Update existing names to match framework design guidelines
      - [x] `EGL` as `Egl`
      - [x] Name overrides in generator config

INameTrimmer.Order

- [x] Add priority system to `INameTrimmer`.
   - I implemented the new name affixer system using trimmers, so this introduced 3 new trimmers in order to separate each stage properly. There is now a new `INameTrimmer.Order` property and internal `TrimmerOrder` enum.

Fix misc issues I noticed

- [x] `VkVendorId` members are not trimmed properly
- [x] Figure out why the members of the enum `AttribMask` in OpenGL are not getting rewritten.
   - Initial investigation shows that the `fieldSymbol.ConstantValue` is null, but why it is null is unclear. Maybe a Roslyn bug? Example member: `DepthBufferBit = unchecked((uint)0x00000100)`. Note that similar expressions in Vulkan are fine, eg: `unchecked((ulong)0x00000002UL)`. Maybe it is because the literal type and cast type are different.
   - Wait, are these the enums that we generate?
   - Yup, they are. MixKhronosData was outputting subtly invalid type names. I changed it to use `ParseTypeName(baseType)` instead of `IdentifierName(baseType)`.

Fix issues introduced by name affix system

- [x] Identify and fix issues introduced by name affixation. Also just identify and fix naming issues in general.
   - [x] Data type trimming in OpenAL seems to be broken. Probably OpenGL as well. 
   - [x] Check if I need to reimplement vendor suffix removal for OpenGL. I probably do.
      - [x] Discuss: I decided to change the vendor suffix removal for enum members so that it removes the member suffix if it matches the type suffix. This prevents cases where a member of a different suffix is added, causing all members to have their suffixes restored.
      - [x] Discuss: The above change made it so that the member suffix removal code is usable for Vulkan so I enabled it for Vulkan as well. See PresentModeKHR to see how it looks (the KHR suffixed members have their suffixes removed, but EXT remains the same).
   - [x] Reapply some of my `MixKhronosData` code removals to the current `develop/3.0` branch and see what changes would be made. This is to ensure I don't accidentally remove functionality when deleting code.
- [x] Ensure other APIs are configured consistently with Vulkan. Eg: Ensure affix priorities are set, etc.

Name affix system

- [x] Properly trim prefixed/suffixed names, such as `PFNVkDebugUtilsMessengerCallbackEXT` and `BufferTHandle`
   - [x] `PFNVkDebugUtilsMessengerCallbackEXT` prefix
   - [x] `PFNVkDebugUtilsMessengerCallbackDelegateEXT` suffix
   - [x] `AccelerationStructureHandleKHR` suffix
   - [x] Extract vendor extensions from names and store into `NameSuffix` attributes.
   - [x] Revert "hacky" THandle fix.
- [x] Consider renaming handle types as `HandleEXT` instead of `EXTHandle` for readability.
   - This is fairly doable. The above todos track this. 

Name metadata

- [x] Add NativeName attribute
- [x] Ensure NativeName attribute is added by all mods that generate new types/members
   - Technically not strictly required, but nice from a API consistency standpoint. 
   - [x] MixKhronosData
      - [x] Enum members
      - [x] Enum declarations
         - [x] Vulkan (uses FlagBits version of enum names since it is more helpful when googling)
         - [x] Discuss validity in other Khronos APIs (some APIs don't have actual enums)
            - Eg: `GLEnum` and the other OpenGL enums
   - [x] ExtractNestedTyping
      - [x] Function pointers
      - [x] Others?
   - [x] TransformHandles 

Name prettification

- [x] ~~Change naming convention to not capitalize acronyms (i.e., strictly pascal cased)~~
   - ~~See: https://discord.com/channels/521092042781229087/1376331581198827520/1395097379329282089~~ 
   - [x] ~~Update generated method class names in the rsp files to be strictly pascal case~~
- [x] Fix `StdVideoEncodeH264SliceHeaderFlags` property names. They currently aren't getting prettified.

Fix bugs from original PR

- [x] Enums containing negative values should not have an unsigned backing type.

Handle struct improvements

- [x] Handle structs don't have an easy way of being constructed (field is readonly and no constructor).


eng/silktouch/openal/alsoft/generate.rsp
```diff
@@ -5,7 +5,7 @@ _GUID
 --file
 openal-soft-al.h
 --methodClassName
-AL
+Al
```
Exanite: I replaced _most_ usages of `AL` and `GL` with `Al` and `Gl` (I still need to double check comments since my renames don't catch that).
However, I did not change namespaces, so they still use OpenAL/GL instead of OpenAl/Gl. 

This is more consistent with the rest of the bindings, but I'm not sure if we want to make an exception here and use `GL` instead. Notably `SDL` is already lowercased as `Sdl` so I didn't change that.

sources/OpenAL/OpenAL/al/EfxEaxReverbPropertiesFlLateReverbPan.gen.cs
```diff
@@ -10,8 +10,10 @@
 
 namespace Silk.NET.OpenAL;
 
+[NativeName("_flLateReverbPan_e__FixedBuffer")]
```
Exanite, line 13: Not sure how I feel about this native name. Probably fine?

sources/OpenAL/OpenAL/al/EfxEaxReverbPropertiesFlLateReverbPan.gen.cs
```diff
@@ -10,8 +10,10 @@
...
 [InlineArray(3)]
 public partial struct EfxEaxReverbPropertiesFlLateReverbPan
 {
+    [NativeName("e0")]
```
Exanite, line 17: Also not sure how I feel about this native name. Probably fine?

sources/OpenAL/OpenAL/al/EfxReverbPreset.gen.cs
```diff
@@ -12,8 +12,8 @@ namespace Silk.NET.OpenAL;
 
 public unsafe partial class EfxReverbPreset
 {
-    [NativeTypeName("const EFXEAXREVERBPROPERTIES")]
-    public static ref readonly EfxEaxReverbProperties Generic
+    [NativeName("EFX_REVERB_PRESET_GENERIC_")]
+    public static ref readonly EfxEaxReverbProperties EfxReverbPresetGeneric
```
Exanite: Need to investigate this.
Edit: Fixed.

sources/SilkTouch/SilkTouch/Naming/NameUtils.cs
```diff
@@ -37,42 +37,43 @@ public static partial class NameUtils
         "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789_"
     );
 
+    private static readonly IStringTransformer[] _prettifyTransformers = [new NameTransformer()];
```
Exanite: This avoids an allocation due to the `params` keyword every time Prettify is called.
Edit: Outdated since this change was reverted and incompatible with the old setup.

sources/Core/Core/Annotations/NameAffixAttribute.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// The priority with which the affix is used
+    /// to create secondary names in case of conflicts.
+    /// <para/>
+    /// Negative means the affix is required, but won't be used to create secondaries.
+    /// Non-negative means the affix is optional, but will be used to create secondaries.
+    /// Higher means the names created using the affix is tried first.
+    /// <para/>
+    /// Affixes with the same priority are applied together as a group.
+    /// </summary>
+    public int Priority { get; }
```
Exanite: Consider renaming Order and Priority since this is confusing.
Edit: Resolved. `Priority` is now renamed to `DiscriminatorPriority`.

sources/SilkTouch/SilkTouch/Mods/Common/ModUtils.cs
```diff
@@ -160,44 +153,6 @@ public static string DiscrimStr(
     public static string DiscrimStr(BaseParameterSyntax param) =>
         DiscrimStr(param.Modifiers, param.Type);
 
-    /// <summary>
-    /// Modifies the <see cref="DllImportAttribute"/>s the method may have to make them resistant to method identifier
-    /// changes.
-    /// </summary>
-    /// <param name="node">The original method.</param>
-    /// <returns>The modified method.</returns>
-    public static MethodDeclarationSyntax WithRenameSafeAttributeLists(
```
Exanite, line 169: All Attribute related code in ModUtils has been moved to AttributeUtils

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
```diff
@@ -1208,12 +1191,10 @@ var ext in xml.Element("registry")?.Element("extensions")?.Elements("extension")
         }
     }
 
-    private bool _outputVendorInformationWarning = false;
-
     /// <inheritdoc />
     public void Trim(NameTrimmerContext context)
     {
-        if (context.Names is null || context.JobKey is null)
```
Exanite, line 1216: Most of Trim was replaced by RewriterPhase3 below

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
```diff
@@ -1800,7 +1505,7 @@ public bool TryGetSymbolMetadata(
     /// while the lookbehind asserts that the ending match will not overreach into the end of a word.
     /// </summary>
     // NOTE: LET THIS BE A LESSON! Do NOT add x (fixed) here, these will frequently conflict with integer overloads.
-    [GeneratedRegex("(?<!xe)([fdh]v?|u?[isb](64)?v?|v|i_v|fi|hi)$")]
+    [GeneratedRegex("(?<!xe)((?<!Rewin)[dhf]v?|u?[isb](64)?v?|v|i_v|fi|hi)$")]
```
Exanite: Rewindv in OpenAL was getting trimmed as Rewin

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
```diff
@@ -38,19 +34,8 @@ public partial class MixKhronosData(
         IApiMetadataProvider<IEnumerable<SupportedApiProfileAttribute>>
 {
     internal ConcurrentDictionary<string, JobData> Jobs = new();
-    private static readonly ICulturedStringTransformer _transformer = new NameUtils.NameTransformer(
-        4
-    );
     private static readonly char[] _listSeparators = { ',', '|', '+' };
 
-    private static readonly Dictionary<string, string> _defaultEnumNativeTypeNameMaps =
```
Exanite, line 46: This dictionary was effectively unused.

Exanite (commented):
Initial self code review. PR is not yet ready for review or merge.

Exanite (commented):
Self code review complete. Will mark as ready for review shortly.

Note that I only have the non-generated changes committed.

The rest is in a separate commit that I will apply after the initial review.
This is because the generated changes hide the non-generated changes (it goes over the 1k file limit).

DO NOT merge the PR until the generated changes are committed, and preferably, reviewed.

Exanite:
Ready for review!

If you want to see the generated code see: https://github.com/dotnet/Silk.NET/pull/2503/commits/9c9fbfc037b662a260ff5c8eccbdeda4c064cc78.

Note that the red/removed code is the new code since this code is meant to temporarily revert the generated changes.

sources/SilkTouch/SilkTouch/Mods/PrettifyNames.cs
```diff
@@ -686,276 +725,709 @@ out var tnff
...
+                );
+            }
+            if (!responseFile.GeneratorConfiguration.DontUseUsingStaticsForGuidMember)
+            {
+                logger.LogWarning(
+                    "{} (for {}) should use exclude-using-statics-for-guid-members as PrettifyNames does not resolve "
+                    + "conflicts with members of other types.",
+                    responseFile.FilePath,
+                    key
+                );
```
Exanite, line 854: This is from the Microsoft branch

sources/SilkTouch/SilkTouch/Mods/PrettifyNames.cs
```diff
@@ -336,26 +323,100 @@ z.MethodKind is MethodKind.Constructor or MethodKind.Destructor
...
-            RenameInComments: false,
-            RenameFile: false
-        );
+    /// <summary>
+    /// Applies the prettify only pipeline.
+    /// This currently consists of checking for name overrides first.
+    /// Then if no override is found, then the name's affixes are removed,
+    /// the name is prettified, and the name's affixes are added back.
+    /// </summary>
+    private string ApplyPrettifyOnlyPipeline(
```
Exanite, line 462: This is from the Microsoft branch, but heavily modified.

sources/SilkTouch/SilkTouch/Mods/PrettifyNames.cs
```diff
@@ -336,26 +323,100 @@ z.MethodKind is MethodKind.Constructor or MethodKind.Destructor
...
+        string? container,
+        string name,
+        Dictionary<string, string> nameOverrides,
+        Dictionary<string, TypeAffixData> affixTypes,
+        ICulturedStringTransformer nameTransformer)
+    {
+        // Check for overrides
+        foreach (var (nativeName, overriddenName) in nameOverrides)
```
Exanite, line 471: Ideally this code isn't duplicated with the code in Trim(), but to do this performantly requires a decent amount of refactoring.

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
```diff
@@ -1887,53 +1586,42 @@ private class RewriterPhase1(JobData job, ILogger logger) : CSharpSyntaxRewriter
...
+                    var attributes = SingletonList(
+                        AttributeList(
+                            [Attribute(IdentifierName("Transformed"))]));
+
+                    if (groupInfo.NativeName != null)
+                    {
+                        attributes = attributes.WithNativeName(groupInfo.NativeName);
+                    }
+
+                    var baseTypeSyntax = ParseTypeName(baseType);
```
Exanite, line 1597: ParseTypeName fixes an issue where IFieldSymbol.ConstantValue was returning null.

Exanite (commented):
Self code review complete. Going to mark as ready for review.

Note that I only have the non-generated changes committed.
The generated changes are here: https://github.com/Exanite/Silk.NET/pull/19

This is because the generated changes hide the non-generated changes (it goes over the 1k file limit).

Feel free to comment on the linked PR as well.
It will be merged into this PR after the non-generated changes are reviewed.

DO NOT merge this PR until the generated changes are merged, and preferably, reviewed.

sources/SilkTouch/SilkTouch/Mods/MarkNativeNames.cs
```diff
@@ -0,0 +1,113 @@
...
+    private class Rewriter : ModCSharpSyntaxRewriter
+    {
+        private SyntaxList<AttributeListSyntax> TryAddNativeNameAttribute(SyntaxList<AttributeListSyntax> attributeLists, SyntaxToken identifier)
+        {
+            if (attributeLists.TryGetNativeName(out _))
+            {
+                return attributeLists;
+            }
+
+            return attributeLists.WithNativeName(identifier.Text);
```
Perksey, line 60: I think we should change this to not apply NativeName if:
- the native name is identical to the current name, and
- it is obvious that the symbol is representing a native symbol (i.e. one of the parent syntax nodes has a `NativeName`)

This way `NativeName` is still used to signify that this a representation of a native symbol, but where that's already known, we're not adding attributes that aren't conveying any extra information (such as when the names are identical)

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
Perksey: Why didn't the EndingsNotToTrim regex immediately below this not pick it up?

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
```diff
@@ -2140,6 +1828,190 @@ private bool TryGetManagedEnumType(SyntaxList<AttributeListSyntax> attributes, [
...
+            var variable = node.Declaration.Variables.First();
+            return node.WithAttributeLists(ProcessAndGetNewAttributes(node.AttributeLists, variable.Identifier, false));
+        }
+
+        public override SyntaxNode VisitDelegateDeclaration(DelegateDeclarationSyntax node) =>
+            node.WithAttributeLists(ProcessAndGetNewAttributes(node.AttributeLists, node.Identifier, false));
+
+        public override SyntaxNode VisitEnumDeclaration(EnumDeclarationSyntax node)
+        {
+            // Special case for enums since this code needs information about the enum type and its members at the same time.
```
Perksey: Why does it need to know its type?

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
```diff
@@ -2140,6 +1828,190 @@ private bool TryGetManagedEnumType(SyntaxList<AttributeListSyntax> attributes, [
...
+            // Figure out the enum's exclusive vendor
+            var exclusiveVendor = groupInfo?.ExclusiveVendor ?? typeVendor;
+            if (exclusiveVendor == null || !node.Members.All(member => member.Identifier.Text.EndsWith(exclusiveVendor)))
+            {
+                // Not all enum members share the exclusive vendor
+                // This means the vendor suffix isn't actually exclusive
+                exclusiveVendor = null;
+            }
+
+            // Check if the enum contains unsuffixed members
```
Perksey: Why?

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
```diff
@@ -2140,6 +1828,190 @@ private bool TryGetManagedEnumType(SyntaxList<AttributeListSyntax> attributes, [
...
+                var memberName = member.AttributeLists.GetNativeNameOrDefault(member.Identifier);
+                if (job.Vendors.FirstOrDefault(memberName.EndsWith) == null)
+                {
+                    return true;
+                }
+
+                return false;
+            });
+
+            var isSafeToTrimMembers = !containsUnsuffixed;
```
Perksey: Why does it not containing any unsuffixed vendors make it safe? How is this different to exclusiveVendor?

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
```diff
@@ -2140,6 +1828,190 @@ private bool TryGetManagedEnumType(SyntaxList<AttributeListSyntax> attributes, [
...
+
+            if (config.TrimEnumTypeNonExclusiveVendors && typeVendor != null)
+            {
+                var shouldTrimType = typeVendor != exclusiveVendor;
```
Perksey: I won't keep asking questions out loud, but generally I think this code could do with some more comments with some examples of what it's actually doing as it's quite hard to visualise. 

I think what it's doing makes sense but it requires a lot of thought and careful reading to figure out what it's doing.

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
```diff
@@ -2140,6 +1828,190 @@ private bool TryGetManagedEnumType(SyntaxList<AttributeListSyntax> attributes, [
...
+
+                // Check if there are other versions of the enum (this includes the core variant and other vendor variants)
+                var isSafeToTrimType = job.Groups.Count(x => x.Key.StartsWith(typeName[..^typeVendor.Length])) <= 1;
+
+                if (shouldTrimType)
+                {
+                    if (isSafeToTrimType)
+                    {
+                        // Remove the exclusive vendor from the enum name since it is wrong and it is safe to do so
```
Perksey: Why is it safe to do so? Why is it wrong? Can you give an example of where this code is used? This could result in breaking changes!

sources/SilkTouch/SilkTouch/Naming/INameTrimmer.cs
```diff
@@ -15,6 +15,12 @@ public interface INameTrimmer
     /// </summary>
     Version Version { get; }
 
+    /// <summary>
+    /// Used to define the order that the trimmers will run in.
+    /// Higher values indicate that the trimmer should run later.
+    /// </summary>
+    int Order { get; }
```
Perksey, line 22: I don't think this should be the way we determine the order of execution - this is not something individual augments should be able to influence without careful consideration of what it's trying to achieve. If the order of execution matters and has different effects, this should be user configurable. As I've said before, I want to get rid of things like `Version` where you arbitrarily change the number to achieve the result you want - if there is a need to control the order of execution to achieve the results the user wants, then that's what the user-facing design should reflect.

sources/SilkTouch/SilkTouch/Mods/PrettifyNames.cs
Perksey: Worth doing at some point though.

generator.json
```diff
@@ -300,13 +361,22 @@
...
         "Namespace": "Silk.NET.Vulkan",
         "FlagsTypes": [
           "VkFlags",
           "VkFlags64"
-        ]
+        ],
+        "ExcludeVendorSuffixIdentification": [
+          "VK_VENDOR_ID_VIV",
+          "VK_VENDOR_ID_VSI",
+          "VK_VENDOR_ID_MESA"
```
Perksey, line 405: What happened here?

Perksey (changes requested):

generator.json
Exanite: Here is the enum with just the relevant code.

These enums are more-so vendor names rather than vendor suffixes
Note that they used `Khronos` instead of `KHR` here.

On the `develop/3.0` branch, these were getting trimmed as `VendorIdVIV` and `VendorIdKhronos`.
Silk 2 also matches the output below.
```cs
// From this PR
public enum VendorId : uint
{
    [NativeName("VK_VENDOR_ID_KHRONOS")]
    Khronos = 65536,

    [NativeName("VK_VENDOR_ID_VIV")]
    Viv = 65537,

    [NativeName("VK_VENDOR_ID_VSI")]
    Vsi = 65538,

    [NativeName("VK_VENDOR_ID_KAZAN")]
    Kazan = 65539,

    [NativeName("VK_VENDOR_ID_CODEPLAY")]
    Codeplay = 65540,

    [NativeName("VK_VENDOR_ID_MESA")]
    Mesa = 65541,

    [NativeName("VK_VENDOR_ID_POCL")]
    Pocl = 65542,

    [NativeName("VK_VENDOR_ID_MOBILEYE")]
    Mobileye = 65543,
}
```

sources/SilkTouch/SilkTouch/Naming/INameTrimmer.cs
Exanite: I fully agree. In this case, I only documented the functionality of the property, but I probably should have also documented the intended usage.

This is the intended usage:
Mods should only use the values `-1` or `0` for `Order` and `DiscriminatorPriority`.
`-1` corresponds to "disabling" the property and `0` corresponds to leaving it as the default value.
All other values should come from the generator config.
Mods also should not attempt to compute what values to use for `Order` and `DiscriminatorPriority`.

The above usage matches how the properties are currently used by the mods in my PR.

Arguably, even `0` and `-1` should not be hardcoded, but I opted not to do that since it would expose a lot of configuration properties.

sources/SilkTouch/SilkTouch/Mods/MarkNativeNames.cs
Exanite: This should be something we do as a post-process step and not at a pre-process step.

MarkNativeNames is designed as a pre-process step. It goes early in the mod order.
MarkNativeNames goes right after ClangScraper because, at that point, we know that all names match the native names (outside some exceptions like in bitfield structs where ClangSharp introduces an extra field).

Attribute removal is currently handled by `StripAttributes`, which is a post-process step. It goes late (last currently) in the mod order.
This mod can be generalized as an `CleanupAttributes` mod where optimizing attributes is part of the configurable steps that the mod applies.

The reason why optimizing attributes should be done as a post-process is because optimization requires knowledge of the output. If we introduce the "don't add if redundant" logic during pre-processing, then no `NativeName` attributes would ever get added.

An alternative is for mods to only add the `NativeName` attribute if they change the name, but that also requires all mods to understand the rules for adding/removing the `NativeName` attribute, which greatly increases complexity.

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
Exanite: This code mostly matches the functionality of the code that was originally in `MixKhronosData.Trim`.
Well... to the best of my ability since that code wasn't really commented either.

The exceptions are that I removed/prevented some edge cases and restructured the code.

Edit: Another exception is that I changed the enum member name trimming condition to be if the member matches the enum type name vendor suffix. Previously it was if **all** member names match the type vendor suffix (hence the config option being called `TrimEnumMemberImpliedVendors` instead of `TrimEnumMemberSharedVendors`). This was the part I discussed with you in the discord.

I provide a better explanation of the functionality of the code in the `Configuration` record struct at the top of the class, but I agree that that this code is likely hard to understand without context and local examples. I'll go through the code and add these.

Perksey:
Oh... lol. Ok yeah maybe this is the appropriate fix.

generator.json
Perksey: Good spot.

Exanite:
As in, combine the two regexes? I think that would make it confusing to edit.

Edit: I think I see what you mean. You mean to make it so that the regex considers if trimming the suffix would trim into part of another word. Eg: Match for `Rewind` before matching for `dv` or `v`.

Perksey:
I'm concerned about only having Rewind in EndingsToTrim if that is indeed the valid fix. I still feel like there should be a way to use EndingsNotToTrim to get what we want but I'm not sure how. But I don't want the two regexes to not enact the correct behaviour because they have different sets of applicable words.

sources/SilkTouch/SilkTouch/Naming/INameTrimmer.cs
Exanite: Oh wait, I completely misunderstood what code this comment was addressing.
I thought it was referring to the `NameAffix` attribute. For that case, my comment still holds and I'll still update the documentation to include the intended usage of the attribute.

As for the trimmer case though, I'm not sure.
The order property here, as it currently stands, is for ensuring correctness rather than for allowing for customizability. The affixes have to trimmed in a certain order for my NameAffix system to work, as seen here.
image

However, for trimmers not part of the required set (eg: the trimmers defined by mods), those probably should have their own configuration.

Exanite: The only external trimmer (external as in not part of PrettifyNames) here is MixKhronosData.
This means we currently don't really have a usage pattern to analyze.

What other trimmers will there be that isn't covered by the existing functionality?
Even `NameTrimmer` can probably use the name affix system instead, which has a well defined and configurable execution order. From what I understand, it mainly identifies and trims global prefixes, so it can be implemented as a `IdentifyGlobalPrefixes` mod that has a syntax rewriter. This also lets configure whether global prefixes should be identified; if they should be identified, but kept; and if they should be kept, but reordered.

MixKhronosData does provide one case, which is to prevent/rewind trimming in the case of `GLEnum`.
Generally speaking, this would be more of a name replacement/transformation. This is not covered by the affix system, but is partially covered by the name override system.

sources/SilkTouch/SilkTouch/Mods/MarkNativeNames.cs
Perksey: Hmmmmm ok, I agree - I must admit I haven't checked the bindings yet so I don't know whether this is being done already.

Exanite: The optimization of NativeName attributes isn't, but it's not too hard to implement.
To me, this is a lower priority task since it's not critical to the bindings working.

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
Exanite: I've updated the code to answer these questions. Same for the similar review comments in this area of the code.

This is how this section now looks:
image

I'm not exactly sure what you mean by breaking changes since this behavior exactly matches the original behavior of the code as it as in MixKhronosData.Trim. Keep in mind that this isn't new functionality. This is old functionality ported over to the new name affix system.

That being said, both the old and new implementations do have the possibility of introducing breaking changes if upstream introduces another variant of the enum that prevents the vendor suffix from getting trimmed.

For example, CombinerRegisterNV currently exists and has the NV vendor suffix trimmed due to it containing GL_TEXTURE0_ARB, similar to this example:
image

If Khronos was to add CombinerRegisterAMD, then CombinerRegisterNV's suffix would no longer be trimmed and CombinerRegister would no longer exist as an enum in the generated bindings.

That being said, this is probably a rare case.

Note that if Khronos added CombinerRegister, there likely would be no breaking change.

Exanite:
For the record, the PR is ready to review again, but I'm currently doing some additional reviewing of the generated bindings to ensure that I haven't changed anything that I didn't mean to change.

sources/SilkTouch/SilkTouch/Mods/PrettifyNames.cs
```diff
@@ -1,10 +1,8 @@
-﻿using System.Collections.Concurrent;
-using System.Diagnostics;
+﻿using System.Diagnostics;
```
Exanite, line 1: Ignore the location of this review comment.

While reviewing the generated bindings, I found that my changes lead to certain functions, such as `alGetErrorDirect`, to be output as `GetErrorDirectRawDirect` instead of `GetErrorDirectRaw`.

Not sure what the cause is yet. Will investigate.
Edit: Looks like the `Raw` suffix is being added by `FunctionTransformer`, but my name affix system doesn't know about it.

Exanite: This is an easy enough fix, but does reveal an issue with how I designed the name affix system: anything that adds affixes needs to add configuration options so that the user can configure those affixes.

I have a decent re-design planned out, but we should discuss it first.

The re-design is to change it so that the NameAffix attribute specifies the affix category:
Eg:
`NameAffix(type: "Suffix", category: "KhronosDataType", affix: "i")`
instead of
`NameAffix(type: "Suffix", affix: "i", order: 0, discriminatorPriority: 0)`

This means mods can no longer specify the order/discriminator values.
In other words, mods can only identify the affixes.
Instead, order/discriminator will be entirely specified by the user as part of the PrettifyNames mod configuration.
If not specified, order will default to 0 and discriminator will default to -1.
```json5
"PrettifyNames": {
  "Affixes": {
    "HandleType": {
      "Order": 1
    },
    "KhronosDataType": {
      "Discriminator": 0
    },
    // Other configs as needed
    // Eg: -1 order for KhronosNonExclusiveVendor, KhronosImpliedVendor
  }
}
```

The main downside I can think of for this is that this might lead to some verbose configs.
Otherwise, this is very flexible.

Perksey: Sounds like a good plan!

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
```diff
@@ -1807,12 +1520,14 @@ public bool TryGetSymbolMetadata(
...
     /// </summary>
+    // Rewindv is to prevent Rewindv from being trimmed as Rewin
+    // TODO: Consider replacing this with something that prevents a list of words from being trimmed into instead of not being trimmed at all
     [GeneratedRegex(
         "(sh|ib|[tdrey]s|(?<![A-Z])[eE]n[vd]|bled|Attrib|Access|Boolean|Coord|Depth|Feedbacks|Finish|Flag|"
             + "Groups|IDs|Indexed|Instanced|Pixels|Queries|Status|Tess|Through|Uniforms|Varyings|Weight|Width|Bias|Id|"
             + "Fixed|Pass|Address|Configs|Thread|Subpass|Deferred|Extended|Affix|Annex|Box|Aux|Ex|Index|Vertex|Path|"
             + "Arch|Arith|Afresh|Both|High|Math|Mesh|Sinh|Bench|Brush|Bunch|Crash|Flush|Depth|Latch|Morph|Pinch|"
-            + "Pitch|Stretch|Smooth|Matrix|Radix|Sound|Rewind|Supported)$"
+            + "Pitch|Stretch|Smooth|Matrix|Radix|Sound|Supported|Rewind|Rewindv)$"
```
Perksey, line 1513: I'm still not overly comfortable with this, can we add improving this to the to-do list for a subsequent PR? It's probably just as simple as repeating the `EndingsToTrim` as an optional match group in this regex.

Right now I'm very conscious that we're only applying this workaround to this specific word, where there are clearly others that could be matched like this in `EndingsNotToTrim`

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
```diff
@@ -1887,53 +1602,42 @@ private class RewriterPhase1(JobData job, ILogger logger) : CSharpSyntaxRewriter
...
                                                         SeparatedList(
                                                             groupInfo.Enums.Select(x =>
-                                                                EnumMemberDeclaration(
-                                                                        x.Identifier.ToString()
-                                                                    )
-                                                                    // TODO actually eval the expression to see if necessary?
+                                                                EnumMemberDeclaration(x.Identifier.ToString())
+                                                                    .WithAttributeLists(
+                                                                        new SyntaxList<AttributeListSyntax>()
+                                                                        .WithNativeName(x.Identifier.Text))
```
Perksey: nit: please ensure you're running the CSharpier code formatter! Rider and VSCode have functionality to run it on save with the correct plugins installed

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
Perksey: Thanks!

sources/SilkTouch/SilkTouch/Mods/TransformEnums.cs
```diff
@@ -219,9 +219,7 @@ private class Rewriter(Configuration config, List<EnumMemberFilter> removeMember
                     .ToList();
             }
 
-            var isFlagsEnum = node.AttributeLists.SelectMany(list => list.Attributes)
-                .Any(attribute => attribute.IsAttribute("System.Flags"));
-
+            var isFlagsEnum = node.AttributeLists.ContainsAttribute("System.Flags");
```
Perksey, line 234: Not relevant for this PR I don't think, but note that whether enums come out of `ClangScraper` with `Flags` attributes is a bit hit or miss.

BuildTools has the function [`IsProbablyABitmask`](https://github.com/dotnet/Silk.NET/blob/c87fc664cc5d15abd754997b0a9ad138470554c4/src/Core/Silk.NET.BuildTools/Common/Utilities.cs#L627) to add the `Flags` attribute where it looks like it's needed, but again this probably isn't relevant for this PR. Just something to bear in mind.

sources/SilkTouch/SilkTouch/Naming/INameTrimmer.cs
Perksey: `NameTrimmer` and `MixKhronosData` should be assumed to operate in a configurable order. For the former, I don't think this is implemented today, but I had hoped that you could dependency inject many `INameTrimmer`s and configure which ones and in which order they run in, as they have different effects. I kept `NameTrimmer217`/`NameTrimmer218` in a hope that it sort of demonstrated that, but even that was butchered with my introduction of the `Version` property (which is also a smell and needs to be eradicated)

_(aside: Personally, given that all of these trimmers have the express function of _shrinking_ a name, I'm a bit concerned if we're not able to make this happen without order dependencies. But this is probably a naive statement)_

The magic ordering numbers just feel off and have no place in a public API like `INameTrimmer`. The `INameTrimmer` API has no concept about how it's being used, this is all realised within `PrettifyNames`. Therefore, it should be `PrettifyNames` determining the order of operations and not the `INameTrimmer` API. I appreciate we have no use cases to analyse right now, but mechanisms like this always end up in an arms race of who has the highest/lowest number in a world where we have lots of competing `INameTrimmer`s and I don't want to approve an API that exposes us to those sorts of problems.

For the avoidance of doubt: while the SilkTouch API is not a formal deliverable (and therefore not bound by the proposals process), it is still something that external packages are expected to develop against (unlike 2.X, i.e. to develop their own mods) and I don't want to be changing it more frequently than we have to as a result. These sorts of review comments may seem like "grand stances" (apologies!) but they're all to that end. _(at the same time, I do not have all the answers, and I am only trying to identify what feels like a potential smell waiting to bite us later, I could be wrong and please do tell me if I am definitely wrong)_

I really don't want to back down from this but would appreciate further commentary as it is possible I'm being way too pernickety about this.

sources/SilkTouch/SilkTouch/Mods/MarkNativeNames.cs
Perksey: Ok, please ensure we do do this though.

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
Perksey: The new comments explain this much better, thanks!

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
Perksey: The comments help. And yeah, you're right, this does introduce breaking changes in that if you have `BufferTargetARB` in one version of the spec, and update to one that for some reason has `BufferTargetNV`, then you've got an API break because this code will revive the suffix for `BufferTargetARB` after it had previously been trimmed to `BufferTarget`.

But given that OpenGL is the only case where I expect this will be used, I'm not too worried about it. That was the case I was worried about though, generally I strive to make our API maximally resistant to changes like these.

sources/SilkTouch/SilkTouch/Mods/PrettifyNames.cs
Perksey: Will review Exanite/Silk.NET#21 which marks this comment off

Perksey (commented):
That review felt like a breeze, thanks so much for your hard work!

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
Exanite: Yup, I added it to the todo list last time we discussed this.
image

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
Exanite: Ah, that's good to know. I'll look into running it for all of the files I've modified in this PR.

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
Perksey: Awesome, thanks. I suspect there's something a little bit fundamental about the approach e.g. maybe we should trim just one data type at a time and compare with `EndingsNotToTrim` after every trim. Little things to play with later.

sources/SilkTouch/SilkTouch/Mods/TransformEnums.cs
Exanite: Ah, that might be worth adding to the TransformEnums mod. I currently use the Flags enums added by ClangSharp and the KnownBitmask property from MixKhronosData's group information.

sources/SilkTouch/SilkTouch/Naming/INameTrimmer.cs
Perksey: Just updating this to reflect the [Discord discussion](https://discord.com/channels/521092042781229087/587346162802229298/1457867211841474807): current thinking is to reuse the affix system for the removal of the identified prefixes which is believed to alleviate the requirement for `INameTrimmer.Order`.

Perksey: Merging on condition of future completion of this task.

Perksey (approved):
Approved once marked ready for review... lol.

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
Perksey: Would like this investigated in a future PR. (Pretty sure it's already on the todo list)

Perksey (approved):

PR #2504, opened 2025-11-27
3.0 - Simple IHandle interface for XYZHandle structs

aquagoose:
# Summary of the PR
Adds a simple IHandle struct for XYZHandle structs, and implements an `IsNull` property.

https://discord.com/channels/521092042781229087/587346162802229298/1443561471245291592

# Further Comments
Only tested locally with SDL.

### DIFF
--- a/sources/Core/Core/Abstractions/IHandle.cs
+++ b/sources/Core/Core/Abstractions/IHandle.cs
@@ -1,0 +1,9 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Silk.NET.Core;
+
+public interface IHandle
+{
+    public bool IsNull { get; }
+}
--- a/sources/SilkTouch/SilkTouch/Mods/TransformHandles.cs
+++ b/sources/SilkTouch/SilkTouch/Mods/TransformHandles.cs
@@ -513,6 +513,11 @@ public class TransformHandles(IOptionsSnapshot<TransformHandles.Config> config, 
                         Token(SyntaxKind.UnsafeKeyword),
                         Token(SyntaxKind.PartialKeyword)
                     )
+                )
+                .WithBaseList(
+                    BaseList(
+                        [SimpleBaseType(ParseTypeName("IHandle"))]
+                    )
                 );
         }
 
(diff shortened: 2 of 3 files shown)


sources/Core/Core/Abstractions/IHandle.cs
```diff
@@ -0,0 +1,9 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Silk.NET.Core;
+
+public interface IHandle
```
Perksey, line 6: Missing docs comments are resulting in compilation failure.

sources/SilkTouch/SilkTouch/Mods/TransformHandles.cs
```diff
@@ -700,6 +705,23 @@ private static IEnumerable<MemberDeclarationSyntax> GetDefaultHandleMembers(stri
...
+                    PredefinedType(Token(SyntaxKind.BoolKeyword)),
+                    Identifier("IsNull")
+                )
+                .WithModifiers(TokenList(Token(SyntaxKind.PublicKeyword)))
+                .WithExpressionBody(
+                    ArrowExpressionClause(
+                        BinaryExpression(
+                            SyntaxKind.EqualsExpression,
+                            CastExpression(IdentifierName("IntPtr"), IdentifierName("Handle")),
+                            LiteralExpression(SyntaxKind.NumericLiteralExpression, Literal(0))
```
Perksey, line 720: Could we do `this == default` to make it agnostic to non-pointer-sized handles?

sources/Core/Core/Abstractions/IHandle.cs
Perksey, line 6: Hmmm on second thought, maybe we could do `IHandle<TSelf> : IEquatable<TSelf>, IEqualityOperators<TSelf, TSelf, bool>`? At this point we totally negate the need for `IsNull` as users generalising over `T: IHandle<T>`>? Could maybe keep this non-generic interface and DIM it in the generic one if we want (is that a thing?) If we go that route no changes will be required to TransformHandles other than the base list.

Perksey (changes requested):

Perksey:
Is this still being worked on?

sources/Core/Core/Abstractions/IHandle.cs
Exanite: I recently worked on some unofficial HarfBuzz bindings using Silk 3 and HarfBuzz uses a null object pattern, which can be exposed as an `IsNullOrEmpty(IHarfBuzz)` or `IsEmpty(IHarfBuzz)` _method_ on the HarfBuzz handles. This means having the `IsNull` _property_ would be nice for consistency.

To be clear, not asking for the methods to be added to the `IHandle` interface. There likely will need to be a HarfBuzz-specific handle interface.

https://harfbuzz.github.io/object-model-lifecycle.html
> Finally, object constructors (and, indeed, as much of the shaping API as possible) will never return NULL. Instead, if there is an allocation error, each constructor will return an “empty” object singleton.

HarfBuzz exposes these "empty" singletons as `*GetEmpty()` functions:
image

This matters more if we ever add HarfBuzz bindings officially to Silk, but does represent a potential edge case.

Also, open to having both the equality interfaces and the property/methods, but I have to give it some more thought.

PR #2511, opened 2025-12-13
Update GLFW binaries

silkdotnet:


Perksey:
We haven't actually updated the version in use, closing to reduce risk.

PR #2523, opened 2025-12-20
[3.0] Rewrite the SilkTouch cache implementation to actually work

Perksey:
Does what it says on the tin.

This is basically so we can more rapidly iterate on larger bindings like Win32.


Perksey:
The Linux/Windows tests and the website build will be fixed once #2526 lands and I've updated to use the latest experimental build.

PR #2525, opened 2026-01-02
[3.0] Fix EnumGroupRegression test case

Exanite:
# Summary of the PR
This prevents the `Constants_*` and `enums_*` groups from being output for the OpenCL XML spec in the EnumGroupRegression test case. The change to `cl_compiler_mode_altera` is justifiable since it is an empty enum.

Note that empty enums still contain important information, such as whether they are `[Flags]` types or not.
These notably affect the generation of the Vulkan bindings where ignoring these empty enums causes some enums to turn into non-Flags types.

Discord discussion: https://discord.com/channels/521092042781229087/587346162802229298/1454577375063970045

# Further Comments


Perksey (approved):
yes

PR #2534, opened 2026-01-12
[3.0] Improve handling of Khronos enum native names

Exanite:
# Summary of the PR

This changes the native name for namespaced enums like `GLEnum` to be `GLenum`.
In the process, I also found a bug where Vulkan enums had inconsistent native names.

Discord conversation: https://discord.com/channels/521092042781229087/587346162802229298/1460042806952198310
PR where this issue was first discovered: https://github.com/dotnet/Silk.NET/pull/2503

This PR contains the following sub-PR:
- https://github.com/Exanite/Silk.NET/pull/30

# Further Comments

N/A


sources/Vulkan/Vulkan/Enums/DataGraphPipelineDispatchFlagsARM.gen.cs
```diff
@@ -8,7 +8,7 @@
 
 namespace Silk.NET.Vulkan;
 
-[NativeName("VkDataGraphPipelineDispatchFlagBitsARM")]
+[NativeName("VkDataGraphPipelineDispatchFlagsARM")]
```
Exanite, line 11: The rest of the Vulkan enums use Flags instead of FlagBits for the native name.

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
```diff
@@ -2176,8 +2197,8 @@ internal void ReadGroups(XDocument doc, JobData data, HashSet<string> vendors)
             }
 
             // Vulkan/OpenXR enum name
-            nativeName ??= groupName;
             groupName = groupName?.Replace("FlagBits", "Flags");
+            nativeName ??= groupName;
```
Exanite, line 2274: This is because we want the Flags version of the name to be the native name.

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
```diff
@@ -1986,16 +1986,18 @@ public override SyntaxNode VisitDelegateDeclaration(DelegateDeclarationSyntax no
         // Trimming is done by PrettifyNames, if PrettifyNames is configured to do so
         public override SyntaxNode VisitEnumDeclaration(EnumDeclarationSyntax node)
         {
-            var typeName = node.AttributeLists.GetNativeNameOrDefault(node.Identifier);
-            var groupInfo = job.Groups.GetValueOrDefault(typeName);
+            var nativeTypeName = node.AttributeLists.GetNativeNameOrDefault(node.Identifier);
+            var managedTypeName = node.Identifier.Text;
 
-            var typeVendor = job.Vendors.FirstOrDefault(typeName.EndsWith);
+            var groupInfo = job.Groups.GetValueOrDefault(managedTypeName);
```
Exanite, line 2033: The `job.Groups` dictionary is keyed using the managed type name so we need to make the distinction here.

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
```diff
@@ -2139,16 +2142,34 @@ public override SyntaxNode VisitMethodDeclaration(MethodDeclarationSyntax node)
...
+        //
+        // Each API has slight variations in how the XML is structured.
+        //
+        // Main enum group:
+        // Either has the "group" and "namespace" properties (OpenGL-style).
+        // Or has the "name" property.
+        //
+        // Enum member:
+        // "group" property spcifies a list of additional OpenGL-style groups the enum member belongs to.
+
```
Exanite, line 2210: This is to provide context on some terminology used and also highlights some differences that anyone in the future should know about.

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
```diff
@@ -2207,7 +2228,28 @@ static bool IsUngroupable(string groupName) =>
...
+            // 2. Empty groups are recorded properly
+            if (
+                groupName != null
+                && !IsUngroupable(groupName)
+                && !data.Groups.ContainsKey(groupName)
+            )
+            {
+                data.Groups[groupName] = new EnumGroup(
+                    groupName,
+                    nativeName,
```
Exanite: This is the core fix/change. This change in initialization order is important because the loop below initializes the enums if it hasn't already been initialized and initializes it with *slightly* different information if the XML is for an OpenGL style API (`groupName`/`nativeName` specifically refer to the namespace enum; `group` refers to to any group the member is part of). Totally not an edge case.

tests/SilkTouch/SilkTouch/Khronos/MixKhronosDataTests.cs
```diff
@@ -448,7 +449,7 @@ public enum GLEnum { }
                             "GLEnum",
                             new MixKhronosData.EnumGroup(
                                 "GLEnum",
-                                "GLEnum",
+                                "GLenum",
```
Exanite: Changed for consistency. This second parameter refers to the native name.

tests/SilkTouch/SilkTouch/Khronos/MixKhronosDataTests.cs
```diff
@@ -424,6 +424,7 @@ public async Task IdentifiesNamespaceEnumPrefix()
             .AddDocument(
                 "GLEnum.gen.cs",
                 """
+                [NativeName("GLenum")]
```
Exanite, line 427: Added because MixKhronosData prioritizes the native name for identifying namespace enums. Because of this, the test breaks without the NativeName attribute. Whether it prioritizes it or not doesn't really matter, but I've been moving towards making the codebase prefer the native name where possible since the native name is more stable.

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
```diff
@@ -2236,12 +2278,12 @@ static bool IsUngroupable(string groupName) =>
...
-                        .Concat(glGroups ?? [])
-                        .Concat(block.Attribute("group")?.Value is { Length: > 0 } g ? [g] : [])
-                        .Distinct()
-                )
+                // Add the enum member to the namespace enum, the main enum group, and its additional OpenGL-style groups
+                var memberGroups = (groupName is null ? Enumerable.Empty<string>() : [groupName])
+                    .Concat(block.Attribute("group")?.Value is { Length: > 0 } g ? [g] : [])
+                    .Concat(glGroups ?? [])
+                    .Distinct();
+                foreach (var group in memberGroups)
```
Exanite: I moved the LINQ here out of the loop for easier debugging. I also reordered the lines to better match my new comment.

Exanite (commented):
Self review complete. Marking as ready for review and subsequent merge.

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
```diff
@@ -2207,7 +2228,28 @@ static bool IsUngroupable(string groupName) =>
...
+            if (
+                groupName != null
+                && !IsUngroupable(groupName)
+                && !data.Groups.ContainsKey(groupName)
+            )
+            {
+                data.Groups[groupName] = new EnumGroup(
+                    groupName,
+                    nativeName,
+                    anyGLStyleGroups ? "GLenum" : null,
```
Perksey: Are we sure that hardcoding `GLenum` is correct here?

Exanite: This parameter specifies the base type of the enum and matches the code in the "Parse enum members" for loop below.

MixKhronosData has a type map during `InitializeAsync` that is hardcoded to add `GLenum` as an entry:
```cs
job.TypeMap.TryAdd("GLenum", "uint");
```

Not specifying the type in the `new EnumGroup` code means that the enum group is skipped when MixKhronosData goes to generate the enum later:
```
Enum "ALEnum" has no base type. Please add TypeMap entry to the configuration. This enum group will be skipped.
```

Compared to before, this is now needed because this line of code takes precedence over the `new EnumGroup` code in the "Parse enum members" for loop, meaning that it needs to take responsibility for initializing the base type in certain cases.

Perksey: I think this is fine. We can remove the hardcoded GLenum in the TypeMap for consistency if desired, the intention behind this was to have a "sensible default" for the configuration.

I don't want the mod to be hardcoded to OpenGL specifically, as OpenAL is one example of a GL-style binding. GLenum is not an OpenAL type and our code doesn't have any business assuming that ALEnum is the same as GLenum and therefore uint, considering the rest of the code is agnostic to this.

Exanite: That makes sense. I assumed the intention behind using `GLenum` specifically was to indicate that the enum is a GL-style enum, not specifically that it was a `GLenum`.

`GLbitfield` is also declared in the type map, but it is not used anywhere.

It looks like the code that handles OpenCL later just hardcodes the base type, so this change would be consistent with that approach:
image

Proposed change:
image

Also debating if I want to pull the result of the condition into a variable so we don't have issues when we modify one, but not the other in the future. However, I don't want to wrongly imply that the resulting type needs to be used as a default (there are four places where we use `new EnumGroup`).

Perksey: I believe both of these cases should not hardcode a specific integer type. The OpenCL one is likely a leftover from the admittedly quite complex OpenCL-specific logic ported from BuildTools, I agree this new change is consistent but hardcoded logic for determining what the enum base type should be should be avoided, as this changes depending on what we're binding to and therefore should be specific in the configuration, optionally with "sensible defaults" in MixKhronosData that the configuration overrides.

Exanite: Do you have a specific change in mind?

One approach for OpenGL/AL is to have the type be based on the namespace (eg: `GLenum`, `ALenum`, etc).
I'm unfamiliar with OpenCL so I'm not sure if the same approach will work there.

That being said, it should be "safe" to leave the type map unspecified (as in, not provide defaults for the config to override) since the generator will log errors and skip over enums that don't have base types. Anyone working on the bindings should be able to quickly see that and fix it by updating the configuration.

Perksey: Leaving it unspecified and updating the configuration is what I had in mind.

That being said, ideally we don't want every single enum to be specified in the config, so we do really only want it to be the base types. For the OpenGL code, this should be an easy case of using `$"{namespace}enum"`.

For the OpenCL code, I'm not quite sure how to make this more generic primarily because that logic is so esoteric and there's a lot hardcoded already, I would just suggest hardcoding this to `cl_bitfield` and we'll have to look at it in depth another time unless you can see something obvious that follows the same spirit as the OpenGL change.

Perksey: Wrt OpenCL, I think removing all the hardcoded stuff is likely a larger work item, so the best we can do is not make this problem worse and tackle it another time.

The changes likely aren't that complicated:
- [This loop](https://github.com/Exanite/Silk.NET/blob/fix/incorrect-native-name-glenum/sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs#L2447) removes knowledge of `cl_properties`/`cl_bitfield`, and instead of checking for those specific types should check whether the inner `type` element has a name in `TypeMap`.
- Configuration should contain a bitmask types array, that if not provided has sensible defaults (i.e. `cl_bitfield`)
- Additional work to either use other data within the registry in place of hardcoded `cl`/`CL` in strings, or to derive this from configuration.

But your call on how much to do for this PR, it does feel out-of-scope.

Exanite: I haven't looked into OpenCL yet, but I agree, I'd much rather do this once we start working on the OpenCL bindings so we can review the output and work with actual data. Trying to rework the OpenCL code without looking at the output seems like a recipe for disaster.

curin (approved):

PR #2535, opened 2026-01-13
[3.0] Improve handling of affixed names when singularizing (OES/SGIS issue) and expose NameAffix API

Exanite:
# Summary of the PR

This fixes the interaction between Khronos vendor suffixes and the Singularize method, which led to cases like the following:

image

In order to fix this, this also exposes the NameAffix API, allowing mods to read NameAffix attributes and process the name without the affixes attached.

Original Discord discussion: https://discord.com/channels/521092042781229087/1376331581198827520/1456785841753035005

# Further Comments

Todo list:

- [x] Prevent `OES` and similar from being singularizing into `O` in `ArrayParameterTransformer.TransformArrayParameterRewriter`.
   - Eg: `glFeedbackBufferxOES` 
   - Discussed solution is to expose the name affix API so that `TransformArrayParameterRewriter` can singularize the base name instead of the full name.
   - [x] `OES` (incorrect version is `OOES`)
   - [x] `SGIS` (incorrect version is `SgiSGIS`)
- [x] Allow mods to read `NameAffix` data.
- [x] Allow mods to strip a name of its affixes
- [x] Allow mods to re-apply a list of affixes to the name, but without allowing mods to deeply configure the process
   - Reasoning is that more involved transformations should be restricted to PrettifyNames.
   - This is because PrettifyNames can be configured by the generator user, while other mods must expose their own (possible duplicate) configuration.
- [x] Consider updating `AddNameAffix` API to use the new `NameAffixType` enum for consistency and simplicity


sources/SilkTouch/SilkTouch/Mods/Common/AttributeUtils.cs
```diff
@@ -203,69 +203,6 @@ out _
...
-    /// </summary>
-    /// <param name="attributeLists">The attribute lists to add the attribute to.</param>
-    /// <param name="category">The affix category.</param>
-    /// <param name="prefix">The value of the affix.</param>
-    /// <param name="addToInner">
-    /// Use true if the affix comes from the inside of the name.
-    /// Use false if not (outside or appended to end).
-    /// True means that the attribute is added to the start of the attribute list, meaning that the affix is re-appended earlier.
-    /// </param>
-    public static SyntaxList<AttributeListSyntax> AddNamePrefix(
```
Exanite, line 217: I moved these to the new `NameAffixer` static class to keep everything related to name affixes together.

I also changed the API to use the new `NameAffixType` enum and changed it so that we only have the `AddNameAffix` method instead of `AddNamePrefix/Suffix` separately.

sources/SilkTouch/SilkTouch/Mods/Transformation/ArrayParameterTransformer.cs
```diff
@@ -231,15 +228,16 @@ class TransformArrayParameterRewriter(
...
+        public override SyntaxNode VisitMethodDeclaration(MethodDeclarationSyntax node)
+        {
+            var identifier = node.Identifier.Text;
+            var affixes = node.AttributeLists.GetNameAffixes();
+            var singularizedIdentifier = NameAffixer.ApplyAffixes(
+                NameAffixer.StripAffixes(identifier, affixes).Singularize(),
+                affixes
+            );
+
+            return node.WithIdentifier(Identifier(singularizedIdentifier))
```
Exanite, line 240: This is the core fix for the OES/SGIS singularize issue.

sources/SilkTouch/SilkTouch/Mods/PrettifyNames.cs
```diff
@@ -870,13 +870,6 @@ private record struct RenamedType(
         Dictionary<string, string> Functions
     );
 
-    private record struct NameAffix(
-        bool IsPrefix,
-        string Category,
-        string Affix,
-        int DeclarationOrder
```
Exanite, line 877: This been moved to a separate file as a public type.

sources/SilkTouch/SilkTouch/Mods/PrettifyNames.cs
```diff
@@ -958,52 +951,13 @@ _typeInProgress is not null
             || _enumInProgress is not null
             || node.Ancestors().OfType<BaseTypeDeclarationSyntax>().Any();
 
-        private bool TryGetAffixData(
```
Exanite, line 961: Replaced by `NameAffixer.GetNameAffixes()`, which returns an array directly instead of the TryGet pattern. The array is empty in the "false" case and does not allocate.

sources/SilkTouch/SilkTouch/Mods/PrettifyNames.cs
```diff
@@ -1286,7 +1241,7 @@ public override SyntaxNode VisitMethodDeclaration(MethodDeclarationSyntax node)
 
     /// <param name="affixTypes">The affix data retrieved by the <see cref="Visitor"/>.</param>
     /// <param name="config">The configuration from <see cref="Configuration.Affixes"/>.</param>
-    private class NameAffixer(
+    private class PrettifyNamesAffixer(
```
Exanite, line 1244: This class is the `NameAffixer`, but specialized for the needs of `PrettifyNames`.

sources/SilkTouch/SilkTouch/Naming/NameAffixer.cs
```diff
@@ -0,0 +1,241 @@
...
+                    affixes =
+                    [
+                        .. affixes,
+                        new NameAffix(
+                            type == "Prefix" ? NameAffixType.Prefix : NameAffixType.Suffix,
+                            category,
+                            affix,
+                            declarationOrder
+                        ),
+                    ];
```
Exanite, line 53: Not sure what is more performant here. Most names right now have 0 affixes so this method won't allocate at all.

Later, if we change NameTrimmer to be a mod that adds name affixes, most names will have 1 affix (the shared prefix), which would allocate, but would be the minimal amount.

Also not sure if this gets stack allocated or not by .NET 10. I need to look up the escape analysis rules.

sources/SilkTouch/SilkTouch/Naming/NameAffixType.cs
```diff
@@ -0,0 +1,20 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace Silk.NET.SilkTouch.Naming;
+
+/// <summary>
+/// The type of name affix.
+/// </summary>
+public enum NameAffixType
```
Exanite, line 9: I decided to add this enum since the name affix API is now fully exposed.

Exanite (commented):
Self review completed

curin (approved):

PR #2555, opened 2026-03-24
[3.0] Add referenced affix system to improve handling of nested struct names and similar names

Exanite:
# Summary of the PR

## The problem

Nested struct names are an edge case in the name affix system because they have the following format:
```
NestedStructName = ParentBaseName + ParentSuffix + FieldBaseName + FieldSuffix
```

Currently, nested struct names are generated by directly combining the full parent type name with the full field name.
This leads to issues where the ParentSuffix is not handled correctly.

For example, `PerformanceCounterDescriptionArmName` should be `PerformanceCounterDescriptionARMName`.

## The solution

This PR adds a "referenced affix" system where name affixes can reference the names of other symbols.
This is based on the observation that these names are compound names where part of the name makes use of a name of another symbol.

For example, `PerformanceCounterDescriptionARMName` is annotated as the following:
```cs
[NativeName("_name_e__FixedBuffer")]
[InlineArray(256)]
[NameAffix("Prefix", "NestedStructParent", nameof(PerformanceCounterDescriptionARM))]
[SupportedApiProfile("vulkan")]
public partial struct PerformanceCounterDescriptionARMName
{
    // ...
}
```

This allows PrettifyNames to use the final value of the referenced name (`PerformanceCounterDescriptionARM`) as the final value for the affix. This means all changes are automatically synchronized between the two names.

Original discussion: https://discord.com/channels/521092042781229087/587346162802229298/1460017726813900975

# Further Comments

## Tasks

- [x] Improve handling of nested struct names
	- Eg: `PerformanceCounterDescriptionArmName` should be `PerformanceCounterDescriptionARMName`
	- Original conversation here: https://discord.com/channels/521092042781229087/587346162802229298/1460017726813900975
- [x] Consider splitting `ExtractNestedTyping` and `TransformHandles` into a new set of `Extract-` mods.
    - This this mainly for maintainability for `ExtractNestedTyping`, but for `TransformHandles` it is useful for `AddApiProfiles` to be executed after all types are extracted, but not yet transformed.
    - [x] Split `TransformHandles` into `TransformHandles` and `ExtractHandles`
    - ~~Also split up `ExtractNestedTyping`~~
       - Note: Currently holding off on this since this has performance implications.
- [x] Address `INameTrimmer.Order/Version` issue (we don't want magic numbers) and consider removing `INameTrimmer` entirely.
- [x] Rename `INameTrimmer` to `INameProcessor` and update names where relevant
   - [x] Do a second documentation pass to ensure that all docs and variable names are updated
- [x] Regenerate all bindings using Windows
- [x] ~~Update OpenAL name overrides~~
   - About half of the overrides can now be removed since they are now handled automatically by the name affix system.
   - Edit: Hmm, the override system actually conflicts here. The current name processing system can't see overridden names since they get removed from the name container at the very start.
      - This means that overrides are not properly taken into account by referenced affixes.
      - Eg: `ALBUFFERCALLBACKTYPESOFTDelegate` uses the native name of `BufferCallbackSOFT` (native name `ALBUFFERCALLBACKTYPESOFT`) as a prefix. This should allow `PrettifyNames` to replace `ALBUFFERCALLBACKTYPESOFT` in `ALBUFFERCALLBACKTYPESOFTDelegate` with `BufferCallbackSOFT`, thus resulting in `BufferCallbackSOFTDelegate`. However, this is not currently the case because the name lookup fails.
      - Idea: Somehow handle the referenced affix + name override edge case. Easy way out is to either manually override or have the name processor look at the name overrides dictionary, but both are hacky.
      - Decision: Handle this by preferring the value from the current name container and falling back to the "final" names dictionary (which needs to be exposed to the processors).
      - This will be done during #2564
- [x] Add unit tests
   - [x] Referenced affix cycle detection
   - [x] Referenced affix missing reference
   - [x] Referenced affix exists, but is in different container
   - [x] Referenced affix success case


sources/SilkTouch/SilkTouch/Naming/NamePrettifier.cs
```diff
@@ -31,11 +31,6 @@ public class NamePrettifier(int longAcronymThreshold)
     /// <exception cref="InvalidOperationException">Thrown when the input or output is an empty identifier.</exception>
     public string Prettify(string identifier, bool allowAllCaps = false)
```
Exanite: The changes are effectively make Prettify a method that prettifies C# identifier *fragments* rather than full names. This is because Prettify is mainly used by PrettifyNames to handle the processing of the "base" name, rather than the full name (which includes prefixes/suffixes). This makes it so that we can prettify fragments independently and have more control over the name transformation process.

sources/SilkTouch/SilkTouch/Mods/ExtractHandles.cs
```diff
@@ -0,0 +1,324 @@
...
+namespace Silk.NET.SilkTouch.Mods;
+
+/// <summary>
+/// Adds empty handle structs by searching for missing types referenced through pointers.
+/// If all references to the missing type are through a pointer,
+/// that missing type is then added as an empty struct.
+/// <para/>
+/// See <see cref="TransformHandles"/> for applying further transformations.
+/// </summary>
+public class ExtractHandles(ILogger<ExtractHandles> logger) : Mod
```
Exanite: Effectively a cut and paste from TransformHandles.

PR #2565, opened 2026-04-10
[3.0] Rework naming system

Exanite:
# Summary of the PR

This PR combines my 3 previous naming-related PRs into a single PR.
This is because the 3 PRs contain interdependent changes and should be merged as a single unit.
That said, each of the 3 PRs focus on different aspects of the naming system and serve as supplementary documentation for the different changes.

PRs contained by this PR:
- https://github.com/dotnet/Silk.NET/pull/2555
- https://github.com/dotnet/Silk.NET/pull/2557
- https://github.com/dotnet/Silk.NET/pull/2564

Discord thread where I noted down my thoughts and decisions: https://discord.com/channels/521092042781229087/1485943448069738608

# Further Comments

This PR does not introduce any new changes on top of the changes made in the 3 contained PRs.
However, prefer reviewing the diff through this PR since this is the PR that is intended to be merged.


sources/OpenAL/OpenAL/al/BufferCallbackSOFT.gen.cs
```diff
@@ -19,7 +19,7 @@ namespace Silk.NET.OpenAL;
 
     public BufferCallbackSOFT(delegate* unmanaged<void*, void*, int, int> ptr) => Pointer = ptr;
 
-    public BufferCallbackSOFT(BufferCallbackDelegateSOFT proc) =>
```
Exanite, line 22: Changed to match my decision here: https://discord.com/channels/521092042781229087/1485943448069738608/1487577229121818766

In short, this reordering is because `-Delegate` is a variant or part of an existing type, similar to how inline array structs are part of their parent struct.

Notably, handle structs are named differently (`BufferHandleEXT`). This is because the `-Handle` suffix is a modification to an existing type.

This is a rule that we can consistently follow and also improves discoverability in IDEs since extensions to a type share a common prefix.

sources/SilkTouch/SilkTouch/Mods/Common/ModLoader.cs
```diff
@@ -16,24 +16,26 @@ public class ModLoader
     public static Type? LoadModByName(string name) =>
         name switch
         {
-            nameof(ChangeNamespace) => typeof(ChangeNamespace),
+            nameof(AddApiProfiles) => typeof(AddApiProfiles),
```
Exanite, line 19: I alphabetically sorted the names here. This should be easier to read through and cause less diff noise in the future. Even if we forget to sort it when adding new names, sorting it again won't cause as much noise as the first sort.

sources/SilkTouch/SilkTouch/Mods/Common/ModUtils.cs
```diff
@@ -101,7 +107,7 @@ string candidate
     /// <param name="toks">Tokens e.g. ref, in, out.</param>
     /// <param name="type">The type syntax.</param>
     /// <returns>The discriminator string.</returns>
-    public static string DiscrimStr(SyntaxTokenList? toks, TypeSyntax? type) =>
+    public static string GetMethodDiscriminator(SyntaxTokenList? toks, TypeSyntax? type) =>
```
Exanite, line 110: I renamed this because I always had to read the name twice to figure out what it meant. This should help with skimming the code.

sources/SilkTouch/SilkTouch/Mods/ExtractHandles.cs
```diff
@@ -0,0 +1,324 @@
...
+namespace Silk.NET.SilkTouch.Mods;
+
+/// <summary>
+/// Adds empty handle structs by searching for missing types referenced through pointers.
+/// If all references to the missing type are through a pointer,
+/// that missing type is then added as an empty struct.
+/// <para/>
+/// See <see cref="TransformHandles"/> for applying further transformations.
+/// </summary>
+public class ExtractHandles(ILogger<ExtractHandles> logger) : Mod
```
Exanite, line 20: This mod is effectively a cut and paste from TransformHandles.
This is part of my move to separate extraction/type creation from type transformation.

Originally, this was because I thought this was needed for better nested struct handling, but it turned out unnecessary. I kept it because it makes the codebase cleaner.

Exanite: Note I removed the `AssumeMissingTypesOpaque` config option when moving the code to this mod. This is because the false value for the config option is equivalent to disabling the entire mod.

sources/SilkTouch/SilkTouch/Mods/ExtractNestedTyping.cs
```diff
@@ -322,11 +322,20 @@ mem is not StructDeclarationSyntax struc
                     continue;
                 }
 
-                var iden = $"{node.Identifier}_{match.Groups[1].Value}";
+                var iden = $"{node.Identifier}{match.Groups[1].Value}";
```
Exanite, line 325: The underscore here is no longer necessary because the affix metadata lets us distinguish the two parts during prettification.

sources/SilkTouch/SilkTouch/Mods/IdentifySharedPrefixes.cs
```diff
@@ -0,0 +1,726 @@
...
+/// <summary>
+/// Identifies shared prefixes,
+/// such as namespace prefixes (eg: vk, gl)
+/// and enum constant prefixes (eg: VK_ACCESS).
+/// <para/>
+/// These are identified as [NameAffix] attributes that
+/// <see cref="PrettifyNames"/> and other mods can then process further.
+/// </summary>
+[ModConfiguration<Configuration>]
+public class IdentifySharedPrefixes(IOptionsSnapshot<IdentifySharedPrefixes.Configuration> config)
```
Exanite, line 21: This mod replaces NameTrimmer and part of the logic in PrettifyNames.
The prefix identification behavior remains the same, but prefixes are annotated as `[NameAffix]` attributes. PrettifyNames can then be configured to remove the prefix separately.

Notably, the core prefix identification code is a copy-paste from the NameTrimmer class and adapted to just identify the prefix (instead of identifying + trimming).

NameTrimmer has been completely removed.

sources/SilkTouch/SilkTouch/Mods/IdentifySharedPrefixes.cs
```diff
@@ -0,0 +1,726 @@
...
+    : Mod
+{
+    /// <summary>
+    /// This was from the original NameTrimmer code
+    /// Quoting from the original documentation:
+    /// The second pass does not use <see cref="GetTrimmingName"/>.
+    /// The third pass uses naive prefix detection for <see cref="GetPrefix"/>.
+    /// </summary>
+    /// <remarks>
+    /// This documentation should be expanded to explain the reasoning for each pass.
```
Exanite, line 31: The remark comment here is because I genuinely don't know what each of the passes do and don't want to spend too much time figuring it out.

sources/SilkTouch/SilkTouch/Mods/IdentifySharedPrefixes.cs
```diff
@@ -0,0 +1,726 @@
...
+
+    /// <summary>
+    /// Strings that are not allowed to be part of the final prefix.
+    /// </summary>
+    private static readonly HashSet<string> _forbiddenPrefixes = new() { "unsigned", "per" };
+
+    /// <summary>
+    /// The configuration for the <see cref="IdentifySharedPrefixes"/> mod.
+    /// </summary>
+    public record Configuration
```
Exanite, line 43: Options here were originally from PrettifyNames and behave the exact same.

sources/SilkTouch/SilkTouch/Naming/NameUtils.cs
```diff
@@ -38,6 +37,19 @@ public static partial class NameUtils
         "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789_"
     );
 
+    /// <summary>
+    /// Prefixes the given identifier with the specified prefix if it starts with a number.
+    /// </summary>
+    public static string PrefixIfStartsWithNumber(string identifier, string prefix = "X")
```
Exanite, line 43: This behavior was separated out from NamePrettifier in favor of doing prettification and prefixing in separate steps. This is because PrettifyNames does additional processing after a name is prettified and it's much more flexible/correct to handle this step separately.

sources/SilkTouch/SilkTouch/Naming/NameUtils.cs
```diff
@@ -174,47 +186,6 @@ public static string FindCommonPrefix(
         return foundPrefix[..(naive ? foundPrefix.Length : foundPrefix.LastIndexOf('_') + 1)];
     }
 
-    /// <summary>
-    /// Separates the input words with underscore
-    /// </summary>
-    /// <param name="input">The string to be underscored</param>
-    /// <returns></returns>
-    public static string LenientUnderscore(this string input) =>
```
Exanite, line 182: Removed in favor of `NameSplitter.Underscore`. NameSplitter should be easier to understand and maintain (it's a for loop + switch statement instead of nested regex and LINQ). It additionally ensures that name splitting/tokenization is done consistently throughout the codebase. Additionally, NameSplitter has extensive test coverage that documents the expected behavior.

Note that the behavior is slightly different. I explain the reasoning for not keeping the same behavior here: https://discord.com/channels/521092042781229087/1485943448069738608/1489457083857506336

sources/SilkTouch/SilkTouch/ServiceCollectionExtensions.cs
```diff
@@ -96,16 +94,13 @@ IConfiguration config
         // - https://andrewlock.net/how-to-register-a-service-with-multiple-interfaces-for-in-asp-net-core-di/
         // - https://discordapp.com/channels/143867839282020352/663803973119115264/1129546023388332075 (C# Discord)
 
-        // services.AddSingleton<ClangScraper>(); <-- this is a mod now
```
Exanite, line 99: General cleanup, I don't see a case where we want this line back.

sources/SilkTouch/SilkTouch/ServiceCollectionExtensions.cs
```diff
@@ -96,16 +94,13 @@ IConfiguration config
...
         services.AddSingleton<ResponseFileHandler>();
         services.AddSingleton<FunctionTransformer>();
         services.AddSingleton<MSBuildModContextProvider>();
         services.TryAddSingleton<IModContextProvider>(s =>
             s.GetRequiredService<MSBuildModContextProvider>()
         );
-        services.AddSingleton<NameTrimmer>();
         services.AddSingleton<DefaultBakeStrategy>();
-        services.AddSingleton<INameTrimmer>(s => s.GetRequiredService<NameTrimmer>());
```
Exanite, line 108: NameTrimmer has been removed.

`INameTrimmer` has been replaced by `INameProcessor`, but is now private to PrettifyNames. `INameTrimmer` was previously used to modify/extend the name processing behavior of PrettifyNames, but is now unnecessary with the affix system (since mods have enough metadata to handle transforming the affixes themselves; users can also configure the behavior through the generator.json config file).

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
```diff
@@ -1915,9 +1915,8 @@ private SyntaxList<AttributeListSyntax> ProcessAndGetNewAttributes(
             MethodDeclarationSyntax? methodDeclaration = null
         )
         {
-            // Get the name of the identifier, preferring the native one if available
             // This name will be modified by the code below as different suffixes are identified
-            var trimmedName = attributeLists.GetNativeNameOrDefault(identifier);
+            var trimmedName = identifier.ToString();
```
Exanite, line 1919: This is now more correct now that the referenced affix system exists.

This was previously used to handle the `CallbackEXTDelegate` case where the vendor suffix is in the middle of the C# name, but at the end of the native name (since the native name would be `CallbackEXT`).

The referenced affix system bypasses the need to identify the `EXT` extension at all by just using the final name of the `CallbackEXT` function pointer struct.

sources/SilkTouch/SilkTouch/Naming/CandidateNames.cs
```diff
@@ -1,18 +0,0 @@
...
-// The .NET Foundation licenses this file to you under the MIT license.
-
-namespace Silk.NET.SilkTouch.Naming;
-
-/// <summary>
-/// Represents the set of primary and secondary options for the trimmed version of a name.
-/// </summary>
-/// <param name="Primary">The preferred version of the trimmed name.</param>
-/// <param name="Secondary">The fallback versions of the trimmed name in case the primary does not work.</param>
-public record struct CandidateNames(string Primary, List<string> Secondary)
```
Exanite, line 11: Moved into PrettifyNames as a private type.

sources/SilkTouch/SilkTouch/Naming/INameTrimmer.cs
```diff
@@ -1,29 +0,0 @@
-﻿using System;
-using System.Collections.Generic;
-
-namespace Silk.NET.SilkTouch.Naming;
-
-/// <summary>
-/// Represents a name trimmer.
-/// </summary>
-public interface INameTrimmer
```
Exanite, line 9: Moved into PrettifyNames as a private type (`INameProcessor`).

Note that I also removed Version and Order.
These no longer make sense to use in the new implementation (name affixes are extensively configurable by users and transformable by mods).

Version was used to conditionally enable each trimmer.
This shouldn't be necessary anymore and we can use normal config options to configure/enable different processors if necessary.

sources/SilkTouch/SilkTouch/Mods/TransformHandles.cs
```diff
@@ -25,11 +25,6 @@ namespace Silk.NET.SilkTouch.Mods;
 /// <c>VkBuffer*</c>, usages of that pointer will be replaced by <c>VkBufferHandle</c>. For a 2-dimensional pointer,
 /// <c>VkBuffer**</c>, the resulting replacement is <c>VkBufferHandle*</c>.
 /// </example>
-/// <remarks>
-/// It is assumed that all handle types in the generated syntax do not require a <c>using</c> directive in order to be
-/// in scope. That is, if a file with the namespace <c>Silk.NET.OpenGL</c> is encountered and it is referencing
-/// <c>Program</c>, <c>Program</c> must be declared in <c>Silk.NET.OpenGL</c>, <c>Silk.NET</c>, or <c>Silk</c>.
-/// </remarks>
```
Exanite, line 32: I don't think this comment is necessary since it's an assumption made by the entire codebase, not just this mod.

sources/SilkTouch/SilkTouch/Mods/TransformHandles.cs
```diff
@@ -41,23 +36,18 @@ ILogger<TransformHandles> logger
...
-        /// <summary>
-        /// Whether it should be assumed that missing types are likely opaque if they are only used as a pointer type
-        /// and therefore should be subjected to handle transformations.
-        /// </summary>
-        public bool AssumeMissingTypesOpaque { get; init; }
-
         /// <summary>
         /// Whether the DSL (i.e. <c>nullptr</c>) should be usable with handle types.
         /// </summary>
-        public bool UseDSL { get; init; }
```
Exanite, line 53: Changed this to `UseDsl` to be consistent with the Microsoft naming convention guidelines, which is used for the rest of the codebase.

sources/SilkTouch/SilkTouch/Mods/TransformHandles.cs
```diff
@@ -70,45 +60,13 @@ public override async Task ExecuteAsync(IModContext ctx, CancellationToken ct =
             throw new InvalidOperationException("Failed to get compilation");
         }
 
-        var cfg = config.Get(ctx.JobKey);
```
Exanite, line 73: Moved to ExtractHandles.

sources/SilkTouch/SilkTouch/Naming/NameTrimmer.cs
```diff
@@ -1,407 +0,0 @@
...
-using System.Diagnostics;
-using System.Linq;
-using System.Text.Json;
-
-namespace Silk.NET.SilkTouch.Naming;
-
-/// <summary>
-/// The default name trimmer ported from Silk.NET 2.18 with modifications for 3.0.
-/// </summary>
-public class NameTrimmer : INameTrimmer
```
Exanite, line 12: Replaced by IdentifySharedPrefixes.

sources/SilkTouch/SilkTouch/Naming/NameSplitter.cs
```diff
@@ -10,14 +10,20 @@ namespace Silk.NET.SilkTouch.Naming;
...
+
     /// <summary>
     /// Splits the given C# identifier into separate words.
     /// </summary>
     /// <remarks>
     /// See the test cases for this method to see examples on how this method behaves.
     /// </remarks>
     /// <param name="identifier">A string that contains only valid C# identifier characters.</param>
-    public static List<string> BreakIntoWords(string identifier)
+    public static List<string> SplitIntoWords(string identifier)
```
Exanite, line 26: Changed for consistency.

sources/SilkTouch/SilkTouch/Naming/NameSplitter.cs
```diff
@@ -10,14 +10,20 @@ namespace Silk.NET.SilkTouch.Naming;
 /// </summary>
 public static class NameSplitter
 {
+    /// <summary>
+    /// Splits the given C# identifier into separate words and rejoins them using an underscore as a separator.
+    /// </summary>
+    public static string Underscore(string identifier) =>
```
Exanite, line 16: This replaces NameUtils.LenientUnderscore. I explain more in a comment there.

sources/SilkTouch/SilkTouch/Naming/NamePrettifier.cs
```diff
@@ -31,12 +32,7 @@ public class NamePrettifier(int longAcronymThreshold)
     /// <exception cref="InvalidOperationException">Thrown when the input or output is an empty identifier.</exception>
     public string Prettify(string identifier, bool allowAllCaps = false)
     {
-        if (identifier.Length == 0)
```
Exanite, line 34: Prettify is now allows empty strings as input/output.
This is because Prettify is now designed to handle name fragments instead of full names. This notably allows PrettifyNames to be more granular/correct in how it processes names.

This also goes for the "prefix numbers with X" behavior below.

sources/SilkTouch/SilkTouch/Naming/NameAffixer.cs
```diff
@@ -32,13 +32,21 @@ public static NameAffix[] GetNameAffixes(this SyntaxList<AttributeListSyntax> at
 
                 var argumentList = attribute.ArgumentList;
                 if (
-                    argumentList != null
```
Exanite, line 35: The changes in this file are related to the addition of the referenced affix system.

sources/Vulkan/Vulkan/Vulkan/DebugMarkerMarkerInfoEXT.gen.cs
```diff
@@ -26,5 +26,5 @@ public unsafe partial struct DebugMarkerMarkerInfoEXT
 
     [NativeName("color")]
     [SupportedApiProfile("vulkan", ["VK_EXT_debug_marker"], ImpliesSets = ["VK_EXT_debug_report"])]
-    public DebugMarkerMarkerInfoExtColor Color;
```
Exanite, line 29: This uses the referenced affix system to ensure that the vendor suffix in the middle of the name is prettified correctly.

sources/SilkTouch/SilkTouch/Mods/PrettifyNames.cs
```diff
@@ -341,53 +176,16 @@ public async Task ExecuteAsync(IModContext ctx, CancellationToken ct = default)
...
-                        ]
-                    );
-            }),
-            logger,
-            ct
-        );
+        // Gather symbols and rename
+        var symbolVisitor = new NameSymbolVisitor(newNames);
+        symbolVisitor.Visit(compilation.Assembly);
+        await NameUtils.RenameAllAsync(ctx, symbolVisitor.ToRename, logger, ct);
```
Exanite, line 188: I replaced the LINQ code on the left with another visitor since I find this much easier to debug and understand.

The `-Value` suffixing on the left has now been moved to be part of `ResolveConflictsProcessor` closer to the end of the file. The `-Value` suffix was used when a non-method conflicted with a method with the same output name. More info here: https://discord.com/channels/521092042781229087/1485943448069738608/1491053058216558592

sources/SilkTouch/SilkTouch/Mods/PrettifyNames.cs
```diff
@@ -106,212 +100,53 @@ public record NameAffixConfiguration
...
-            }
-        }
+            var namePrettifier = new NamePrettifier(cfg.LongAcronymThreshold);
 
-        // Prettify the prettify only names
-        foreach (var (typeName, memberNames) in visitor.PrettifyOnlyTypes)
-        {
-            if (!newNames.TryGetValue(typeName, out var renamedType))
+            // Define name processors
+            var nameProcessors = new INameProcessor[]
```
Exanite, line 122: Compared to before, name processors are now statically defined here. No DI or sorting. This is less flexible, but we currently don't need that flexibility. Plus, it's easy enough to add back if necessary.

Also, all names now flow sequentially through the same pipeline (there were 3 separate pipelines before).

Previously, names were separated by "container", which is not to be confused with the concept of "scope". A container was a grouping of certain categories of names in the same scope (eg: all methods or all constants for a type).

Now, all processors see the state of every name in the project at the same time. This allows for better optimization (bulk processing), correctness (conflict resolution sees all names), and flexibility (again, can see all names). If the old grouping is ever necessary for anything, we can group in the relevant processor instead of as a global process.

sources/SilkTouch/SilkTouch/Mods/PrettifyNames.cs
```diff
@@ -397,16 +195,21 @@ z.MethodKind is MethodKind.Constructor or MethodKind.Destructor
...
-            )
+            if (doc?.FilePath == null)
+            {
+                continue;
+            }
+
+            var firstMatch = typeNamesLongestFirst.FirstOrDefault(x =>
+                doc.FilePath.Contains(x.Key) || doc.Name.Contains(x.Key)
+            );
+            if (firstMatch is not { Key: { } oldName, Value: { } newName })
```
Exanite, line 212: Slight performance/readability optimization here.

sources/SilkTouch/SilkTouch/Mods/PrettifyNames.cs
```diff
@@ -453,856 +256,551 @@ z.MethodKind is MethodKind.Constructor or MethodKind.Destructor
         ctx.SourceProject = proj;
     }
 
-    /// <summary>
-    /// Applies the prettify only pipeline.
-    /// This currently consists of checking for name overrides first.
-    /// Then if no override is found, then the name's affixes are removed,
-    /// the name is prettified, and the name's affixes are added back.
-    /// </summary>
-    private string ApplyPrettifyOnlyPipeline(
```
Exanite, line 462: ApplyPrettifyOnlyPipeline is completely removed since the functionality is now covered by a separate mod.

sources/SilkTouch/SilkTouch/Mods/PrettifyNames.cs
```diff
@@ -453,856 +256,551 @@ z.MethodKind is MethodKind.Constructor or MethodKind.Destructor
...
-        var result = name;
-        result = nameAffixer.RemoveAffixes(result, container, name, null);
-        result = namePrettifier.Prettify(result, allowAllCaps);
-        result = nameAffixer.ApplyAffixes(result, container, name, null);
-
-        return result;
+        return Task.FromResult(rsps);
     }
 
-    private void Trim(
```
Exanite, line 515: Trim is removed, with the relevant functionality being moved to individual INameProcessors at the end of the file.

sources/SilkTouch/SilkTouch/Mods/PrettifyNames.cs
```diff
@@ -453,856 +256,551 @@ z.MethodKind is MethodKind.Constructor or MethodKind.Destructor
...
+    /// Stores additional data for each scope member.
+    /// </summary>
+    /// <param name="Affixes">The affixes declared for the name.</param>
+    /// <param name="Declarations">The declaration syntaxes for the member.</param>
+    private record struct MemberData(
+        NameAffix[] Affixes,
+        List<MemberDeclarationSyntax> Declarations
+    );
+
+    private class NameDataVisitor : CSharpSyntaxWalker
```
Exanite, line 297: This was rewritten because we no longer need to separate trim+prettify vs prettify-only names. During this rewrite, I rewrote the entire visitor to be in the same style as the visitor I wrote for IdentifySharedPrefixes. This should be easier to understand since the core logic is concentrated in one method.

sources/SilkTouch/SilkTouch/Mods/PrettifyNames.cs
```diff
@@ -453,856 +256,551 @@ z.MethodKind is MethodKind.Constructor or MethodKind.Destructor
...
+
+        public override void VisitMethodDeclaration(MethodDeclarationSyntax node) =>
+            ReportName(node.Identifier, node.AttributeLists, node);
+    }
+
+    /// <summary>
+    /// Discovers and stores symbol to new name mappings to be used by <see cref="NameUtils.RenameAllAsync"/>.
+    /// </summary>
+    /// <param name="newNames">The new names in the format defined by <see cref="NameProcessorContext.FinalNames"/>.</param>
+    private class NameSymbolVisitor(Dictionary<string, Dictionary<string, string>> newNames)
```
Exanite, line 399: This is used in `ExecuteAsync` to map names to symbols before the renaming process happens. This is because the renamer requires symbols instead of raw names.

sources/SilkTouch/SilkTouch/Mods/PrettifyNames.cs
```diff
@@ -453,856 +256,551 @@ z.MethodKind is MethodKind.Constructor or MethodKind.Destructor
...
+            (
+                (MethodDeclarationSyntax)base.VisitMethodDeclaration(node)!
+            ).WithRenameSafeAttributeLists();
+    }
+
+    /// <summary>
+    /// Applies name overrides and moves overridden names to the final set of names.
+    /// Overridden names are removed from the working set to prevent later processors from directly seeing them.
+    /// </summary>
+    private class HandleOverridesProcessor(Dictionary<string, string> nameOverrides)
```
Exanite, line 482: This is a rewritten version of the original override handling code in `Trim()`.
No behavioral changes, however.

sources/SilkTouch/SilkTouch/Mods/PrettifyNames.cs
```diff
@@ -453,856 +256,551 @@ z.MethodKind is MethodKind.Constructor or MethodKind.Destructor
...
         }
     }
 
-    /// <inheritdoc />
-    public Task<List<ResponseFile>> BeforeScrapeAsync(string key, List<ResponseFile> rsps)
+    /// <summary>
+    /// Removes identified affixes so that other name processors can process the base name separately.
+    /// These affixes should be reapplied by <see cref="ReapplyAffixesProcessor"/>.
+    /// </summary>
+    private class StripAffixesProcessor(NameDataVisitor nameData) : INameProcessor
```
Exanite, line 556: This isn't new, but was modified a decent amount. Namely, `PrettifyNamesAffixer` is gone and StripAffixes has been moved to this class as a private method.

sources/SilkTouch/SilkTouch/Mods/PrettifyNames.cs
```diff
@@ -453,856 +256,551 @@ z.MethodKind is MethodKind.Constructor or MethodKind.Destructor
...
+    }
 
-        // ----- Types -----
-
-        public override void VisitClassDeclaration(ClassDeclarationSyntax node)
+    /// <summary>
+    /// Prettifies the primary and secondary candidate names.
+    /// Also see <see cref="NamePrettifier.Prettify"/>.
+    /// </summary>
+    private class PrettifyProcessor(NamePrettifier namePrettifier) : INameProcessor
```
Exanite, line 605: Mostly unchanged.

sources/SilkTouch/SilkTouch/Mods/PrettifyNames.cs
```diff
@@ -453,856 +256,551 @@ z.MethodKind is MethodKind.Constructor or MethodKind.Destructor
...
-            // Merge with existing data in case of partials
-            if (!TrimmableTypes.TryGetValue(identifier, out var typeData))
-            {
-                typeData = new TypeData([], []);
-                TrimmableTypes.Add(identifier, typeData);
-            }
+    /// <summary>
+    /// Reapplies and transforms identified affixes based on <see cref="NameAffixConfiguration"/>.
+    /// </summary>
+    private class ReapplyAffixesProcessor(
```
Exanite, line 633: Heavily modified.

This now contains a topological sort in order to handle referenced affixes in the right order (dependencies must be processed first).

Similarly to `StripAffixesProcessor`, the related `ApplyAffixes` method has been moved here as a private method.

`ApplyAffixes` has been updated to handle referenced affixes and the prettification of individual affixes.

sources/SilkTouch/SilkTouch/Mods/PrettifyNames.cs
```diff
@@ -1448,109 +1078,604 @@ private NameAffixConfiguration GetConfiguration(string category) =>
     }
 
     /// <summary>
-    /// Removes identified affixes so that other trimmers can process the base name separately.
-    /// These affixes should be reapplied by <see cref="NameAffixerLateTrimmer"/>.
+    /// Prefixes candidate names that start with a number.
+    /// This is required because C# identifiers cannot start with a number.
     /// </summary>
-    private class NameAffixerEarlyTrimmer(PrettifyNamesAffixer affixer) : INameTrimmer
+    private class PrefixIfStartsWithNumberProcessor : INameProcessor
```
Exanite, line 1084: This is new and was previously implicitly handled by `PrettifyProcessor` since `NamePrettifier.PrettifyNames` used to assume that input/output names represented full identifiers.

This was necessary because `PrettifyProcessor` only sees part of the name and prefixing identifiers requires seeing the full name (or at least the very start of it). This is not the case until after `ReapplyAffixesProcessor`.

For reference:
```cs
var nameProcessors = new INameProcessor[]
{
    new HandleOverridesProcessor(cfg.NameOverrides),
    new StripAffixesProcessor(visitor),
    new PrettifyProcessor(namePrettifier),
    new ReapplyAffixesProcessor(visitor, namePrettifier, cfg.Affixes),
    new PrefixIfStartsWithNumberProcessor(),
    new ResolveConflictsProcessor(visitor, logger),
    new OutputFinalNamesProcessor(),
    new RemoveUnmodifiedFinalNamesProcessor(),
};
```

sources/SilkTouch/SilkTouch/Mods/PrettifyNames.cs
```diff
@@ -1448,109 +1078,604 @@ private NameAffixConfiguration GetConfiguration(string category) =>
...
-        public void Trim(NameTrimmerContext context)
+    /// <summary>
+    /// Resolves conflicts where multiple names have been transformed into the same output name.
+    /// This also considers cases where the methods with the same name are compatible because of method overloading rules.
+    /// </summary>
+    /// <remarks>
+    /// This also copies the names from the final set back into the working set.
+    /// This is to ensure that all names are considered by the conflict resolution algorithm.
+    /// </remarks>
+    private class ResolveConflictsProcessor(NameDataVisitor nameData, ILogger logger)
```
Exanite, line 1114: The core logic here is verbatim copied from the now removed `Trim()` method.
However, I renamed some identifiers, edited comments, and rewrote some of the code for readability purposes as I was figuring out how it worked.

The changes I made are related to non-member/member conflict resolution and secondary selection priority.

Non-member/member conflict resolution was previously handled adhoc right before applying the renames, which I assume was because of how "containers" were exposed to the old `INameTrimmer` implementations. `INameTrimmer`s could not see all names in a type and thus couldn't resolve these conflicts properly.

Secondary selection priority now prioritizes the last secondary candidate names available in the secondary list. Previously, it was prioritizing the shortest name, which could lead to unwanted results (such as ignoring the affix discriminator priority config).

sources/SilkTouch/SilkTouch/Mods/PrettifyNames.cs
```diff
@@ -1448,109 +1078,604 @@ private NameAffixConfiguration GetConfiguration(string category) =>
...
     /// <summary>
-    /// Reapplies and transforms identified affixes based on <see cref="NameAffixConfiguration"/>.
+    /// Outputs all primary names to the final set of names.
     /// </summary>
-    private class NameAffixerLateTrimmer(PrettifyNamesAffixer affixer) : INameTrimmer
+    /// <remarks>
+    /// This is intentionally implemented in a naive manner.
+    /// The working set of names is not cleared and existing data in the final set can be overwritten.
+    /// </remarks>
+    private class OutputFinalNamesProcessor : INameProcessor
```
Exanite, line 1556: This is new and just copies the names from the working set of names to the final set.

This distinction between working/final sets is made because of name overrides.
Most processors should not be able to see overridden names, but some do, such as `ResolveConflictsProcessor`.

sources/SilkTouch/SilkTouch/Mods/PrettifyNames.cs
```diff
@@ -1448,109 +1078,604 @@ private NameAffixConfiguration GetConfiguration(string category) =>
...
+        /// </summary>
+        public void ProcessNames(NameProcessorContext context);
+    }
 
-        /// <inheritdoc/>
-        public int Order => (int)TrimmerOrder.PrettifyNamesTrimmer;
+    /// <summary>
+    /// State made available to <see cref="INameProcessor"/> implementations.
+    /// </summary>
+    private class NameProcessorContext
```
Exanite, line 1632: This is similar to the original `NameTrimmerContext` struct, but simplified to just hold the working and final set of names. Previously, `NameTrimmerContext` also contained config information, but is now unnecessary since we can just pass config and context information to the processors when creating them.

sources/SilkTouch/SilkTouch/Mods/PrettifyNames.cs
```diff
@@ -1448,109 +1078,604 @@ private NameAffixConfiguration GetConfiguration(string category) =>
...
-            }
-        }
+    /// <summary>
+    /// Represents the set of primary and secondary candidates for the prettified version of a name.
+    /// </summary>
+    /// <param name="Primary">The preferred version of the output name.</param>
+    /// <param name="Secondary">
+    /// The fallback versions of the output name.
+    /// Used in the case the primary causes conflicts.
+    /// Names later in the list have higher priority.
```
Exanite, line 1674: This is the same as `CandidateNames` previously. However, I added a note that specifies names later in the Secondary list have priority. Previously, the shortest name was assumed to have highest priority by the name conflict resolution code, which can lead to bad behavior.

tests/SilkTouch/SilkTouch/Naming/NameTrimmerTests.cs
```diff
@@ -1,159 +0,0 @@
-// Licensed to the .NET Foundation under one or more agreements.
-// The .NET Foundation licenses this file to you under the MIT license.
-
-using Silk.NET.SilkTouch.Mods;
-using Silk.NET.SilkTouch.Naming;
-
-namespace Silk.NET.SilkTouch.UnitTests.Naming;
-
-public class NameTrimmerTests : NameTrimmer
```
Exanite, line 9: Test cases in this class were ported over to the new IdentifySharedPrefixTests class.

tests/SilkTouch/SilkTouch/Naming/NamePrettifierTests.cs
```diff
@@ -71,17 +71,21 @@ public string AcronymsWithNumbers2(string input, int longAcronymThreshold = 0) =
     public string ConsecutiveAcronyms(string input, int longAcronymThreshold = 0) =>
         new NamePrettifier(longAcronymThreshold).Prettify(input);
 
+    /// <summary>
+    /// C# identifiers cannot start with numbers, but <see cref="NamePrettifier.Prettify"/> no longer handles this
+    /// in favor of calling <see cref="NameUtils.PrefixIfStartsWithNumber"/> separately.
+    /// </summary>
     [Theory]
-    // C# identifiers cannot start with numbers
```
Exanite, line 75: The tests in this file were updated to match the change I made to Prettify (which now processes name fragments instead of full names).

generator.json
```diff
@@ -59,17 +61,23 @@
...
         }
       },
-      "PrettifyNames": {
+      "IdentifySharedPrefixes": {
         "GlobalPrefixHints": ["SDL"]
       },
+      "PrettifyNames": {
+        "Affixes": {
+          "SharedPrefix": {
+            "Remove": true
```
Exanite, line 77: For SharedPrefix, we can either configure it as Remove or as a Discriminator. Both have the same effect until names conflict *and* no other discriminators are available. Fortunately, we have no names conflict in this way right now and I assume this case will be very rare, so I'm defaulting to Remove.

generator.json
```diff
@@ -171,14 +180,17 @@
...
+      "IdentifySharedPrefixes": {
         "GlobalPrefixHints": ["gl"],
         "PrefixOverrides": {
           "SyncObjectMask": "GL_SYNC",
           "OcclusionQueryParameterNameNV": "GL",
           "TexStorageAttribs": "GL",
-          "ContainerType": "GL"
-        },
+          "ContainerType": "GL",
+          "SeparableTargetEXT": "GL"
```
Exanite, line 190: This is new override is required since my new name underscorer/splitter handles 2D differently. It outputs `2_D` instead of `2D` since `2_D` is arguably more consistent. I explain why here: https://discord.com/channels/521092042781229087/1485943448069738608/1489457083857506336

It's easier to handle the common case consistently while handling edge cases using overrides.

generator.json
```diff
@@ -294,25 +311,20 @@
...
           "ALBUFFERCALLBACKTYPESOFT": "BufferCallbackSOFT",
-          "ALBUFFERCALLBACKTYPESOFTDelegate": "BufferCallbackDelegateSOFT",
           "ALCEVENTPROCTYPESOFT": "ContextEventProcSOFT",
-          "ALCEVENTPROCTYPESOFTDelegate": "ContextEventProcDelegateSOFT",
           "ALEVENTPROCSOFT": "EventProcSOFT",
-          "ALEVENTPROCSOFTDelegate": "EventProcDelegateSOFT",
           "ALDEBUGPROCEXT": "DebugProcEXT",
-          "ALDEBUGPROCEXTDelegate": "DebugProcDelegateEXT",
-          "LPALFOLDBACKCALLBACK": "FoldbackCallback",
-          "LPALFOLDBACKCALLBACKDelegate": "FoldbackCallbackDelegate"
```
Exanite, line 315: Most of these overrides are no longer necessary because of referenced affixes. Overrides to the referenced name automatically propagate to names that reference them.

Exanite (commented):
Self code review complete. The review comments I added provide additional context on the changes I made, but the code itself should make sense without them.

curin (approved):

PR #2575, opened 2026-05-01
Fix some tutorial bits

aquagoose:
# Summary of the PR
- Fixes the Texture tutorial so it uses `glTexParameter` instead of `glTextureParameter`
- Replaces uses of StbImageSharp with ImageSharp
  - This was done because all the other tutorials use ImageSharp, and at the time when I wrote the texture tutorial I decided to replace ImageSharp with StbImageSharp for some reason, but only in the texture tutorial and the *existing* abstraction tutorial which makes no sense. Also there was a lot of pushback after it was merged after the team noticed it had been replaced, so I'm righting the wrong here.

[Links go here.](https://discord.com/channels/521092042781229087/607634593201520651/1499733224648343663)

# Further Comments


examples/CSharp/OpenGL Tutorials/Tutorial 1.4 - Abstractions/Texture.cs
```diff
@@ -19,14 +20,24 @@ public unsafe Texture(GL gl, string path)
             _handle = _gl.GenTexture();
             Bind();
             
-            // Load the image from memory.
-            ImageResult result = ImageResult.FromMemory(File.ReadAllBytes(path), ColorComponents.RedGreenBlueAlpha);
-            
-            fixed (byte* ptr = result.Data)
+            //Loading an image using imagesharp.
```
Exanite: Nit: Ideally we have consistent comment formatting and capitalization here and in the rest of the file since this is part of tutorial code.

Exanite (approved):
Looks good to me. We can merge once we get a review from Beyley.

I ran the examples locally on Linux and they work fine.
Behavior also matches the `main` branch.

Perksey (approved):
I'm happy to bless this on behalf of Beyley

examples/CSharp/OpenGL Tutorials/Tutorial 1.4 - Abstractions/Texture.cs
Exanite: Didn't realize that this formatting style was consistent with the rest of the file, but I added a commit to clean up the comments in the file a bit.

Exanite:
@Perksey Feel free to merge. I don't want to merge this myself since I added an extra commit above.

Perksey (approved):

PR #2576, opened 2026-05-01
[3.0] Add Khronos StructureType field initializers

Exanite:
# Summary of the PR

This adds default field initializers to structures that have structure type members, such as `sType` for `VkStructureType` in `Vulkan`.

Original discussion with Perksey: https://discord.com/channels/521092042781229087/1376331581198827520/1395826788101263463

Discord thread for work done during this PR: https://discord.com/channels/521092042781229087/1499891853217103902

# Further Comments

## Tasks

- [x] Scrape structure type member information
- [x] Add field initializers to structure type members
   - [x] Add explicit constructor since field initializers require them.
- [x] Update generator.json to use the new config option
- [x] Wait for previous PRs to be merged
   - [x] Rebase
- [x] Regenerate Vulkan bindings using Windows


Exanite:
Verified Windows generation output is the exact same, so not adding an extra commit.

eng/silktouch/vulkan/settings.rsp
```diff
@@ -1,5 +1,6 @@
 @../../common.rsp
 --define-macro
+VK_ENABLE_BETA_EXTENSIONS
```
Exanite, line 3: TerraFX also does the same: https://github.com/terrafx/terrafx.interop.vulkan/blob/86629a4d0afc1de12d10111b4d1e8bbd4be4363b/generation/settings.rsp#L36

This macro mainly guards a few enum members. Without these enum members, the generated bindings fail to compile.

.silktouch/vulkan-clangsharp.stout
Exanite, line 1: I did a search over the repo for any missing structure type initializers, and looks like we are good:
Image

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
```diff
@@ -402,6 +465,18 @@ public async Task ExecuteAsync(IModContext ctx, CancellationToken ct = default)
             ).Project;
         }
 
+        // Rewrite phase 4
```
Exanite, line 468: I've decided that it's better to use separate rewriters for everything in MixKhronosData.

There is a performance impact of doing this, but it's much more maintainable to keep every transformation independent of the others. The performance impact is roughly 2 seconds for Vulkan so it is very much worth paying. It also does not affect non-Khronos bindings.

I also plan to keep any new metadata storage structures separate from existing structures. This is to further keep things independent. Every transformation should scrape their own data (within reason).

In particular, what I have in mind relates to `StructureTypeMembers` within this PR and any metadata we need for the Vulkan struct chaining feature later on. I plan to keep metadata used for Vulkan struct chaining separate from the structure type field initialization code added in this PR.

generator.json
```diff
@@ -428,6 +428,9 @@
       "MixKhronosData": {
         "SpecPath": "eng/submodules/vulkan/xml/vk.xml",
         "Namespace": "Silk.NET.Vulkan",
+        "StructureTypes": [
+          "VkStructureType"
```
Exanite, line 432: Similar to `FlagsTypes` immediately below, `StructureTypes` should let us handle OpenXR easily once we get to that point.

sources/Vulkan/Vulkan/Vulkan/AccelerationStructureBuildGeometryInfoKHR.gen.cs
```diff
@@ -132,4 +132,14 @@ public unsafe partial struct AccelerationStructureBuildGeometryInfoKHR
...
+
+    [SupportedApiProfile(
+        "vulkan",
+        ["VK_KHR_acceleration_structure"],
+        ImpliesSets = [
+            "VK_KHR_deferred_host_operations+VK_VERSION_1_1+VK_EXT_descriptor_indexing+VK_KHR_buffer_device_address",
+            "VK_KHR_deferred_host_operations+VK_VERSION_1_2",
+        ]
+    )]
+    public AccelerationStructureBuildGeometryInfoKHR() { }
```
Exanite, line 144: Structs that have field initializers need a constructor.

Exanite (commented):
Self code review done.

curin (approved):

PR #2579, opened 2026-05-12
docs: fix broken source-file links in managed chaining intro

mokashang:
# Summary of the PR

Fixes the three broken links in the opening paragraph of the Managed Chaining docs page (`documentation/vulkan/structure-chaining/managed-chaining.md`).

The links used repo-relative paths (`../../src/Vulkan/...`). On the rendered docs site they get rewritten to `/Silk.NET/docs/src/...`, which doesn't exist, so all three 404. Two of the targets were also renamed during the 3.0 work — `Chain.gen.tt` is now `Chain.g.tt`, and `Chain.gen.cs` is now `Chain.g.cs`.

Switched the three links to absolute GitHub URLs pointing at the current paths on `main`:

- `Chain.cs` — same file
- `Chain.gen.tt` → `Chain.g.tt`
- `Chain.gen.cs` → `Chain.g.cs`

All three new URLs return HTTP 200.

# Related issues, Discord discussions, or proposals

Fixes the three links initially called out by #2566

# Further Comments

Scope is intentionally limited to the three links called out in the issue (the opening paragraph). No other links in the file were changed. The unrelated CRLF normalization on `(WIP) Proposal - Axis Input Devices.md` that `git status` shows on a fresh clone is not included in this PR.

PR #2580, opened 2026-05-12
[2.x] Change Build workflow to always run

Exanite:
# Summary of the PR

The Build workflow is required to run and pass before a PR is allowed to be merged.
However, the workflow currently is excluded from running when the changes only modifies documentation files.
This blocks any PR that falls under the mentioned criteria from being merged.

This PR changes it so that the Build workflow always runs to fix that.
From what I can tell, there's no way to add a similar exclusion to the Ruleset enforcing this requirement.

The reason for this PR is due to https://github.com/dotnet/Silk.NET/pull/2579 since the Build workflow is not triggering for that PR and causing the merging of that PR to be blocked.

This is the original PR that added those exclusions: https://github.com/dotnet/Silk.NET/pull/691
- I'm guessing we didn't have the branch rules back then.

# Further Comments

I also recommitted the Axis Input Devices proposal file. This file seems to be causing issues due to its encoding.
Github shows this as a massive change that modifies every line in the file, but the only change I made was committing it again.

It also looks like CI takes a long time to run, but since this is for 2.x which is slowing down in development, I'm not too concerned.

Edit: Maybe I am concerned. 1 hour is a bit long. Since the pack step takes the longest and isn't critical for verifying code behavior, I'll open another PR that splits the workflow into two, with the pack step being optional when merging PRs.
Edit 2: I tested this in #2585, but the difference is only 15 minutes. While that seems significant, it still takes 30+ minutes and is long enough to where a contributor would need to step away for a while. It's not worth the cost of duplicating workflows and adding more complexity just to speed up documentation changes.


.github/workflows/build.yml
```diff
@@ -49,14 +44,14 @@ jobs:
         dotnet-version: 7.0.102
     - name: Install Workloads
       run: dotnet workload install android ios maccatalyst
+    - name: Validation Checks
+      if: ${{ github.repository != 'dotnet/Silk.NET' || !startsWith(github.ref, 'refs/tags/') }}
+      run: ./build.sh ValidateSolution
```
Exanite, line 49: This is reordered since it's better to fail on the faster check.

documentation/proposals/(WIP) Proposal - Axis Input Devices.md
```diff
@@ -1,794 +1,794 @@
-# Summary
```
Exanite, line 1: No actual changes. Just an encoding change since Git kept marking the file as changed when it wasn't. This is an old issue that I kept seeing and other people report having the same issue.

Perksey (approved):

PR #2585, opened 2026-05-17
[2.x] Split Build workflow into two

Exanite:
# Summary of the PR

This is an alternative of #2580, which explains the primary motivation for modifying the build system (in summary, branch protection rules need checks to run so we can't simply exclude them from being ran to save on time).

#2580 does this by just making the Build workflow always run, thus satisfying the branch protection requirement.
However, the workflow takes 1 hour to complete.

This workflow tries a different approach by splitting the Build workflow into two:
1. Check - This workflow is fast (to be confirmed) and always runs
2. Build - This workflow is almost the same as Check, but includes the extremely slow Pack steps

Related PR (also linked above): #2580

# Further Comments

Merging this PR will require the branch protection rules to require the `Check` workflow instead of the `Build` workflow. Currently the rules require `Build`.


Exanite:
So splitting the workflows saves about 15 minutes compared to not splitting them. Not fully convinced this is worth doing since it's still long enough that it is easier to leave running and come back later. We can automerge anyway so we can just hit automerge and walk away once a PR is acceptable for merging.

The main downside is iterating on contributions, but technically, we can see the results of the check while it is running so it's the difference between seeing a green checkmark vs clicking into the workflow to see what step it is on. This does make it is less contributor friendly, but I'd rather just have a working system (if slow) for Silk 2.x and focus on Silk 3.0 instead.

Issue #2547, opened 2026-02-21
Vulkan instance functions exposed by device extensions fail to be loaded


Exanite:
# Summary

Silk 2 assumes that all Vulkan functions exposed by device extensions can be retrieved by calling `vkGetDeviceProcAddr`, but apparently device level extensions can expose instance level functions that can only be retrieved by calling `vkGetInstanceProcAddr` (such as `vkGetPhysicalDeviceCalibrateableTimeDomainsKHR`).
 
The way that Vulkan functions are loaded in Silk 3 means that it will work (try device first, then instance), but it will also report a Vulkan validation warning:
```
vkGetDeviceProcAddr(): pName is trying to grab vkGetPhysicalDeviceCalibrateableTimeDomainsKHR which is an instance level function
```

Unfortunately, there doesn't seem to be an easy way to get around this either other than properly categorizing the functions according to the Vulkan spec for `vkGetDeviceProcAddr`:
> The returned function pointer must only be called with a dispatchable object (the first parameter) that is device or a child of device e.g. VkDevice, VkQueue, or VkCommandBuffer.

Silk 2 implementation:
https://github.com/dotnet/Silk.NET/blob/f91291d97f911fa36b7ee437f2c11d21db223578/src/Vulkan/Silk.NET.Vulkan/Vk.cs#L149-L155

Silk 3 implementation:
https://github.com/dotnet/Silk.NET/blob/7a5299e51253102b65f25675d19ab31e3a15e868/sources/Vulkan/Vulkan/Vk.cs#L181-L187

# Steps to reproduce

On a device that has support for the `VK_KHR_calibrated_timestamps` extension, load the extension using `TryGetDeviceExtension` and call `GetPhysicalDeviceCalibrateableTimeDomain`.

You will get a `Native symbol not found (Symbol: vkGetPhysicalDeviceCalibrateableTimeDomainsKHR)` error and a Vulkan validation warning if Vulkan validation layers are turned on.

Calling `GetInstanceProcAddr`, as suggested by the warning, will result in a valid function pointer and no validation errors.

# Comments

Since I'm the one in charge of the Vulkan bindings for Silk 3, I'll fix this at some point before the full release.

I'll use a workaround in my own codebase for Silk 2.
This seems to require reading the object type graph defined in the Vulkan XML spec and modifying codegen, so not a trivial fix.

This works as a workaround:
```cs
// Workaround for https://github.com/dotnet/Silk.NET/issues/2547
[...]

Issue #2586, opened 2026-06-06
[3.0] Exanite's Tracking Issue


Exanite:
# Exanite's Tracking Issue

This tracks a bunch of miscellaneous tasks that I accumulated while working on Silk 3.
I'm open for anyone to work on these tasks, but please discuss with me in the Silk Discord first.

A task being here on this tracker does not mean it is guaranteed to be completed.
This tracker's primary goal is so that we do not accidentally forget about tasks and that any tasks not completed are intentionally left uncompleted.

## In-progress

Tasks in progress will be moved to the corresponding pull request.

## Uncategorized

This section contains informal reports, suggestions, or complaints about Silk 3.
I'm logging them here so they don't get forgotten about until I either open a formal issue or address them.

From Aqua:

- [ ] libSDL3.so is humongous (should not be nearly 14 MB)
   - Exanite: I checked my system version and that one is only 3 MB.
- [ ] Silk 3 does not prefer its own native binaries? I.e., it prefers the system ones instead of the ones shipped by Silk.
- [ ] Go through the thread Aqua has made: https://discord.com/channels/521092042781229087/1512914692853465088

From Ethereal:

- [ ] SilkTouch generator takes an immense amount of RAM: https://discord.com/channels/521092042781229087/607634593201520651/1513592461493145601
   - I need to also check with Curin to see how much RAM their device has. If their device doesn't have enough RAM, this might explain why the Microsoft job takes so long (aside from it already processing a lot of data).
   - [ ] Run the Microsoft job locally and see how much RAM it takes on my device.

From Nolram (https://discord.com/channels/521092042781229087/607634593201520651/1524213012402147348):

- [ ] Add Imgui bindings
   - Nolram recommended https://github.com/dearimgui/dear_bindings over cimgui (https://github.com/cimgui/cimgui)
- [ ] Investigate whether direct P/Invoke and static linking works in Silk 3 and how we want to support it.
   - See: I'm not familiar with what direct P/Invoke means though
   - Specifically, it involves some configuration, providing static builds, and maybe switching over to LibraryImport over DllImport.

From dfkeenan:

- [ ] Add SDL shadercross support: https://discord.com/channels/521092042781229087/587346162802229298/1543100984270520340
   - My opinion on this: Shadercross is worth doing, worth considering the rest of the SDL addons as well, since Silk 3 leans much more heavily on SDL. 
[...]

PR #2590, opened 2026-06-12
[3.0] Add OpenCL and OpenXR bindings

Exanite:
# Summary of the PR

This PR was originally going to be a mega-PR for all of the remaining Khronos-related bindings and Assimp.
However, I've changed the scope of this PR to focus primarily on OpenCL and OpenXR.

This PR is intended to get the bindings to a point where they exist, but not yet guaranteed to be usable. This is because I want to ensure that future changes are made taking into account all of the Khronos bindings that we plan to ship with.

As such, I do not plan on testing the bindings in this PR. I will be writing smoke-tests for the different bindings available in a separate PR.

Update: I've changed the scope of this PR to only include OpenCL and OpenXR. This PR also includes some improvements to the existing Khronos bindings.

Discord thread for this PR: https://discord.com/channels/521092042781229087/1514735149915439274

# Further Comments

## Tasks

- [x] Port back the option to only run selected jobs. Done during https://github.com/dotnet/Silk.NET/pull/2591.
- [x] Check the tracking issue and figure out what APIs are part of this
- [x] Add required submodules
   - [x] OpenXR
   - [x] OpenCL
   - [x] Assimp
   - [x] SPIRV Reflect
   - [x] SPIRV Cross
   - [x] SPIRV Tools
   - [x] SPIRV Headers
   - [x] Shaderc
   - [x] MoltenVK
   - [x] GLSLang
   - [x] VMA
   - Note: I'm not going to revert the submodule additions here since I plan on using them in my next PR.
- [x] Set submodules to their latest tagged commit
- [x] Update existing Khronos submodules and regenerate
- [x] Generate initial bindings
   - This step will just involve getting the generator to run and output bindings without any obvious issues.
   - [x] OpenXR
      - Initial generation had only one problem, which was that OpenXR's XML uses constants to define array sizes. This was fixed by updating `MetadataUtils.GetTypeDetails` to naively resolve symbols.
   - [x] OpenCL
      - [x] Fix constexpr error during ClangScraper
      - ~~Exclude `cl_bool` from being generated? Question mark because I later realized that enums don't have any members. Figure out whether `cl_bool` is useful (contains members) and check if other similar APIs have a similar enum (I think OpenAL might).~~
         - Looked at Silk 2 and the XML. Bool seems to contain True, False, NonBlocking, Blocking. The last two are aliases. Will keep `cl_bool` since Silk 2 has it and there is *some* use to keeping it, even if it's not much.
      - [x] Fix vendor suffix identification not working since vendor suffixes are capitalized while OpenCL type names are not
      - [x] Fix `context` being identified as `cont + ext` where `ext` is a vendor suffix.
      - [x] Fix enums not having any members.
      - [x] Figure out why so many enums are missing base types. Note: there is one case where enums are expected to be missing base types, which is when the enum is defined in the XML, but not in the C header we parsed.
         - [x] `ErrorCodes` probably should be manually specified. Need to look in the headers for the correct type.
         - Reason is because the XML doesn't specify the base type and unlike the OpenGL-like specs, we don't have a namespace enum type (eg: `GLenum`/`CLenum`) that we can type map.
         - I opted to manually typemap these since I don't believe there will be too many changes in the future, but we can look into automating this if preferred.
      - [x] Fix issue where `cl_accelerator_info_intel` is prettified as `AcceleratorInfointel` instead of `AcceleratorInfoINTEL`.
      - [x] Decide on whether I want to special case vendor affix identification for OpenCL. OpenCL specifies vendors as prefixes, suffixes, and infixes, which is totally not annoying at all. The infix case is technically a suffix, but ends up in the middle of a name due to the dot notation that they use and how Silk converts dots to underscores.
         - I ended up not special casing vendor affix identification, but instead added a special case to transform the weird names to be in the "usual" format where the vendor affix is at the end of the name. This transformation means the vendor suffix identification code works as normal.
         - Note that I consider the implementation of this transformation to be controversial, but also believe it brings more benefits than not. See my thoughts here: https://discord.com/channels/521092042781229087/1514735149915439274/1518124361263878225
      - [x] ~~Improve handling of "blocked" enums (for lack of a better term). Some enums have numbers at the end of their name (eg: `cl_intel_advanced_motion_estimation.cl_motion_detect_desc_intel.2`). We may want to combine these enums together.~~
         - [x] Look at what Silk 2 does
            - Silk 2 leaves the enum~~s~~ as is.
         - [x] Check if this is a valid transformation
            - The different sections must have the same base type and must not have enum members that share the same values (ie, a conflict) 
            - Yes, this is valid for the single enum this affects
         - Apparently this affects 1 singular enum. Probably still worth handling though. `IntelAdvancedMotionEstimationClMotionDetectDescIntel2` is the current name and contrasts with `IntelAdvancedMotionEstimationClMotionDetectDescINTEL`, which has the vendor suffix properly identified.
         - Notably, this enum has this comment in the XML spec: `Part of previous block?`
         - Aside: The usage of `block` in the tasks here and SilkTouch source code is to be consistent with this comment. I was going to call these "sectioned" enums rather than "blocked" enums, but noticed that the spec calls the different sections "blocks".
         - Edit: Decided against merging `cl_intel_advanced_motion_estimation.cl_motion_detect_desc_intel.2` with the non-2 version. The non-2 version focuses on direction while the 2 version focuses on weights/fractions, so it makes sense to keep them separate.
      - [x] Add missing opaque structs. These have the pattern of `typedef struct _Type* Type` in the headers. I handled these the same way `GLsync` is handled in OpenGL since that type has the same pattern.
          - I originally used `AddOpaqueStructs` for these, but then realized `ExtractHandles` works on OpenCL. Not sure why it doesn't for `OpenGL`. 
      - [x] Add TransformHandles mod so that these opaque structs are transformed as handle types.
- [x] Update OpenGL to use TransformHandles. This has the effect of replacing `Sync*` with `SyncHandle`.
- [x] Do a polish pass over the bindings
   - This step will involve creating the manually written files required for the bindings and reviewing every single file for issues. 
   - [x] OpenXR
      - [x] Add the required manually written files
      - [x] Figure out what to do for OpenXR's graphics and platform integrations. They can't be generated using a single platform unlike when using the XML. This is because we'll have missing types that are OS-specific (Metal, Win32, X11, etc). Does a cross-compilation like approach help with this? Separate integration projects? Manual stubs for missing types? Need to investigate and discuss.
         - Decision: I'm going to defer implementing this to a later PR. This means that OpenXR is going to be unusable until that PR, but handling multi-platform generation needs additional work in the core generator. See: https://discord.com/channels/521092042781229087/587346162802229298/1529139068288503900
   - [x] OpenCL
      - [x] Add the required manually written files
      - [x] Fix naming of function pointer structs like `EnqueueSvmFreeArmPfnFreeFunc`
         - "Native" name is `clEnqueueSVMFreeARM_pfn_free_func` (function + function param name)
         - [x] Document the new `FunctionPointerNameFallbackParent` affix category.
         - [x] Document that qualified scopes are now supported for referenced name affixes.
      - [x] Undo my underscore change to `ExtractNestedTypes` or apply the same change to `ExtractFunctionPointers` since it also has the same behavior.
      - [x] Investigate why `ExtractEnumConstants` seems to have bad behavior with OpenCL.
          - Initial guess: Probably because I put it in front of MixKhronosData.
          - Yup, and there's also no need for `ExtractEnumConstants` since the XML data is enough.
- [x] Investigate `AddVTables.ThisThread` mod `GenerateMakeCurrentPartial` and `GenerateFactoryPartial` options
   - I need to figure out exactly when these should be used. I'm probably using them wrong right now.
   - From what I can tell:
      - Use `GenerateMakeCurrentPartial` if you want anything other than the default `Underlying.Value = ctx`.
         - Only OpenGL needs this right now
      - Use `GenerateFactoryPartial` if you want the API to be usable without calling `ThisThread.MakeCurrent`.
         - Currently used for the Khronos bindings since these are instance-based bindings.
- Other improvements 
   - Vulkan
      - [x] Switch to outputting the Flags version of the Flags/FlagBits enum names for the enums' `[NativeName]` attributes.
         - [x] Document decision in API-specific notes doc (was because the Flags version is easier to work with and causes less edge cases when handling metadata)
- [x] Lightly review generated bindings
   - Review note: Structs are usually fine. Functions and enums are usually where issues occur.
   - [x] Compare current generator settings against Silk 2's settings
   - ~~Compare current bindings output against Silk 2's output~~
      - This is going to be a bit too painful to do, so I'm only comparing suspicious cases that I happen to find.
   - [x] Investigate any empty enums
      - OpenCL has a significant number of these, but most involve platform-specific headers. Deferring to later PR.
      - Relevant regex: `enum.+\{ \}`
- [x] Ensure native builds are up to date
   - [x] OpenAL
- [x] Update docs
   - [x] Generator usage docs
      - [x] CLI options
      - [x] Clarify that non `.gen.cs` `.cs` files do not get modified by SilkTouch
      - [x] Verify usage against my experience during this PR
   - [x] Any mod changes
   - [x] Add new bindings to changelog
   - [x] Static vs Instance bindings docs. OpenCL might need to be added to this list.
- [x] Rebase onto `develop/3.0`
- [x] Generate using Windows for consistency
   - This must include regenerating *every* existing binding
- [x] Deeply review generated bindings

## Split into future PRs

Khronos vendor suffixes:

- [x] ~~Reconsider the usage of KhronosImpliedVendor suffixes~~
   - Apparently aliased enums are simply replaced when bindings are updated. This means that users will naturally replace usages of those enums to point to their new versions. Eg: EXT to KHR. Issue is, with implied vendors, this also means that users might accidentally use the wrong enum member since the KHR version of the enum member now takes the name of the old EXT enum member. It might be better to forgo prettiness in favor of correctness here.
   - [x] ~~Also tempted to remove KhronosNonExclusiveVendors since I find this behavior to be more incorrect than it is helpful. It also lets me delete a lot of code (yay!).~~
   - History note: I believe both of these behaviors were ported from Silk 2. I ported these two behaviors from the INameTrimmer-based implementation to the current name-affix-based implementation.

Platform interop / additional headers:

Most Khronos APIs have additional headers that interop with platform-specific/other Khronos/other graphics APIs.
OpenXR notably will not work without these additional headers, but generating bindings for these headers requires additional work. See this to understand the planned solution: https://discord.com/channels/521092042781229087/587346162802229298/1529139068288503900

- [ ] Note down what headers are not being scraped for each Khronos API
   - [ ] Eg: Vulkan:
      - [ ] header.h
- [ ] Add aforementioned headers to each bindings generation process
- [ ] Investigate OpenCL empty enums
   - Most of these should be populated once the other headers are included
   - [ ] cl_command_buffer_info_khr
   - [ ] cl_command_buffer_properties_khr
   - [ ] cl_command_buffer_state_khr
   - [ ] cl_command_buffer_update_type_khr
   - [ ] cl_command_properties_khr
   - [ ] cl_compiler_mode_altera
   - [ ] cl_d3d10_device_set_khr
   - [ ] cl_d3d10_device_source_khr
   - [ ] cl_d3d11_device_set_khr
   - [ ] cl_d3d11_device_source_khr
   - [ ] cl_dx9_device_set_intel
   - [ ] cl_dx9_device_source_intel
   - [ ] cl_gl_context_info
   - [ ] cl_gl_object_type
   - [ ] cl_gl_texture_info
   - [ ] cl_layer_properties
   - [ ] cl_media_adapter_set_khr
   - [ ] cl_media_adapter_type_khr
   - [ ] cl_mutable_command_info_khr
   - [ ] cl_mutable_dispatch_asserts_khr (Flags so by default 1 None member)
   - [ ] cl_mutable_dispatch_fields_khr (Flags)
   - [ ] cl_platform_command_buffer_capabilities_khr (flags)
   - [ ] cl_queue_properties_khr
   - [ ] cl_semaphore_reimport_properties_khr
   - [ ] cl_svm_alloc_properties_khr
   - [ ] cl_svm_capabilities_khr (Flags)
   - [ ] cl_svm_free_flags_khr (Flags)
   - [ ] cl_svm_free_properties_khr
   - [ ] cl_svm_pointer_info_khr
   - [ ] cl_va_api_device_set_intel
   - [ ] cl_va_api_device_source_intel 
- [ ] Also investigate any other empty enums, if there are any. Make sure to check for [Flags] enums as well, since these will not be empty and instead have a single `None` member.

Khronos bindings improvements:

- [ ] Unconditional usage of `clGetExtensionFunctionAddressForPlatform` limits our OpenCL bindings to OpenCL v1.2.
   - Don't fix directly. This will be fixed once we track which functions can be loaded using each approach.
   - See: https://discord.com/channels/521092042781229087/587346162802229298/1532127670303854632
   - See: https://github.com/dotnet/Silk.NET/issues/2547
- [ ] OpenXR/IdentifiedSharedPrefixes: Not sure why `XrPerfSettingsNotificationLevelEXT` needs a prefix override for proper shared prefix identification. I know the type name affects it, but I thought that was a fallback case.
   - Same issue with `XrStructureType` and `XrSceneMarkerQRCodeSymbolTypeMSFT`
- [ ] OpenXR: Consider remapping vector/box/extent/etc types
   - See my list here: https://github.com/dotnet/Silk.NET/pull/2590#discussion_r3686865142
   - [ ] Also do the same for the other APIs where relevant
- [ ] Decide what to do with names in the format: `1St`, `2Nd`

Config/DI/registration system:

- [ ] Ensure mod order for all jobs are consistent

Remaining bindings I plan to work on:

MoltenVK would just be a native build. VMA is not on the tracking issue, but we already discussed that we want this to be in Silk 3 (and I want it for my personal projects as well).

- [ ] Generate initial bindings
   - [ ] Assimp
   - [ ] SPIRV Reflect
   - [ ] SPIRV Cross
   - [ ] SPIRV Tools
   - [ ] Shaderc
   - [ ] VMA
- [ ] Do a polish pass over the bindings
   - [ ] Assimp
   - [ ] SPIRV Reflect
   - [ ] SPIRV Cross
   - [ ] SPIRV Tools
   - [ ] Shaderc
   - [ ] VMA
- [ ] Native builds
   - [ ] Assimp
   - [ ] SPIRV Reflect
   - [ ] SPIRV Cross
   - [ ] SPIRV Tools
   - [ ] Shaderc
   - [ ] VMA
   - [ ] MoltenVK
- [ ] Update docs
   - [ ] Any mod changes
   - [ ] Add new bindings to changelog
- [ ] Rebase onto `develop/3.0`
- [ ] Generate using Windows for consistency
   - This must include regenerating *every* existing binding


eng/submodules/egl
Exanite, line 1: I renamed this so that we're more explicit on the submodule names. Some APIs have multiple repos (usually 1 for XML spec and 1 for the headers) and I want to make it easier to see which is which.

sources/OpenAL/OpenAL/al/IALContext.gen.cs
Exanite, line 1: Only changes are due to updating the OpenAL submodule. Notably, some parameters got updated to use the Constant type.

sources/SilkTouch/SilkTouch/Mods/TransformHandles.cs
```diff
@@ -233,9 +233,7 @@ public override SyntaxNode VisitStructDeclaration(StructDeclarationSyntax node)
             var structName = node.Identifier.Text;
             return node.WithIdentifier(Identifier(structName))
                 .WithAttributeLists(
-                    new SyntaxList<AttributeListSyntax>()
-                        .WithNativeName(structName)
```
Exanite, line 237: WithNativeName has been moved to ExtractHandles and AddOpaqueStructs since that is where the structs are first introduced.

sources/SilkTouch/SilkTouch/Mods/Metadata/MetadataUtils.cs
```diff
@@ -41,7 +42,8 @@ public static int GetIndirectionLevels(this ReadOnlySpan<char> cType) =>
     public static void GetTypeDetails(
         this ReadOnlySpan<char> cType,
         Span<bool> mutability,
-        out int outerCount
+        out int outerCount,
+        Dictionary<string, string>? symbolMap = null
```
Exanite, line 46: This lookup is required for properly parsing the OpenXR XML, which uses named constants instead of literals. Fortunately, naive lookup will work here.

The other Khronos XMLs also use a similar syntax, but we never hit the edge case until OpenXR since OpenXR is the first to use it for function parameters.

sources/SilkTouch/SilkTouch/Mods/AddOpaqueStructs.cs
```diff
@@ -81,21 +84,27 @@ public Task ExecuteAsync(IModContext ctx, CancellationToken ct = default)
...
                 CompilationUnit()
                     .WithMembers(
                         SingletonList<MemberDeclarationSyntax>(
                             FileScopedNamespaceDeclaration(ns)
                                 .WithMembers(
                                     SingletonList<MemberDeclarationSyntax>(
-                                        StructDeclaration(name[(qualified + 1)..])
+                                        StructDeclaration(name)
+                                            .WithAttributeLists(
+                                                new SyntaxList<AttributeListSyntax>().WithNativeName(
```
Exanite, line 104: Only real change. The is me renaming things for clarity.

sources/SilkTouch/SilkTouch/Mods/PrettifyNames.cs
```diff
@@ -959,50 +972,23 @@ out _
 
             // Build result by merging fragments
             var result = "";
-            var previousFragment = new NameFragment("", false);
-            foreach (var nameFragment in nameFragments)
```
Exanite, line 963: Don't know what case I was defensively programming for here. My new test cases indicate that this is unnecessary.

This aside, I also added the Capitalize option to handle OpenCL's lowercase vendor affixes.

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
```diff
@@ -35,6 +35,17 @@ public partial class MixKhronosData(
     internal ConcurrentDictionary<string, JobData> Jobs = new();
     private static readonly char[] _listSeparators = { ',', '|', '+' };
 
+    private class NameAffixes
```
Exanite, line 41: This class is likely temporary.
I plan on adding a SilkConstants class for all/most constants used in the project.

Already added as a task in my tracker.

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
```diff
@@ -56,6 +67,7 @@ internal class JobData
         /// <remarks>
         /// This is OpenGL and OpenCL specific.
         /// </remarks>
+        // TODO: Due to historical reasons, this is keyed by the managed name of the enum. We should use the native name as the key instead since the native name does not change.
```
Exanite: Currently, MixKhronosData uses a mix of managed and native names. I'll be working to slowly update the code to use native names for metadata and managed names for transformation in the future.

sources/SilkTouch/SilkTouch/Mods/PrettifyNames.cs
Exanite: Ah wait, the consecutive acronyms case. That's what.
I need to add a test case where `AA` and `BB` are prettified affixes that are adjacent to each other. Also another test case for when `AA` is a prefix and `BB` is the base name.

tests/SilkTouch/SilkTouch/TransformHandlesTests.SuccessfullyTransformsHandleType_NoDsl.verified.txt
```diff
@@ -1,5 +1,4 @@
 ﻿// Vk.gen.cs
-[NativeName("VkInstance_T")]
```
Exanite, line 2: TransformHandles no longer outputs NativeNames. ExtractHandles does it instead.

tests/SilkTouch/SilkTouch/ExtractNestedTypesTests.SuccessfullyExtractsNestedInlineArray.verified.txt
```diff
@@ -1,16 +1,16 @@
-﻿// VkPerformanceCounterDescriptionARM.gen.cs
+﻿// VkPerformanceCounterDescriptionARM_name.gen.cs
```
Exanite, line 1: Decided to revert the change where I removed the underscore joining the two names. This is because ExtractFunctionPointers does the same thing and this is more readable in the case someone doesn't prettify the names.

tests/SilkTouch/SilkTouch/Khronos/MixKhronosDataTests.EnumGroupRegression.cl.xml.verified.txt
```diff
@@ -1,16 +1,94 @@
 ﻿NOTE: This file lists differences/regressions between 3.0's enum grouping system baselined against 2.21's. If this file is non-empty and verified, it means the regressions have been justified. Justified differences are usually as a result of a discrepancy between the regression baseline JSON and the baseline XML.
 
+Removed Group: cl_intel_advanced_motion_estimation_cl_motion_detect_desc_intel
```
Exanite, line 3: This is because I decided post-process these names so that the vendor suffixes are at the end and not repeated.

This ensures that OpenCL vendor prefixes/infixes/suffixes are all moved to the end and capitalized like in the other Khronos APIs.

tests/SilkTouch/SilkTouch/Khronos/MixKhronosDataTests.cs
```diff
@@ -94,17 +94,22 @@ private static async Task<Profile> ReadProfile(string fPath)
     }
 
     [Test, TestCaseSource(nameof(EnumTestCases))]
-    public Task EnumsToGroups(string file, object data) =>
-        Verifier
+    public Task EnumsToGroups(string file, object data)
```
Exanite, line 97: The EnumsToGroups and EnumGroupRegression are changed because the enum data is keyed by native name now instead of the managed name. The changes are to map the native name back to the managed name so that the Verify snapshot remains the same as before.

sources/Vulkan/Vulkan/Vk.cs
```diff
@@ -16,18 +16,24 @@ public partial class Vk
 
     static Vk()
     {
-        LoaderInterface.RegisterHook(Assembly.GetExecutingAssembly());
```
Exanite, line 19: Temporary solution to RegisterHook being called twice issue.
Need to discuss and implement more permanent solution in a separate PR since the issue is a bit more complex than deduping the RegisterHook call itself.

Already added as a task in my task tracker.

docs/for-contributors/Generator/generator-mods.md
```diff
@@ -170,6 +173,9 @@ provide cross-platform bindings.
 - Enums use `uint` as their backing type on Linux instead of `int` like on Windows. Silk handles this by using
   `TransformEnums` to "coerce" the backing types to their Windows equivalents when possible.
 
+- C `long` is output as C# `long` on Windows and C# `nint` on Linux. Silk handles this by generating bindings on Windows
```
Exanite, line 176: We don't support 32 bit right?

examples/CSharp/OpenCL/Tutorial001.SimpleKernel/Program.cs
```diff
@@ -0,0 +1,280 @@
+﻿// TODO: Exanite: This example is easier to read as a Program class. We should probably do the same for the other examples.
+//
+// TODO: Exanite: I plan to rename these "tutorials" to be "smoke tests" and to reserve tutorials to be for examples properly designed for educational purposes.
+// TODO: Exanite: The existing tutorials just test if the API works and I don't want to misrepresent that.
+
+using System.Runtime.CompilerServices;
+using Silk.NET.OpenCL;
+
+internal unsafe class Program
```
Exanite, line 9: Lots of todos here. I plan on addressing these in later PRs. Some todos are due to my inexperience with the Silk 3 DSL types, but are marked as todos since these are valuable cases to document.

sources/OpenCL/OpenCL/Enums/CommandBufferInfoKHR.gen.cs
```diff
@@ -0,0 +1,13 @@
...
+// Original source is Copyright 2013-2026 The Khronos Group Inc. Licensed under the Apache 2.0 license.
+using System;
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+namespace Silk.NET.OpenCL;
+
+[NativeName("cl_command_buffer_info_khr")]
+public enum CommandBufferInfoKHR : uint { }
```
Exanite, line 13: Empty enums are generally due to the headers that are not being traversed right now. Properly fixing this requires multi-platform scraping.

Plan is to handle that before checking for any more empty enums.
I'm also tracking a list of these empty enums.

sources/SilkTouch/SilkTouch/Mods/PrettifyNames.cs
Exanite: Ended up deciding to not handle this case since there's not that much of a benefit.

sources/SilkTouch/SilkTouch/Mods/PrettifyNames.cs
```diff
@@ -959,113 +974,110 @@ out _
...
             NameProcessorContext context,
-            string referenceScope,
-            string referencedMember,
-            [NotNullWhen(true)] out string? referencedMemberScope,
-            [NotNullWhen(true)] out string? referencedMemberValue,
+            string resolutionScope,
+            string reference,
+            [NotNullWhen(true)] out string? memberScope,
+            [NotNullWhen(true)] out string? memberName,
+            [NotNullWhen(true)] out string? memberValue,
```
Exanite, line 1012: Changes here were to support qualified references in referenced affixes.
Eg: `[NameAffix("Prefix", "FunctionPointerNameFallbackParent", nameof(Cl.EnqueueSvmFreeARM))]`

sources/SilkTouch/SilkTouch/Naming/NameAffixer.cs
```diff
@@ -66,24 +63,17 @@ argumentList.Arguments[2].Expression is LiteralExpressionSyntax
...
                 {
                     affixes =
                     [
                         .. affixes,
                         new NameAffix(
-                            type == "Prefix" ? NameAffixType.Prefix : NameAffixType.Suffix,
+                            affixType,
                             category,
-                            referencedAffix,
+                            nameofExpression.ToString(),
```
Exanite, line 76: Changes here were to support qualified references in referenced affixes.
Eg: `[NameAffix("Prefix", "FunctionPointerNameFallbackParent", nameof(Cl.EnqueueSvmFreeARM))]`

sources/SilkTouch/SilkTouch/Mods/ExtractNestedTypes.cs
```diff
@@ -121,7 +121,7 @@ member is not StructDeclarationSyntax structNode
                     continue;
                 }
 
-                var extractedIdentifier = $"{node.Identifier}{match.Groups[1].Value}";
+                var extractedIdentifier = $"{node.Identifier}_{match.Groups[1].Value}";
```
Exanite, line 124: Decided to revert my change here since ExtractFunctionPointers also uses the underscore and I decided this looks better.

sources/SilkTouch/SilkTouch/Mods/ExtractFunctionPointers.cs
```diff
@@ -144,25 +144,63 @@ public override SyntaxNode VisitFunctionPointerType(FunctionPointerTypeSyntax no
             }
 
             // As above, get the native type name if we can and also get a fallback name based on context.
-            var (currentNativeTypeName, fallback) = current switch
+            const string functionPointerNameFallbackParent = "FunctionPointerNameFallbackParent";
```
Exanite, line 147: Changes here were to ensure that OpenCL function pointer structs get named properly.

Eg: Previously, the `ARM` in `clEnqueueSVMFreeARM_pfn_free_func` was being prettified as `Arm` since that lack of a referenced affix meant we didn't have vendor suffix data.

In this case, the NameAffix looks like this:
`[NameAffix("Prefix", "FunctionPointerNameFallbackParent", nameof(Cl.EnqueueSvmFreeARM))]`

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
```diff
@@ -43,23 +54,23 @@ internal class JobData
         public required Configuration Configuration { get; init; }
 
         /// <summary>
-        /// A mapping of enum native names to group names.
+        /// A mapping from enum native names to group native names.
```
Exanite, line 57: MixKhronosData now uses the native names of enums to key these dictionaries.

sources/OpenXR/OpenXR/OpenXR/Vector2F.gen.cs
```diff
@@ -0,0 +1,22 @@
...
+// Original source is Copyright 2017-2026 The Khronos Group Inc. Licensed under the MIT license.
+using System;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+namespace Silk.NET.OpenXR;
+
+[NativeName("XrVector2f")]
+[SupportedApiProfile("openxr")]
+public partial struct Vector2F
```
Exanite, line 13: We probably can remap these. I think I remember seeing a task/issue about this somewhere.

Exanite: Other relevant types:
Uuid
UuidMSFT

Vector2F
Vector3F
Vector4F
Vector4SFB

Color3F
Color4F

Rect2Df
Rect2Di
Rect3DfFB

Extent2Df
Extent2Di
Extent3Df
Extent3DiMETA

Offset2Df
Offset2Di
Offset3DfFB

Boxf
Spheref
Quaternionf

Also in the Silk 2 typemap:
```json
"XrUuidMSFT": "Guid"
"XrSceneSphereBoundMSFT": "Silk.NET.Maths.Sphere<float>"
"XrVector4sFB": "Silk.NET.Maths.Vector4D<short>"
```

Less relevant:
Fovf
Posef
Frustumf 

Note there may be others. We should review this more closely when addressing this issue.
We should regex for these (eg: ending in f/i) and also check the non-XrStructure types that make use of the above types.

sources/OpenXR/OpenXR/OpenXR/Version.gen.cs
```diff
@@ -0,0 +1,28 @@
...
+[NativeName("XrVersion")]
+[SupportedApiProfile("openxr")]
+public enum Version : ulong
+{
+    [SupportedApiProfile(
+        "openxr",
+        ["XR_LOADER_VERSION_1_0", "XR_VERSION_1_0", "XR_VERSION_1_1"],
+        MinVersion = "1.0"
+    )]
+    X1x0 = 1,
```
Exanite: No native name? And probably want to reduce the trimming here.

sources/OpenXR/OpenXR/OpenXR/PerfSettingsNotificationLevelEXT.gen.cs
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Silk.NET.OpenXR;
+
+[NativeName("XrPerfSettingsNotificationLevelEXT")]
+[SupportedApiProfile("openxr")]
+public enum PerfSettingsNotificationLevelEXT : uint
+{
+    [NativeName("XR_PERF_SETTINGS_NOTIF_LEVEL_NORMAL_EXT")]
+    [SupportedApiProfile("openxr", ["XR_EXT_performance_settings", "XR_EXT_thermal_query"])]
+    NotifLevelNormal = 0,
```
Exanite: Why is this not getting shared prefix trimmed?

Edit: I added a prefix override for now, but it was only identifying `XR_PERF_SETTINGS` before. I added a task for investigating this (note I know the type name affects it).

sources/OpenXR/OpenXR/OpenXR/SceneMarkerQRCodeSymbolTypeMSFT.gen.cs
```diff
@@ -0,0 +1,30 @@
...
+[SupportedApiProfile("openxr")]
+public enum SceneMarkerQRCodeSymbolTypeMSFT : uint
+{
+    [NativeName("XR_SCENE_MARKER_QR_CODE_SYMBOL_TYPE_QR_CODE_MSFT")]
+    [SupportedApiProfile(
+        "openxr",
+        ["XR_MSFT_scene_marker"],
+        ImpliesSets = ["XR_MSFT_scene_understanding", "XR_VERSION_1_0"]
+    )]
+    QrCodeSymbolTypeQrCode = 1,
```
Exanite: Same issue with prefix trimming here.

Exanite (commented):
Self code review complete. Not going to mark PR as ready to review just yet. I need to review my own comments first since it's been so long. Most just explain the changes I made, but some represent future tasks (already added to task tracker, either in this PR description or in my tracking issue here: https://github.com/dotnet/Silk.NET/issues/2586).

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
Exanite: This has been implemented.

sources/OpenXR/OpenXR/OpenXR/Version.gen.cs
Exanite: Seems to be an issue in ExtractEnumConstants. I've noted this down as a task for a later PR.
Edit: Also, marking as resolved because this enum shouldn't actually exist. The members are from include guard macros.

Exanite:
Build currently fails due to https://github.com/dotnet/ClangSharp/issues/836, but otherwise, this PR is ready for review.

docs/for-contributors/Generator/generator-mods.md
Perksey: Silk.NET 2.X made an explicit choice that 32-bit support was on a "best effort" basis. We never tried to explicitly _not_ support 32-bit, and indeed with .NET being as wonderful as it is there are seldom issues with doing so. But I think when we added Direct3D 9 bindings we made an explicit choice that 32-bit support would not be at any cost, we'd prioritise 64-bit if necessary. Regarding this specific issue, it is my expectation that the CLong type is used where possible, but I obviously recognise that can't be used for an enum's backing type. That being said, if you do `Constant<CLong, SomeEnum>`, it _should_ do the right thing. I can't remember whether there's anything in the generator currently that would stop that being possible.

examples/CSharp/OpenCL/Tutorial001.SimpleKernel/Program.cs
```diff
@@ -0,0 +1,280 @@
+﻿// TODO: Exanite: This example is easier to read as a Program class. We should probably do the same for the other examples.
+//
+// TODO: Exanite: I plan to rename these "tutorials" to be "smoke tests" and to reserve tutorials to be for examples properly designed for educational purposes.
```
Perksey: We should probably discuss this at some point. I've mostly been viewing these tutorials as having a strict linear progression, whereas I think that's probably wrong. That it, there isn't really any need for our tutorials to be numbered on the website or have "chapters" etc. Granted, it's nice for a user to be able to follow _an_ order of articles where each one builds on knowledge in the last, so I'd still want us to have a specific order of our articles (which also means our tutorials/examples/whatever we end up calling them), but the point I'm more trying to illustrate is that I'm happy with having arbitrary examples so long as we're still able to order them. The more resources we're able to provide to the user, the better.

docs/for-contributors/Generator/generator-mods.md
Exanite: I see. I'll add a task to my tracker to look over the bindings and update the bindings to use `CLong` where possible.

Edit: Although, I should probably mention that CLong is for supporting the difference between Unix and Windows ABIs. I think I was reminded of 32-bit here so even though it's not the same thing, I put the comment here. Notably, CLong is required for FreeType since they use CLongs in their public API.

docs/for-contributors/Generator/generator-mods.md
```diff
@@ -111,6 +111,9 @@ Stateless APIs are typically accessed through static methods, while stateful API
 objects. More information on this can be found in the
 [Static vs Instance Bindings](../../silk.net/static-vs-instance-bindings.md) document.
 
+`[DllImport]` is used over `[LibraryImport]` because `[LibraryImport]`'s source generator ends up using `[DllImport]`
+in its generated code.
```
Perksey: ...and all of our generated bindings are either blittable or has marshalling generated by SilkTouch instead of LibraryImport.

docs/for-contributors/Generator/using-the-generator.md
```diff
@@ -211,14 +214,59 @@ static Vk()
...
+### Initial Project Creation
+
+You must create the initial project for the generated bindings to go in. This also goes for the solution file.
+SilkTouch does not modify either file.
+
+The paths for these files correspond to the following options in the `generator.json` file:
+- `Jobs.JOB_NAME.SourceProject`
+- `Jobs.JOB_NAME.Solution`
+
+(TODO: Does this apply to `Jobs.JOB_NAME.TestProject` as well?)
```
Perksey: Tests are optional but where provided it's the same.

examples/CSharp/OpenCL/Tutorial001.SimpleKernel/Program.cs
```diff
@@ -0,0 +1,280 @@
...
+
+    public static int Main(string[] args)
+    {
+        // TODO: Exanite: The todos in this file represent issues that I found or thoughts that I had while working on this smoke test.
+        // TODO: Exanite: These might not represent actual problems, but I want to document these so we can address them later.
+
+        // Load API
+        // TODO: These todos apply to all Silk bindings.
+        // TODO: Consider making INativeContext public. It's nice for debugging.
+        // TODO: Consider returning the concrete class instead of the interface. Having to cast to access Cl.CurrentPlatform is annoying.
```
Perksey: I would love to do this, but becuase C# forbids static and non-static methods having the same signature, the instance ones are explicit interface implementations (and therefore won't be visible on the concrete class). It is possible that some assembly rewriting could make both the instance and the static methods visible, but I have no idea how the language & IDEs would react.

examples/CSharp/OpenCL/Tutorial001.SimpleKernel/Program.cs
```diff
@@ -0,0 +1,280 @@
...
+            hostData1[i] = _initialData1[i % _initialData1.Length];
+            hostData2[i] = _initialData2[i % _initialData2.Length];
+            expectedOutput[i] = hostData1[i] + hostData2[i];
+        }
+
+        // Select platform and device
+        // TODO: I'm going to prefer the DSL types whenever possible
+        // TODO: Return type should be ErrorCodes
+        var platform = default(PlatformIdHandle); // TODO: Using the DSL types requires us to initialize the variable first, which is annoying
+        CheckError(cl.GetPlatformIDs(1, platform.AsRef(), nullptr));
```
Perksey, line 90: Is there a GetPlatformID overload?

sources/OpenCL/OpenCL/Enums/CommandBufferInfoKHR.gen.cs
Perksey: Sounds good to me.

sources/OpenCL/OpenCL/OpenCL/NameVersionKHRName.gen.cs
```diff
@@ -0,0 +1,18 @@
...
+using System;
+using System.Diagnostics.CodeAnalysis;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+namespace Silk.NET.OpenCL;
+
+[NativeName("_name_e__FixedBuffer")]
+[InlineArray(64)]
+public partial struct NameVersionKHRName
```
Perksey, line 14: It's a shame we can't keep the KHR at the end haha

sources/OpenCL/OpenCL/Cl.cs
```diff
@@ -0,0 +1,118 @@
...
+
+public partial class Cl
+{
+    public const string ErrMultiplePlatformSingleObject =
+        "CurrentPlatform cannot be changed once set, use another API object for additional platforms. For more "
+        + "info, see https://dotnet.github.io/Silk.NET/docs/v3/silk.net/static-vs-instance-bindings";
+
+    static Cl()
+    {
+        LoaderInterface.RegisterAlternativeName("opencl", "OpenCL");
```
Perksey, line 16: OpenCL32? CL?

sources/OpenCL/OpenCL/Cl.cs
```diff
@@ -0,0 +1,118 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using Silk.NET.Core.Loader;
+
+namespace Silk.NET.OpenCL;
+
+public partial class Cl
+{
+    public const string ErrMultiplePlatformSingleObject =
```
Perksey: Would rather this not be public

sources/OpenXR/OpenXR/OpenXR/ActionCreateInfoActionName.gen.cs
```diff
@@ -0,0 +1,19 @@
...
+using System;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+namespace Silk.NET.OpenXR;
+
+[NativeName("_actionName_e__FixedBuffer")]
+[InlineArray(64)]
+[SupportedApiProfile("openxr")]
+public partial struct ActionCreateInfoActionName
```
Perksey, line 14: Something to add to the tracker is being able to implicitly cast to this type from a string or array etc (essentially this should behave like a Ref<sbyte>, just with a fixed backing buffer rather than being a pointer to some other buffer)

sources/OpenXR/OpenXR/OpenXR/Extent2Df.gen.cs
```diff
@@ -0,0 +1,30 @@
...
+// Original source is Copyright 2017-2026 The Khronos Group Inc. Licensed under the MIT license.
+using System;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+namespace Silk.NET.OpenXR;
+
+[NativeName("XrExtent2Df")]
+[SupportedApiProfile("openxr")]
+public partial struct Extent2Df
```
Perksey, line 13: Should this implement the Silk.NET.Maths interface when landed? Either in a manual part, or by making SilkTouch smart enough (i.e. in a mod) to detect maths types?

tests/SilkTouch/SilkTouch/Khronos/MixKhronosDataTests.EnumGroupRegression.cl.xml.verified.txt
Perksey: Sounds good to me

tests/SilkTouch/SilkTouch/ExtractNestedTypesTests.SuccessfullyExtractsNestedInlineArray.verified.txt
Perksey: Sounds good to me

sources/SilkTouch/SilkTouch/Mods/MixKhronosData.cs
```diff
@@ -35,6 +35,17 @@ public partial class MixKhronosData(
     internal ConcurrentDictionary<string, JobData> Jobs = new();
     private static readonly char[] _listSeparators = { ',', '|', '+' };
 
+    private class NameAffixes
+    {
+        public const string KhronosFunctionDataType = "KhronosFunctionDataType";
```
Perksey: Could make the RHS a nameof the LHS

Perksey (commented):
Some extremely minor comments... pretty happy overall with this PR :))))))) thank you!!!!!

examples/CSharp/OpenCL/Tutorial001.SimpleKernel/Program.cs
Exanite: I also recently realized that we don't always return the class itself and it can be something like `StaticWrapper<DllImport>`, so this is a point against returning the concrete class in all bindings. I guess the "expose property through interface" approach is more reasonable here.

sources/OpenCL/OpenCL/OpenCL/NameVersionKHRName.gen.cs
Exanite: I'm not sure if you saw the original notes I had on this, but they're available here: https://discord.com/channels/521092042781229087/1485943448069738608/1487577229121818766

What I concluded was that, aside from technical reasons for not doing this, that this approach was more user friendly, since a type and their related types have a shared prefix, which is helpful for auto-complete.

Below is the verbatim copy of the linked message:
> The names here are examples, so ignore the exact names.
> 
> **Case 1**
> For handle structs, we suffix the native names with `-Handle`. This decision was made a long time ago to avoid conflicts with user wrappers of these handles, which often want to use the same name.
> 
> When vendor suffixes are used, Perksey has said he prefers names like `BufferHandleEXT` over `BufferEXTHandle`.
> I agree here since it does make it a lot more readable.
> 
> **Case 2**
> For nested structs (usually inline arrays), we use the native name + the field name the inline array is used for.
> 
> For example with `PerformanceCounterDescriptionARMName`, the name is used for the inline array struct.
> `PerformanceCounterDescriptionARM` is the struct in which the inline array is used.
> `Name` is the name of the field that uses the inline array.
> 
> In this case, pulling `-ARM` to the end feels wrong.
> This is because `-Name` is an addition to an existing type, rather than its own standalone type.
> 
> **Case 3 (The current issue)**
> For function pointer structs and delegates, we use the native name for the struct type and suffix the delegate type with `-Delegate`.
> 
> When vendor suffixes are used, the struct type is unaffected.
> However, for the delegate type, we similarly have two options: `CallbackDelegateEXT` or `CallbackEXTDelegate`.
> 
> I don't believe Perksey said anything on this case (I never asked him directly about it).
> I originally implemented this as __Case 1__ (`CallbackDelegateEXT`) to be consistent with the handle types.
> However, now that I'm implementing __Case 2__, treating the delegate type as __Case 2__ (`CallbackEXTDelegate`) seems better since `-Delegate` similarly is an addition to an existing type, rather than a modification of a base type.
> 
> **Other examples**
> Hmm, as I was looking for examples, I found the following files in the below screenshot.
> Notably, it's not obvious that `DebugReportCallbackDelegateEXT` is added due to `DebugReportCallbackEXT`.

I also have this documented here: https://github.com/dotnet/Silk.NET/blob/b8df8e738ec6debccc8414500fe57e207271aa66/docs/for-contributors/Generator/name-processing.md?plain=1#L334-L354

docs/for-contributors/Generator/using-the-generator.md
Exanite: On this note, do you want me to add test projects for all of the existing bindings? I'll do this in a separate PR if so (already tentatively added to my tracker).

examples/CSharp/OpenCL/Tutorial001.SimpleKernel/Program.cs
Exanite: Good point, there is unfortunately not. I'm guessing this is something we'll need to add to the XML.

sources/OpenCL/OpenCL/Cl.cs
Exanite: Done. I also updated this in OpenAL, Vulkan, and OpenXR alongside OpenCL here.

sources/OpenCL/OpenCL/Cl.cs
Exanite: I based the alt names here on this: https://github.com/dotnet/Silk.NET/blob/57e0f8643c07702a16e14c027b1f15d60809d15b/generator.json#L639-L647

However, I can do some investigation on the other OpenCL library names.

sources/OpenXR/OpenXR/OpenXR/Extent2Df.gen.cs
Exanite: I think it should be manual. I'm not sure what kind of heuristics we'll need to do this automatically.
Alternative is to specify which types should receive transformations, but I'm not sure how much specialization is needed for each API.

sources/OpenXR/OpenXR/OpenXR/ActionCreateInfoActionName.gen.cs
Exanite: I added this to my tracker here: https://github.com/dotnet/Silk.NET/issues/2586

docs/for-contributors/Generator/using-the-generator.md
Perksey: It probably makes sense to do as it's free, but yeah not for this PR.

examples/CSharp/OpenCL/Tutorial001.SimpleKernel/Program.cs
Perksey: Yeah will have to add to XML, but it's worth looking into without doing that as iirc BuildTools had some Khronos-specific logic that would detect this function pattern and do a "benefit of the doubt" overload, which in 3.0 speak would likely be in the IFunctionTransformer implementation in MixKhronosData.

Possibly one for the tracker?

sources/OpenCL/OpenCL/OpenCL/NameVersionKHRName.gen.cs
Perksey: It's clear this is a thought-through decision, I shall agree with it!

examples/CSharp/OpenCL/Tutorial001.SimpleKernel/Program.cs
Exanite: Yup, I already added this to my tracker.

examples/CSharp/OpenCL/Tutorial001.SimpleKernel/Program.cs
Exanite: Okay, so discussed in Discord: https://discord.com/channels/521092042781229087/587346162802229298/1541823300672692275

Conclusion is that we're not going to distinguish between smoke tests and tutorials because people are going to use them as reference material anyway. This means the baseline quality expectation is that we have done our due diligence to ensure that the code is of reasonable quality and demonstrates something of value to the user.

As for numbering/ordering, we'll still have an order to the tutorials/examples, but we will avoid splitting tutorials into multiple chapters where each chapter strictly builds on top of the previous.

sources/OpenCL/OpenCL/Cl.cs
Exanite: Apparently `OpenCL32.dll` and `OpenCL64.dll` both exist, but they are Nvidia-specific from what I can tell. I'm not familiar with the binaries, so I can't say whether they are ABI compatible with the Silk bindings and whether we should register them as alternative names.

`libcl.so` (so lowercase `cl`) exists as part of https://github.com/intel/beignet, but this repo has been archived since 3 years ago. The repo points to https://wiki.freedesktop.org/www/Software/Beignet/ and that in turn points to two repos:
> Right now, we host our project on fdo at: http://cgit.freedesktop.org/beignet/. And the Intel 01.org: https://01.org/beignet

But neither of these links are still alive, so I doubt the project is alive and likely to be used by people picking up Silk 3.

I'm not sure where capital `CL` comes from, so let me know if this something that should be registered. It's really hard to find information on this, so I'd rather defer to your judgement.

All in all, these seem like rarer cases to me, and potentially unsupportable due to ABI differences, so in the case that someone wants to use them, they can register the names themselves. Also, if these exist, why didn't Silk 2 include them?

Perksey: Fair enough

Perksey (commented):
Happy to approve once the example is cleaned up a bit (TODOs removed, oh and don't forget the license header!)

PR #2596, opened 2026-06-28
Fix exception when using ALContext after disposing an extension

midare160:
# Summary of the PR
- Fixed an error when disposing an extension retrieved via `alc.TryGetExtension<T>()` and disposing the parent context itself afterwards
- Moved the disposal of of the `Context`-Property from `NativeAPIContainer` to `NativeAPI` since disposing any extension left the parent context in a broken state


midare160:
@dotnet-policy-service agree

Perksey (approved):
This is great, many thanks!

AltexGames (approved):

PR #2605, opened 2026-07-27
Fix SdlGamepad/SdlJoystick state going stale with no event subscribers

### LINKED ISSUE #2604: SdlGamepad should update internal button/trigger states even if no events are subscribed
## Summary

SdlGamepad never updates its internal button/trigger state unless something is subscribed to `ButtonDown`/`ButtonUp`/`TriggerMoved`. If nobody subscribes to those events, polling `IGamepad.Buttons` or `IGamepad.Triggers` always returns stale data (buttons always show not pressed), even while the controller is actually being pressed.

## Cause

In `SdlGamepad.cs`, the button down/up handling looks like this:

```csharp
case EventType.Controllerbuttondown:
{
    var ogBtn = _buttons[@event.Cbutton.Button];
    ButtonDown?.Invoke
        (this, _buttons[@event.Cbutton.Button] = new Button(ogBtn.Name, ogBtn.Index, true));
    break;
}
```

The `_buttons[...] = new Button(...)` assignment is written inside the argument list of the `ButtonDown?.Invoke(...)` call. In C#, `x?.Invoke(a, b)` short-circuits the entire call, including evaluating the arguments, when `x` is null. So if nobody has subscribed to `ButtonDown`, the assignment to `_buttons[...]` never happens at all.

Same thing for `ButtonUp` and `TriggerMoved`.

Thumbstick axis motion does not have this problem, because the array assignment there happens as its own statement before the event is invoked.

## Repro

1. Get an `IGamepad` from an SDL `IInputContext`, do not subscribe to `ButtonDown`, `ButtonUp`, or `TriggerMoved`.
2. Press a button on the controller.
3. Poll `gamepad.Buttons` each frame. It never shows the button as pressed.
4. Now subscribe an empty handler to `ButtonDown` (`gamepad.ButtonDown += (g, b) => {};`) and repeat. Now it works.

Analog sticks work fine the whole time regardless of subscription, since they don't have this bug.

## Impact

Any app that polls gamepad state instead of listening to events gets completely broken button/trigger reads, with no error and no obvious reason why, since nothing about the polling code looks wrong. We hit this in Gum (a UI library) where we poll `Buttons`/`Triggers` per frame the normal way you'd expect a state query API to work. Took a while to track down since the actual driver code was correct and the bug is in an unrelated backend detail.

## Fix

Move the array assignment out of the `Invoke` argument list so it always runs regardless of subscribers, e.g.:

```csharp
case EventType.Controllerbuttondown:
{
    var ogBtn = _buttons[@event.Cbutton.Button];
    var newBtn = new Button(ogBtn.Name, ogBtn.Index, true);
    _buttons[@event.Cbutton.Button] = newBtn;
    ButtonDown?.Invoke(this, newBtn);
    break;
}
```

Same fix for `ButtonUp` and `TriggerMoved`.

## Workaround

Our current workaround (in our app, not in Silk.NET) is to always subscribe a no-op handler to `ButtonDown`, `ButtonUp`, and `TriggerMoved` right after getting an `IGamepad`, just to force the internal state to update. This works but shouldn't be necessary.

vchelaru:
Fixes #2604

SdlGamepad and SdlJoystick's button/axis/trigger/hat handlers wrote the internal state array update as a side effect inside the argument list of the `?.Invoke(...)` call for their change events (`ButtonDown`, `ButtonUp`, `TriggerMoved` on SdlGamepad; `AxisMoved`, `HatMoved`, `ButtonDown`, `ButtonUp` on SdlJoystick). C#'s null-conditional operator short-circuits the whole call, including evaluating the arguments, when the event has no subscribers. So if nobody's subscribed to those events, the array never actually gets updated, and polling the corresponding state (`Buttons`, `Triggers`, `Axes`, `Hats`) reads permanently stale data even with real input on the device.

ThumbstickMoved on SdlGamepad already did this correctly (array assignment as its own statement before the event fires), so this brings the rest in line with that same shape.

Full writeup, repro steps, and confirmation this affects the latest release/main (but not GlfwGamepad, which rewrites state unconditionally every frame) are in #2604.

### DIFF
--- a/src/Input/Silk.NET.Input.Sdl/SdlGamepad.cs
+++ b/src/Input/Silk.NET.Input.Sdl/SdlGamepad.cs
@@ -128,14 +128,14 @@ namespace Silk.NET.Input.Sdl
                         }
                         case GameControllerAxis.ControllerAxisTriggerleft:
                         {
-                            TriggerMoved?.Invoke
-                                (this, _triggers[0] = new Trigger(0, (float) @event.Caxis.Value / short.MaxValue));
+                            var trigger0 = _triggers[0] = new Trigger(0, (float) @event.Caxis.Value / short.MaxValue);
+                            TriggerMoved?.Invoke(this, trigger0);
                             break;
                         }
                         case GameControllerAxis.ControllerAxisTriggerright:
                         {
-                            TriggerMoved?.Invoke
-                                (this, _triggers[1] = new Trigger(1, (float) @event.Caxis.Value / short.MaxValue));
+                            var trigger1 = _triggers[1] = new Trigger(1, (float) @event.Caxis.Value / short.MaxValue);
+                            TriggerMoved?.Invoke(this, trigger1);
                             break;
                         }
                     }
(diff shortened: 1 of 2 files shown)


vchelaru:
@dotnet-policy-service agree [company="{your company}"]

@dotnet-policy-service agree

Perksey (approved):
Good spot, thanks!

PR #2614, opened 2026-09-04
Improve ImGui fractional scaling support

jaredh:
# Summary of the PR
Looking at improving support for Wayland fractional scaling in [BlackLabelHQ/RecompOne](https://github.com/BlackLabelHQ/RecompOne) ([issue[1]](https://github.com/BlackLabelHQ/SymphonyRecomp/issues/46)) has led me to a couple fixes here. Looks like a lot of this is improving in 3.0 with windowing API improvements, but not sure when to expect that to hit GA.

These fixes are reproducible in the ImGui sample app, and seem to stem from deviation from the original C++ port[[2](https://github.com/ocornut/imgui/blob/v1.82/backends/imgui_impl_glfw.cpp#L360-L362)][[3](https://github.com/ocornut/imgui/blob/v1.82/backends/imgui_impl_opengl3.cpp#L272-L274)][[4](https://github.com/ocornut/imgui/blob/v1.82/backends/imgui_impl_opengl3.cpp#L449-L450)]. The tl;dr is integer division was being performed, truncating fractional values, and never updating the viewport with the framebuffer size. This was resulting in the UI rendering seemingly correct, but mouse input not lining up. Note the cursor position and the selected _Configuration_ row in this screenshot :point_down: 

Screenshot From 2026-09-04 17-22-42

I've also exposed `GetWindowContentScale`, which GLFW added at the same time as the monitor scale API; this is required downstream so we can access the fractional scale value from Wayland/mutter. 

A platform guard has been added to Silk.NET.Input so I can build this on Linux without the iOS target.

# Related issues
[1] https://github.com/BlackLabelHQ/SymphonyRecomp/issues/46
[2] https://github.com/ocornut/imgui/blob/v1.82/backends/imgui_impl_glfw.cpp#L360-L362
[3] https://github.com/ocornut/imgui/blob/v1.82/backends/imgui_impl_opengl3.cpp#L272-L274
[4] https://github.com/ocornut/imgui/blob/v1.82/backends/imgui_impl_opengl3.cpp#L449-L450

# Further Comments

I've fixed and tested this against the OpenGL implementation only. I'm not familiar with the status of Vulkan or if it's required there.


jaredh:
@dotnet-policy-service agree

src/OpenGL/Extensions/Silk.NET.OpenGL.Extensions.ImGui/ImGuiController.cs
```diff
@@ -236,12 +236,19 @@ public void Update(float deltaSeconds)
         private void SetPerFrameImGuiData(float deltaSeconds)
         {
             var io = ImGuiNET.ImGui.GetIO();
+            var size = _view.Size;
+            var framebufferSize = _view.FramebufferSize;
+
+            // Update here since Resize isn't raised on all platforms
+            _windowWidth = size.X;
+            _windowHeight = size.Y;
```
jaredh, line 244: When going fullscreen Resize is fired and the updated window size is correctly reflected, but when returning to windowed mode on Wayland, GLFX does not end up invoking the resize callback so we never end up receiving the updated window dimensions.

Perksey (approved):
Thanks!

PR #2617, opened 2026-09-23
Update ANGLE libraries

uwx:
# Summary of the PR
Updates the native libraries for Silk.NET.OpenGLES.ANGLE.Native:
- linux-arm64
- linux-x64
- osx
- win-x64 (was previously a x86 binary)
- win-x86

All builds use default settings except no Dawn/WebGPU backend (not stable anyway so it doesn't matter)

Spun off from the work in https://github.com/uwx/Maxine.Silk.EGL

# Further Comments
While the builds were done with assistance of AI, all the code that gets compiled is the Google original as you can see from upstream. So it's up to you whether you want to accept it but I don't think it's a problem here.


Perksey:
Outside of the outstanding AI issue, we usually make a point of building our native binaries in CI (see build/nuke/Native) to reduce the possibility of supply chain attacks by trusting self-compiled binaries. Would it be possible to look into updating this? CI will then build the binaries (though I'm not sure whether this works in forks at the moment, feel free to do this as a branch on the main repository)

uwx:
Well that's what I did for the upstream project, but there isn't really a way for me to plug those changes back into Silk that doesn't imply laundering AI code into a no-AI project, since I've already reviewed all of the AI's build scripts and it's not as if I would write them differently myself. Also the existing ANGLE binaries in Silk except for macOS don't have a build script attached (and one of them is mislabeled) so you've been historically trusting a random binary for a while now.

Perksey:
Ah. Hmmm, ok. Will await team's thoughts on the AI contributions as I do think this is something worth having but if them's the breaks then there's nothing I can do I'm afraid :/

